PR #3, opened 2015-09-25
sanlock: do not abort if groups is empty

germanovm:
File "/usr/lib/python2.6/site-packages/vdsm/tool/configurators/sanlock.py", line 87, in isconfigured
    .strip().split(" ")]
    ValueError: invalid literal for int() with base 10: ''

Signed-off-by: Germano Veit Michel <germano@redhat.com>

### DIFF
--- a/lib/vdsm/tool/configurators/sanlock.py
+++ b/lib/vdsm/tool/configurators/sanlock.py
@@ -75,7 +75,7 @@ def isconfigured():
                     if status_line.startswith(proc_status_group_prefix):
                         groups = [int(x) for x in status_line[
                             len(proc_status_group_prefix):]
-                            .strip().split(" ")]
+                            .strip().split(" ") if x]
                         break
                 else:
                     raise InvalidConfig(


nirs:
What is wrong with the current code?

Also, vdam does not aacept pull requests, you sould submit patchs to http://gerrit.ovirt.org.

nirs:
On which system this code fail?
What is the "Groups: ..." line that make this code fail?

germanovm:
Hello nirs.

The current code aborts with an empty "Groups: ".
Sure, split() seems to be a better idea.
I will submit the patch through gerrit then.

Regards,

nirs:
Hi Germano,

We should not blow up with empty groups since we add our groups to sanlock, but this seems like a bug in the sanlock package on your distribution. On RHEL 7, after fresh install of sanlock without vdsm, it runs with:

    Groups:	6 179 

Please open a bug for sanlock package of your distribution.

Thanks

germanovm:
Hey nirs,

Yeah, I know that those groups are not supposed to be empty. It's happening for a single customer in RHEL 6.7. It seems more like a timing issue since vdsm does not wait for sanlock to reach the group setup code before starting. Anyway, I think vdsm should not abort because of this. So it's actually 2 bugs, vdsm should not abort, and there is something wrong with RHEL 6.7 (or that customer setup).

I will work more on this on monday,

Thanks!

germanovm:
Hey Nir, just pushed it according to the ovirt gerrit wiki. Let me know if there is anything else I should do. Thank you!

https://gerrit.ovirt.org/46742

nirs:
Since the patch is in gerrit, we can close this now

PR #4, opened 2016-03-19
tests: Skip test if veth_pair cannot be created

nirs:
The context manager used try-except-finally in the wrong way, always
deleting the interface, even when we failed to create it.

Change-Id: I2f59fe4686df6ec2f7c081ee182fada19fc4cb63


nirs:
Submitted to gerrit.

PR #11, opened 2016-10-31
readme: Fix build status url

nirs:
Previously we pointed to Nir's private fork, now that oVirt's github is
triggering builds we should point to the correct builds.

Change-Id: If40bda074383ffa1c4626bb7e68a6035cf7039e3
Signed-off-by: Nir Soffer <nirsof@gmail.com>


nirs:
Merged

PR #31, opened 2022-01-10
automation: Add network tests

almusil:
Add network tests that will be triggered only when
something in network code changes, there is no need
to run them with every vdsm change as the functional
tests can take a while.

### DIFF
--- a/.github/workflows/network.yml
+++ b/.github/workflows/network.yml
@@ -1,0 +1,34 @@
+name: Network CI
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'lib/vdsm/network/**'
+      - 'lib/vdsm/common/**'
+      - 'tests/network/**'
+      - '.github/workflows/network.yml'
+  pull_request:
+    paths:
+      - 'lib/vdsm/network/**'
+      - 'lib/vdsm/common/**'
+      - 'tests/network/**'
+      - '.github/workflows/network.yml'
+
+jobs:
+  tests:
+    env:
+      IMAGE_TAG: ${{ matrix.tag }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        type: [ unit, integration, functional ]
+        tag: [ centos-8, centos-9 ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install podman
+      - name: Run ${{ matrix.type }} tests
+        run: sudo -E ./tests/network/${{ matrix.type }}/run-tests.sh
--- a/tests/network/nettestlib.py
+++ b/tests/network/nettestlib.py
@@ -214,8 +214,8 @@ class Dummy(Interface):
     def create(self):
         try:
             linkAdd(self.dev_name, linkType='dummy')
-            self.set_managed()
             self.up()
+            self.set_managed()
         except IPRoute2Error as e:
             pytest.skip(
                 f'Failed to create a dummy interface {self.dev_name}: {e}'


almusil:
@tinez @mz-pdm @nirs Please take a look

.github/workflows/network.yml
```diff
@@ -0,0 +1,70 @@
+name: Network CI
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'lib/vdsm/network/**'
+      - 'lib/vdsm/common/network/**'
```
mz-pdm, line 8: Can network tests be affected when something in common/* changes?

.github/workflows/network.yml
```diff
@@ -0,0 +1,70 @@
...
+      - 'tests/network/**'
+      - '.github/workflows/network.yml'
+  pull_request:
+    paths:
+      - 'lib/vdsm/network/**'
+      - 'lib/vdsm/common/network/**'
```
mz-pdm, line 14: The same here.

.github/workflows/network.yml
almusil: Well it might be, I don't mind including whole common there if that's ok with others.

.github/workflows/network.yml
```diff
@@ -0,0 +1,70 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install podman
+      - name: Run functional tests
+        run: |
+          sudo -E ./tests/network/functional/run-tests.sh \
+          --switch-type=${{ matrix.switch_type }} \
```
tinez, line 68: Not necessarily for this PR, but given that we pass 'matrix.switch_type' as a cmdline option maybe it would be better to do the same with 'matrix.tag' instead of passing it as env variable.

tinez (dismissed):

.github/workflows/network.yml
nirs: Sounds right to depend on common.

.github/workflows/network.yml
```diff
@@ -0,0 +1,70 @@
...
+      matrix:
+        tag: [ centos-8, centos-9 ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install podman
+      - name: Run unit tests
+        run: sudo -E ./tests/network/unit/run-tests.sh
```
nirs, line 33: Why do you need to use podman? It does not work with the builtin docker support?

.github/workflows/network.yml
```diff
@@ -0,0 +1,70 @@
...
+      matrix:
+        tag: [ centos-8, centos-9 ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install podman
+      - name: Run integration tests
+        run: sudo -E ./tests/network/integration/run-tests.sh
```
nirs, line 49: All the jobs have the same configuration, it looks like the only change is the name of the script. Can we use a matrix that use the same configuration to run "unit", "integration", and "functional" tests?

.github/workflows/network.yml
nirs, line 68: If you change ./tests/network/functional/run-tests.sh to run all switch types when running
without arguments, the only difference in the jobs is the name of the test subdirectory.

nirs (commented):
How much time the network tests take? How will this affect builds when we change something in common if we depend on entire common directory?

.github/workflows/network.yml
almusil: I had some issues with docker and systemd in containers. Podman is proven to work with that out of the box.

.github/workflows/network.yml
almusil: Unit and integration can be matrix, functional need additional argument.

.github/workflows/network.yml
almusil: The configuration slightly differs for the switch types. And also having it running in one job would mean twice the time for that single job.

almusil:
> How much time the network tests take? How will this affect builds when we change something in common if we depend on entire common directory?

~10 min. So having it depend on whole common means that change in common would trigger this.

.github/workflows/network.yml
almusil: Only issue is that for unit and integration it would require more work as those scripts do not accept cmd options, but definitely possible.

nirs:
> > How much time the network tests take? How will this affect builds when we change something in common if we depend on entire common directory?
> 
> ~10 min. So having it depend on whole common means that change in common would trigger this.

I think this if fine.

.github/workflows/network.yml
nirs: Much better now! could be nice if we had one job with matrix to run all 3 types of tests but this can be cone later.

nirs (approved):

almusil:
Added one tiny commit because there is probably regression in NM which caused some tests to fail.

nirs (approved):

PR #35, opened 2022-01-11
virt: Don't assume that VM is UP after a failed migration

mz-pdm:
Current migration code switches VM status to UP
unconditionally (unless in recovery) after a failed migration.  But
this is wrong, the migrated VM could be paused initially, it could be
paused due to a storage error, it could crash, etc.

Tracking these somewhat corner case during migration would be
difficult.  The migration either succeeds and then the VM is simply
down or it fails.  In the latter case, it's easiest to use our status
recovery routine and ask libvirt about the true status of the VM.
We additionally ensure that we won't get stuck in the
MIGRATION_SOURCE status by still switching to UP in such a case.

Change-Id: I30fbbf71dea9c32cb0af8910111c7483d2475144
Signed-off-by: Milan Zamazal <mzamazal@redhat.com>

### DIFF
--- a/lib/vdsm/virt/migration.py
+++ b/lib/vdsm/virt/migration.py
@@ -352,10 +352,10 @@ class SourceThread(object):
         # either way, migration has finished
         self._switch_state(State.FAILED)
         if self._recovery:
-            self._vm.set_last_status(vmstatus.UP, vmstatus.MIGRATION_SOURCE)
             self._recovery = False
-        else:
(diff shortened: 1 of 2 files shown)


mz-pdm:
Moved from gerrit: https://gerrit.ovirt.org/c/vdsm/+/117959

mz-pdm:
This attempts to handle some corner cases not known to be seen in production and requires more understanding of the possible problems. Not worth the effort now.

PR #38, opened 2022-01-11
tests: yajsonrpc: Use TRAVIS_CI on broken tests

tinez:
We don't use jenkins CI anymore.

### DIFF
--- a/tests/lib/yajsonrpc/jsonrpcserver_test.py
+++ b/tests/lib/yajsonrpc/jsonrpcserver_test.py
@@ -154,7 +154,7 @@ class JsonRpcServerTests(TestCaseBase):
         return res
 
     @permutations(USE_SSL)
-    @broken_on_ci(reason="Fails randomly in CI")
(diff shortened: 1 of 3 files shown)


almusil:
Maybe scope for a different PR, but we should probably consider changing `TRAVIS_CI` to `OVIRT_CI` and adjust the tests accordingly. Another option might be to use something GH specific like `GITHUB_ACTIONS`.

nirs (approved):
Good for quickly fixing the issue, but it would better to update @broken_on_ci to use the right vlaue, so we don't need these manual changes.

nirs:
> Maybe scope for a different PR, but we should probably consider changing `TRAVIS_CI` to `OVIRT_CI` and adjust the tests accordingly. Another option might be to use something GH specific like `GITHUB_ACTIONS`.

+1 to replace TRAVIS_CI with GITHUB_CI. It should be the default replacing OVIRT_CI.

almusil (approved):

nirs (approved):

PR #44, opened 2022-01-16
github: Add experimental OST workflow

nirs:
Start OST run for current pull request when adding a comment like:

    /ost

We need to improve this later so only maintainers can start OST or
starting OST is possible only if the patch is approved.

### DIFF
--- a/.github/workflows/ost.yml
+++ b/.github/workflows/ost.yml
@@ -1,0 +1,10 @@
+name: OST
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  ost:
+    if: ${{ github.event.issue.pull_request && github.event.comment.body == '/ost'}}
+    uses: oVirt/ovirt-system-tests/.github/workflows/ost.yaml@master
+    with:
+      pr_url: ${{ github.event.issue.pull_request.url }}


.github/workflows/ost.yml
```diff
@@ -0,0 +1,10 @@
+name: OST
+on:
+  issue_comment:
+    types: [created]
+jobs:
+  ost:
+    if: ${{ github.event.issue.pull_request && github.event.comment.body == '/ost'}}
```
michalskrivanek, line 7: ideally add check for approved PRs - https://github.com/oVirt/ovirt-system-tests/blob/master/.github/example-trigger.yaml#L9

michalskrivanek:
ok for trying things out. It would be best in future to tie it with build job, since now it expects artifacts to exist. When you push a new commit and comment "/ost" quickly ebfore artifacts are built the ost run will fail. It does take ~2-3 minutes to start to the chances are not that high though. Might metter most for ovirt-engine with slow builds...

.github/workflows/ost.yml
michalskrivanek: and maybe some feedback when it is not approved, perhaps as a separate job with opposite "if" condition. If you create something better please contribute to https://github.com/oVirt/ovirt-system-tests/blob/master/.github/example-trigger.yaml (via gerrit still)

nirs:
> It would be best in future to tie it with build job, since now it expects artifacts to exist. When you push a new commit and comment "/ost" quickly ebfore artifacts are built the ost run will fail. It does take ~2-3 minutes to start to the chances are not that high though. Might metter most for ovirt-engine with slow builds...

Triggering OST form the build may create too many ost runs, I would not do that.

Triggering when adding a comment sounds like the right approach. I think what is
needed to complete this is:
- Check if this pull request was already built, do nothing if not.
- Check if this pull request is approved, do nothing if not.

I'm not sure about creating "interactive" action, that will post a "reply" when
ost cannot run yet (e.g not approved or not built) but I don't have a better
idea now.

It could be nice if the action can add a comment replying the /ost comment:

    > /ost
    Build in progress, please retry when build is complete.

But it would be more useful if the action will wait until the build complete before
triggering ost.

We also need a ost build url so the user can find the build later.

michalskrivanek:
> > It would be best in future to tie it with build job, since now it expects artifacts to exist. When you push a new commit and comment "/ost" quickly ebfore artifacts are built the ost run will fail. It does take ~2-3 minutes to start to the chances are not that high though. Might metter most for ovirt-engine with slow builds...
> 
> Triggering OST form the build may create too many ost runs, I would not do that.
> 
> Triggering when adding a comment sounds like the right approach. I think what is needed to complete this is:
> 
>     * Check if this pull request was already built, do nothing if not.
> 
>     * Check if this pull request is approved, do nothing if not.
> 
> 
> I'm not sure about creating "interactive" action, that will post a "reply" when ost cannot run yet (e.g not approved or not built) but I don't have a better idea now.

the thing is that at least with vdsm it's more likely the build succeeds before OST gets to getting artifacts...so maybe really not needed. let's see

> It could be nice if the action can add a comment replying the /ost comment:
> 
> ```
> > /ost
> Build in progress, please retry when build is complete.
> ```

it can do that...but it doesm't really tell you anything. it is in progress only when you see the check starting...
 
> But it would be more useful if the action will wait until the build complete before triggering ost.

it would, it's not that hard to do when builds are small and quick, it is more tricky if we should try to reuse the last build when it exists 

> We also need a ost build url so the user can find the build later.

it's in the check result

michalskrivanek:
approval check on GH side was wrong...please update the vdsm-side issue trigger with changes made in https://github.com/oVirt/ovirt-system-tests/blob/master/.github/example-trigger.yaml

nirs:
> approval check on GH side was wrong...please update the vdsm-side issue trigger with changes made in https://github.com/oVirt/ovirt-system-tests/blob/master/.github/example-trigger.yaml

Thanks, that looks better than abusing approval for running ost. I will use the same code.
I wonder if we can have a common action for this, instead of duplicating the yaml in all
projects, and trying to keep it synced after changes.

michalskrivanek:
we have common code (in OST repo it refers to), this is the minimal boiler plate to call it:-)

nirs:
> we have common code (in OST repo it refers to), this is the minimal boiler plate to call it:-)

Yes, but hits has too much details that should be the same for all projects:

> name: OST example trigger
> on:
>  issue_comment:
>    types: [created]

All projects need this

> jobs:
>  trigger-ost:
>    if: |
>      ${{ github.event.issue.pull_request &&
>          startsWith(github.event.comment.body, '/ost') &&
>          ( github.event.comment.author_association == 'OWNER' ||
>            github.event.comment.author_association == 'MEMBER' ||
>            github.event.comment.author_association == 'COLLABORATOR' )
>      }}

All projects need this

>    uses: oVirt/ovirt-system-tests/github/ost.yaml@master
>    with:
>      pr_url: ${{ github.event.issue.pull_request.url }}

Same - maybe the action can get github.event.issue.pull_request.url
of the calling action?

It would be nicer if this was:
```
name: ost
jobs:
  ost:
    uses: oVirt/ovirt-system-tests/github/ost.yaml@master
```

nirs:
Replaced by #63

PR #48, opened 2022-01-18
virt: Add support for parallel migration connections

mz-pdm:
libvirt added the capability of using several parallel migration
connections to speed up migrations.  This patch adds support for a new
migration parameter `parallel' to use the feature.

It is expected that Engine will provide maxBandwidth argument to
maximize throughput when parallel connections are requested.  If
maxBandwidth is not provided, the default Vdsm limit is applied as
usual.

It is possible to explicitly disable parallel migration connections by
setting the `parallel' parameter to 0.  This can be useful if implicit
parallel migration connection setting for a VM is introduced in
future and we need to disable it for a particular migration.

Note that it is meaningful to set the number of parallel migration
connections to 1.  In such a case, the feature is enabled for the
migration, which means a different way of copying data is used in QEMU
than without this feature enabled and it can make performance or
reliability differences.

This patch needs support on the Engine side to provide the new
parameter.

Moved from gerrit: https://gerrit.ovirt.org/c/vdsm/+/116313

Change-Id: Ic19764b3d467f1fa007c7b92f11ef07ac3eb806a
Bug-Url: https://bugzilla.redhat.com/1975720
Signed-off-by: Milan Zamazal <mzamazal@redhat.com>

### DIFF
--- a/lib/vdsm/virt/migration.py
+++ b/lib/vdsm/virt/migration.py
@@ -130,6 +130,7 @@ class SourceThread(object):
     A thread that takes care of migration on the source vdsm.
     """
     _RECOVERY_LOOP_PAUSE = 10
+    _PARALLEL_CONNECTIONS_DISABLED_VALUE = 0
 
     ongoingMigrations = DynamicBoundedSemaphore(1)
 
@@ -138,7 +139,7 @@ class SourceThread(object):
                  tunneled=False, dstqemu='', abortOnError=False,
                  consoleAddress=None, compressed=False,
                  autoConverge=False, recovery=False, encrypted=False,
-                 cpusets=None, **kwargs):
+                 cpusets=None, parallel=None, **kwargs):
         self.log = vm.log
         self._vm = vm
         self._dom = DomainAdapter(self._vm)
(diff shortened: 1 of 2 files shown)


lib/vdsm/virt/migration.py
```diff
@@ -151,6 +152,9 @@ def __init__(self, vm, dst='', dstparams='',
         self._consoleAddress = consoleAddress
         self._dstqemu = dstqemu
         self._encrypted = encrypted
+        if parallel == self._PARALLEL_CONNECTIONS_DISABLED_VALUE:
+            parallel = None
```
ahadas, line 156: reading the [wiki page](https://www.ovirt.org/develop/release-management/features/virt/parallel-migration-connections.html) again, you wrote that if the number of connections is set to 1 we should rather not use parallel connections but here it's disabled only in case of null or 0 - do we expect the engine not to specify parallel=1 ?

mz-pdm: Yes, it's up to Engine and not Vdsm to decide.  If we ever want to permit parallel migrations with a single connection, it's better when the support in Vdsm is already present.

ahadas: ack

ahadas (approved):

tinez (approved):
LGTM. Maybe we could use '0' as the default value in the schema and in the function's signature instead of handling 'None'.

liranr23 (approved):

mz-pdm:
Maybe having None instead of just using 0 is indeed superfluous.  I'm not sure, I tend to prefer having a clearly distinguished "unspecified" value rather than using one of the values as a special value.  In any case, it's easier to switch to "0" default value than vice versa in future.

mz-pdm:
OST passed.

PR #49, opened 2022-01-19
Use lvm devices in lvm commands instead of filter

vjuranek:
This is follow-up patch of PR #27.

PR #27 introduced new lvm configuration, which uses devices file instead lvm filter. Not to touch user devices accidentally, vdsm allows lvm to use only local devices and all other devices are by default excluded by lvm filter configured by vdsm. For each lvm command vdsm provides lvm config where vdsm sets lvm filter where given device upon lvm command should act is explicitly enabled. However, once lvm is configured to use devices file, lvm filter is ignored.

With devices files we can achieve same result by adding only local devices into lvm devices file (done in PR #27) and then explicitly allow devices in lvm commands by providing `--devices $dev1, ..., $devN` argument in lvm commands. This is implemented in this PR - when vdsm is configured to use devices file, `--devices` argument is used in lvm commands instead of filter.

As vdsm still can be configured to use lvm filter, no code related to lvm filter is removed from vdsm at this point. The only missing step is to enable lvm devices file as the default choice in vdsm.

### DIFF
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -257,13 +257,18 @@ USER_DEV_LIST = [d for d in config.get("irs", "lvm_dev_whitelist").split(",")
                  if d is not None]
 
 
-def _buildFilter(devices):
-    devices = set(d.strip() for d in chain(devices, USER_DEV_LIST))
+def _prepare_device_set(devs):
+    devices = set(d.strip() for d in chain(devs, USER_DEV_LIST))
     devices.discard('')
+    if devices:
+        devices = sorted(d.replace(r'\x', r'\\x') for d in devices)
+    return devices
+
+
+def _buildFilter(devices):
     if devices:
         # Accept specified devices, reject everything else.
         # ["a|^/dev/1$|^/dev/2$|", "r|.*|"]
-        devices = sorted(d.replace(r'\x', r'\\x') for d in devices)
         pattern = "|".join("^{}$".format(d) for d in devices)
         accept = '"a|{}|", '.format(pattern)
     else:
@@ -406,7 +411,8 @@ class LVMCache(object):
     def _getCachedFilter(self):
         with self._filterLock:
             if self._filterStale:
-                self._filter = _buildFilter(multipath.getMPDevNamesIter())
+                device_set = _prepare_device_set(multipath.getMPDevNamesIter())
+                self._filter = _buildFilter(device_set)
                 self._filterStale = False
             return self._filter
 
@@ -414,7 +420,8 @@ class LVMCache(object):
         newcmd = [constants.EXT_LVM, cmd[0]]
 
         if devices:
-            dev_filter = _buildFilter(devices)
+            device_set = _prepare_device_set(devices)
+            dev_filter = _buildFilter(device_set)
         else:
             dev_filter = self._getCachedFilter()
 
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -256,6 +256,8 @@ backup {
 USER_DEV_LIST = [d for d in config.get("irs", "lvm_dev_whitelist").split(",")
                  if d is not None]
 
+USE_DEVICES = config.get("lvm", "config_method").lower() == "devices"
+
 
 def _prepare_device_set(devs):
     devices = set(d.strip() for d in chain(devs, USER_DEV_LIST))
@@ -424,7 +426,12 @@ class LVMCache(object):
         else:
             device_set = self._cached_devices()
 
-        dev_filter = _buildFilter(device_set)
+        if USE_DEVICES:
+            if device_set:
+                newcmd += ["--devices", ",".join(device_set)]
+            dev_filter = ""
+        else:
+            dev_filter = _buildFilter(device_set)
 
         conf = _buildConfig(
             dev_filter=dev_filter,
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -238,7 +238,7 @@ devices {
  ignore_suspended_devices=1
  write_cache_state=0
  disable_after_error_count=3
- filter=%(filter)s
+ %(filter)s
  hints="none"
  obtain_device_list_from_udev=0
 }
@@ -273,7 +273,10 @@ def _buildFilter(devices):
     return '[{}"r|.*|"]'.format(accept)
 
 
-def _buildConfig(dev_filter, use_lvmpolld="1"):
+def _buildConfig(dev_filter="", use_lvmpolld="1"):
+    if dev_filter:
+        dev_filter = f"filter={dev_filter}"
+
     conf = LVMCONF_TEMPLATE % {
         "filter": dev_filter,
         "use_lvmpolld": use_lvmpolld,
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -421,9 +421,10 @@ class LVMCache(object):
 
         if devices:
             device_set = _prepare_device_set(devices)
-            dev_filter = _buildFilter(device_set)
         else:
-            dev_filter = _buildFilter(self._cached_devices())
+            device_set = self._cached_devices()
+
+        dev_filter = _buildFilter(device_set)
 
         conf = _buildConfig(
             dev_filter=dev_filter,
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -391,9 +391,9 @@ class LVMCache(object):
         """
         self._runner = cmd_runner
         self._cache_lvs = cache_lvs
-        self._filter = None
-        self._filterStale = True
-        self._filterLock = threading.Lock()
+        self._devices = None
+        self._devices_stale = True
+        self._devices_lock = threading.Lock()
         self._lock = threading.Lock()
         self._cmd_sem = threading.BoundedSemaphore(self.MAX_COMMANDS)
         self._stalepv = True
@@ -408,13 +408,13 @@ class LVMCache(object):
     def stats(self):
         return self._stats
 
-    def _getCachedFilter(self):
-        with self._filterLock:
-            if self._filterStale:
-                device_set = _prepare_device_set(multipath.getMPDevNamesIter())
-                self._filter = _buildFilter(device_set)
-                self._filterStale = False
-            return self._filter
+    def _cached_devices(self):
+        with self._devices_lock:
+            if self._devices_stale:
+                self._devices = _prepare_device_set(
+                    multipath.getMPDevNamesIter())
+                self._devices_stale = False
+            return self._devices
 
     def _addExtraCfg(self, cmd, devices=tuple(), use_lvmpolld=True):
         newcmd = [constants.EXT_LVM, cmd[0]]
@@ -423,7 +423,7 @@ class LVMCache(object):
             device_set = _prepare_device_set(devices)
             dev_filter = _buildFilter(device_set)
         else:
-            dev_filter = self._getCachedFilter()
+            dev_filter = _buildFilter(self._cached_devices())
 
         conf = _buildConfig(
             dev_filter=dev_filter,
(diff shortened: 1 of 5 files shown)


vjuranek:
This PR wasn't tested with OST or manually in oVirt env. yet.

lib/vdsm/storage/lvm.py
```diff
@@ -281,6 +283,17 @@ def _buildConfig(dev_filter, use_lvmpolld="1"):
...
 
+def _allowed_devices(devices):
+    devs = set(d.strip() for d in chain(devices, USER_DEV_LIST))
+    devs.discard('')
+    if devs:
+        devs = sorted(d.replace(r'\x', r'\\x') for d in devs)
+        dev_list = ",".join(dev for dev in devs)
+    else:
+        dev_list = None
+    return dev_list
```
nirs, line 299: Why do we need this? we already have code dealing with the list of devices for crating lvm filter.

We also cache the devices, and fetch the devices again when needed, but this code never consider the cached devices. Looks like you need to reuse the current code getting the devices instead of introducing a new partial helper.

lib/vdsm/storage/lvm.py
```diff
@@ -408,6 +421,11 @@ def _getCachedFilter(self):
     def _addExtraCfg(self, cmd, devices=tuple(), use_lvmpolld=True):
         newcmd = [constants.EXT_LVM, cmd[0]]
 
+        if USE_DEVICES:
+            allowed_devices = _allowed_devices(devices)
+            if allowed_devices:
+                newcmd += ["--devices", allowed_devices]
+
```
nirs: Adding this without removing the filter from --config does not make sense. It would be better to first change the filter code to support a config without filter, and then in this patch add either --devices or the filter, not both.

This is the general approach - make the change easy, make the easy change.

tests/storage/lvm_test.py
```diff
@@ -137,13 +139,14 @@ def test_build_command_long_filter(fake_devices):
...
+    fake_runner = FakeRunner()
+    lc = lvm.LVMCache(fake_runner)
     fake_devices.append("/dev/mapper/c")
     lc.invalidateFilter()
 
-    cmd = lc._addExtraCfg(["lvs"])
+    lc.run_command(["lvs"])
+    cmd = fake_runner.calls[0]
+
     assert cmd[3] == build_config(fake_devices)
```
nirs, line 179: Change makes sense, testing public functions is better. But this change should be done first before we make any other change to make sure it is correct with the current code and we don't break the code while adding the new feature.

lib/vdsm/storage/lvm.py
```diff
@@ -238,7 +238,7 @@ def __getattr__(self, attrName):
  ignore_suspended_devices=1
  write_cache_state=0
  disable_after_error_count=3
- filter=%(filter)s
+ %(filter)s
```
nirs, line 241: This change should be done first, before we add devices support, so when we add devices,
we can switch and use only devices without the filter.

lib/vdsm/storage/lvm.py
```diff
@@ -276,8 +276,13 @@ def _buildFilter(devices):
...
+    if dev_filter:
+        lvm_filter = "filter={}".format(dev_filter)
+    else:
+        lvm_filter = ""
+
     conf = LVMCONF_TEMPLATE % {
-        "filter": dev_filter,
+        "filter": lvm_filter,
         "use_lvmpolld": use_lvmpolld,
     }
```
nirs, line 287: Looks good

tests/storage/lvm_test.py
```diff
@@ -43,6 +43,43 @@
...
+
+
+@pytest.fixture
+def use_filter(monkeypatch):
+    monkeypatch.setattr(lvm, "USE_DEVICES", False)
+
+
+@pytest.fixture
+def use_devices(monkeypatch):
+    monkeypatch.setattr(lvm, "USE_DEVICES", True)
```
nirs, line 80: Nice

nirs (changes requested):

lib/vdsm/storage/lvm.py
nirs: This is the code ensuring that we have correct device list:

```
 401     def _getCachedFilter(self):
 402         with self._filterLock:
 403             if self._filterStale:
 404                 self._filter = _buildFilter(multipath.getMPDevNamesIter())
 405                 self._filterStale = False
 406             return self._filter
 407 
 408     def _addExtraCfg(self, cmd, devices=tuple(), use_lvmpolld=True):
 409         newcmd = [constants.EXT_LVM, cmd[0]]
 410 
 411         if devices:
 412             dev_filter = _buildFilter(devices)
 413         else:
 414             dev_filter = self._getCachedFilter()
```

This code should be refactored so we cache the devices, not the filter, and we use
the cached device when crating a filter or formatting the --devices option.

Also when invalidating the cache, the next command should get new list of devices, same
way it happens now with the filter.

This refactoring must come first, before we add devices support, since devices support depends on it.

vjuranek:
Tested common flows (putting iSCSI domain into maintenance and back, create thin and preallocated disks on iSCSI domain) and didn't find any issue so far. Example command:

```
cmd=[\'/usr/sbin/lvm\', \'vgs\', \'--devices\', \'/dev/mapper/360014055ba493238c484d9284005d1bf,/dev/mapper/360014059fff83490d9b4ddfbc7e43c20,/dev/mapper/36001405a77a1ee25cbf4439b7ddd2062,/dev/mapper/36001405cd5268d7e3b64b40a16200729,/dev/mapper/36001405e7bafcb012c74e59a51e98ccb\', \'--config\', \'devices {  preferred_names=["^/dev/mapper/"]  ignore_suspended_devices=1  write_cache_state=0  disable_after_error_count=3    hints="none"  obtain_device_list_from_udev=0 } global {  prioritise_write_locks=1  wait_for_locks=1  use_lvmpolld=1 } backup {  retain_min=50  retain_days=0 }\', \'--noheadings\', \'--units\', \'b\', \'--nosuffix\', \'--separator\', \'|\', \'--ignoreskippedcluster\', \'-o\', \'uuid,name,attr,size,free,extent_size,extent_count,free_count,tags,vg_mda_size,vg_mda_free,lv_count,pv_count,pv_name\', \'eb405d82-0cd7-45c0-bba0-3795a34a6e5b\']
```

Upgrade on the host (and enabling devices file during upgrade) wasn't tested with this patch set yet (some issue with rpm built by GH - have wrong version for some reason).

lib/vdsm/storage/lvm.py
```diff
@@ -279,10 +279,16 @@ def _buildFilter(devices):
 
 
 def _buildConfig(dev_filter, use_lvmpolld="1"):
+    if dev_filter:
+        lvm_filter = "filter={}".format(dev_filter)
+    else:
+        lvm_filter = ""
```
nirs, line 294: I think we can simplify this to:

    def _build_config(dev_filter="", use_lvmpolld="1"):
        if dev_filter:
            dev_filter =  f"filter={dev_filter}"

tests/storage/lvm_test.py
```diff
@@ -43,6 +43,33 @@
...
+global {
+ prioritise_write_locks=1
+ wait_for_locks=1
+ use_lvmpolld=1
+}
+backup {
+ retain_min=50
+ retain_days=0
+}
+"""
```
nirs, line 65: This looks like a duplicate of the template in lvm.py. What is the difference?

Before this change, we have simple text that does not need any processing. I don't
see need to duplicate the code of lvm.py for testing the code.

It is ok to keep static text - once for the case when we don't have a filter,
and one for the case we have a filter.

tests/storage/lvm_test.py
```diff
@@ -43,6 +43,33 @@
...
+
+
+def expected_config(filter=""):
+    config = EXPECTED_LVM_CONFIG % {"filter": filter}
+    return config.replace("\n", " ").strip()
```
nirs, line 70: This mostly duplicate _buildConfig()

lib/vdsm/storage/lvm.py
```diff
@@ -421,9 +421,10 @@ def _addExtraCfg(self, cmd, devices=tuple(), use_lvmpolld=True):
 
         if devices:
             allowed_devices = _allowed_devices(devices)
-            dev_filter = _buildFilter(allowed_devices)
         else:
-            dev_filter = _buildFilter(self._cached_devices())
+            allowed_devices = self._cached_devices()
+
+        dev_filter = _buildFilter(allowed_devices)
```
nirs: See comment in previous commit how this can be solved in a simpler way.

lib/vdsm/storage/lvm.py
```diff
@@ -265,6 +267,13 @@ def _allowed_devices(devs):
     return devices
 
 
+def _build_devices(devs):
+    if devs:
+        return ",".join(dev for dev in devs)
+    else:
+        return None
```
nirs, line 274: This return None which is never correct. This function accept non empty list of devices, and joins them with ",". There is no need fo function to do this.

lib/vdsm/storage/lvm.py
```diff
@@ -424,7 +433,12 @@ def _addExtraCfg(self, cmd, devices=tuple(), use_lvmpolld=True):
         else:
             allowed_devices = self._cached_devices()
 
-        dev_filter = _buildFilter(allowed_devices)
+        if USE_DEVICES:
+            if allowed_devices:
+                newcmd += ["--devices", _build_devices(allowed_devices)]
```
nirs, line 438: It is nice that we have _build_filter() and _build_devices(), but there is really nothing to build for devices, so we can just do:

    if devices:
        newcmd += ["--devices", ",".join(devices)]

nirs (changes requested):
Looks like the right direction now.

lib/vdsm/storage/lvm.py
```diff
@@ -418,7 +417,7 @@ def _addExtraCfg(self, cmd, devices=tuple(), use_lvmpolld=True):
             allowed_devices = _allowed_devices(devices)
             dev_filter = _buildFilter(allowed_devices)
         else:
-            dev_filter = self._getCachedFilter()
+            dev_filter = _buildFilter(self._cached_devices())
```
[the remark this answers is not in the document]
nirs: Sure we can add a helper to merge users devices and to escape the devices. Can be something like:

   def merge(devices):
      devices = set(d.strip() for d in chain(devices, USER_DEV_LIST))
      devices.discard('')
      return devices

   def escape(devices):
      for d in devices:
          yield d.replace(r'\x', r'\\x')

Can be used like:

    for device in escape(merge(devices)):
        ...

Or:

    cmd.append("--devices")
    cmd.append(",".join(escape(merge(devices))))

nirs:
@vjuranek is this ready for review now? All the conversations are resolved,
but you reply only to one.

vjuranek:
> @vjuranek is this ready for review now? 

yes, I forgot to re-request the review

> All the conversations are resolved, but you reply only to one.

By resolving I meant the change was implemented. Next time I'll add explicit comment about it - would be probably more easy also for me as github shows comments for given file from all commits and sometime it's little bit hard not to get lost what bellows to what (and possibly resolving conversation which shouldn't be resolved).

tinez (dismissed):

nirs:
@tinez This PR is rebased on #27 - please review the first patches that come from #27 on the
other PR. Looks like github does not work well for PR based on other PRs, and we better post
a PR only when the previous PR was merged to avoid the confusion.

nirs:
@vjuranek can you rebase on master to resolve the conflicts and make this easier to review?

nirs:
This comment that should not trigger ost run.

vjuranek:
ost: https://redir.apps.ovirt.org/dj/job/ds-ost-baremetal_manual/27238
I did also some manual tests before rebase, but will add `verified` lable after more manual tests.

nirs (dismissed):
Awesome!

vjuranek:
No changes, just rebase.

Tested also manually with various flows, including

* attaching an domain
* detaching a domain
* creating new domain
* removing a domain, zeroing data
* creating thin and preallocated disk
* removing disk
* attaching disk to a VM
* detaching disk from a VM

In all cases were used iSCSI domains/disks.

PR #51, opened 2022-01-24
github: Don't try to push to quay on forks

nirs:
The containers workflow fails every week in my vdsm fork because the
ovirt organization secrets are not available in forks. Skip the push to
quay if repository owner is not 'oVirt'.

Example run from my fork, suceeded without pushing to quay:
https://github.com/nirs/vdsm/actions/runs/1740212112

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -8,6 +8,7 @@ on:
     - cron:  '0 0 * * 0'
 env:
   IMAGE_REGISTRY: quay.io
+  can_push: ${{ github.repository_owner == 'oVirt' }}
 jobs:
   container:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/containers.yml
```diff
@@ -24,6 +25,7 @@ jobs:
         working-directory: docker/network
         run: make ${{ matrix.container }}
       - name: Push to Quay.io
+        if: ${{ env.can_push == true }}
```
almusil, line 28: I wonder, is the `== true` needed?

nirs: Without this, the if matches. Looks like you must have an expression, and single value is not an expression.

mz-pdm (approved):

PR #52, opened 2022-01-24
github: Fix can_push condition

nirs:
When we set:

    can_push: ${{ github.repository_owner == 'oVirt' }}

Github converts the boolean value to the strings 'true' or 'false'[1],
so checking:

    if: ${{ env.can_push == true }}

Does not match and now we never push to quay.

Example run on my fork show that we still skip the push on forks:
https://github.com/nirs/vdsm/actions/runs/1740365670

Unfortunately, the only way to test this properly is to merge to master.

[1] https://docs.github.com/en/enterprise-server@3.0/actions/learn-github-actions/expressions#functions

Fixes 746975bb434e5fbb4c41c3e119728bc3322fb5e9

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -25,7 +25,7 @@ jobs:
         working-directory: docker/network
         run: make ${{ matrix.container }}
       - name: Push to Quay.io
-        if: ${{ env.can_push == true }}
+        if: ${{ env.can_push == 'true' }}
         id: push-to-quay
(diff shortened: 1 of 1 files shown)


mz-pdm (approved):
OK, let's test it by merging it.

nirs:
To save another cycle, I hacked the push condition so it matches on my fork:
https://github.com/nirs/vdsm/actions/runs/1740435813

If this works the containers jobs should fail on my fork now.

tinez (approved):

PR #53, opened 2022-01-25
Use oVirt upload artifacts action

almusil:
In order to use RPMs on OST we need to adopt the common
naming. Use oVirt upload artifacts action that keeps
the convention in one place.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,6 @@ jobs:
     - name: Build RPMs
       run: ./automation/rpm.sh
     - name: Upload artifacts
-      uses: actions/upload-artifact@v2
+      uses: ovirt/upload-rpms-action@v2
       with:
-        name: artifacts
-        path: ${{ env.EXPORT_DIR }}
+        directory: ${{ env.EXPORT_DIR }}


mz-pdm (approved):

vjuranek (dismissed):

automation/rpm.sh
```diff
@@ -28,5 +28,3 @@ find . \
     -maxdepth 1 \
     -iname 'vdsm*.tar.gz' \
     -exec mv {} "${EXPORT_DIR}/" \;
-
-createrepo_c "${EXPORT_DIR}"
```
nirs, line 32: Can we keep this so rpm.sh can be reused locally or on gitlab?

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
     - name: Build RPMs
       run: ./automation/rpm.sh
     - name: Upload artifacts
-      uses: actions/upload-artifact@v2
+      uses: ovirt/upload-rpms-action@v1
       with:
-        name: artifacts
-        path: ${{ env.EXPORT_DIR }}
+        directory: ${{ env.EXPORT_DIR }}
+        distro: el8stream
```
nirs, line 51: In the current form this action is not very useful. It seem that the only
service it gives us is naming the artifact "artifact-el8stream".

mz-pdm: It makes possible to apply contingent future changes globally across oVirt, which is useful by itself.

almusil: Considering what was discussed on sync I'll wait with this change for the distro auto detection.

automation/rpm.sh
almusil: Done

mz-pdm (approved):

nirs (approved):
Thanks!

PR #59, opened 2022-02-01
automation: Update network containers to use centos as base

almusil:
The tag for build container has changed and the container itself
grew in size. With packages for JavaScript and Java there is no
point in keeping it as base.

### DIFF
--- a/docker/network/functional/Dockerfile.centos-8
+++ b/docker/network/functional/Dockerfile.centos-8
@@ -1,7 +1,11 @@
-FROM quay.io/ovirt/buildcontainer:stream8
+FROM quay.io/centos/centos:stream8
 
 # Add runtime dependencies.
 RUN dnf -y install dnf-plugins-core \
+    && \
+    dnf copr enable -y ovirt/ovirt-master-snapshot centos-stream-8 \
+    && \
+    dnf install -y ovirt-release-master \
     && \
     dnf update -y \
     && \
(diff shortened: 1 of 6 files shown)


mz-pdm (dismissed):

docker/network/functional/Dockerfile.centos-8
```diff
@@ -1,14 +1,20 @@
-FROM quay.io/ovirt/buildcontainer:stream8
+FROM centos/centos:stream8
```
nirs, line 1: We used quay.io before - who provide this container? If this goes to dockerhub, we may have throttling issues.

docker/network/functional/Dockerfile.centos-8
```diff
@@ -1,14 +1,20 @@
...
 
 # Add runtime dependencies.
 RUN dnf -y install dnf-plugins-core \
+    && \
+    dnf copr enable -y ovirt/ovirt-master-snapshot centos-stream-8 \
+    && \
+    dnf install -y ovirt-release-master \
```
nirs, line 8: In other containers we use different format:

   && do something ... \
      arg \
      arg \
   && next ...

Of course container already use this foramt. If will be nice to unify all container
to use the same format later.

docker/network/functional/Dockerfile.centos-8
almusil: Right, let's put it back just to be safe

mz-pdm (approved):

nirs (approved):

PR #60, opened 2022-02-01
net: Lower severity of nic not configured for monitoring

almusil:
The message was logged with warning, which results in some confusion
and spam in the journal. Let's have it as debug as the intention
is really to know what was happening with the monitoring.


mz-pdm:
So the message is not any helpful in production logs, right?

almusil:
Correct

mz-pdm (approved):

nirs (dismissed):

PR #63, opened 2022-02-02
github: Add OST workflow

hbraha:
Start OST run for current pull request when adding a comment like:

    /ost

We need to improve this later so only maintainers can start OST or
starting OST is possible only if the patch is approved.

replaceing PR: https://github.com/oVirt/vdsm/pull/44

### DIFF
--- a/.github/workflows/ost.yml
+++ b/.github/workflows/ost.yml
@@ -1,0 +1,16 @@
+name: OST
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  trigger-ost:
+    if: |
+      ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/ost') &&
+          ( github.event.comment.author_association == 'MEMBER' ||
+            github.event.comment.author_association == 'COLLABORATOR' )
+      }}
+    uses: oVirt/ovirt-system-tests/.github/workflows/ost.yaml@master
+    with:
+      pr_url: ${{ github.event.issue.pull_request.url }}


tinez (commented):
LGTM, but I don't think we should call it "experimental" anymore

.github/workflows/ost.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  trigger-ost:
+    if: |
+      ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/ost') &&
+          ( github.event.comment.author_association == 'OWNER' ||
```
almusil, line 11: I remember some discussion about the OWNER is this role really needed? OWNER should be oVirt org which I guess won't push PR. But don't want to block it so we can adjust in follow up.

almusil (approved):

nirs:
> LGTM, but I don't think we should call it "experimental" anymore

Fixed

.github/workflows/ost.yml
nirs: We need a link to the docs explaining what is OWNER, MEMBER, and COLLABORATOR

nirs (commented):
Looks good, but linking to the relevant docs will be nice.

.github/workflows/ost.yml
almusil: https://docs.github.com/en/graphql/reference/enums#commentauthorassociation

almusil: So from the description it seems we only need the MEMBER and COLLABORATOR.

hbraha: fixed

almusil (approved):

mz-pdm (approved):

vjuranek (approved):

PR #67, opened 2022-02-07
update passwd to use scram-sha-256

hbraha:
currently Passwd use sha-1 which is deprecated, we should move to
sha-256.

### DIFF
--- a/lib/vdsm/tool/configurators/passwd.py
+++ b/lib/vdsm/tool/configurators/passwd.py
@@ -52,6 +52,8 @@ def libvirt_sasl_isconfigured():
         # check for new default configuration - since libvirt 3.2
         if 'mech_list: gssapi\n' in lines:
             return NO
+        if 'mech_list: scram-sha-256\n' not in lines:
+            return NO
     return MAYBE
 
 
@@ -94,8 +96,8 @@ def removeConf():
 
 def configure_libvirt_sasl():
     with io.open(_SASL2_CONF, 'w', encoding='utf8') as f:
-        f.writelines([u'## start vdsm-4.20.0 configuration\n',
-                      u'mech_list: scram-sha-1\n',
+        f.writelines([u'## start vdsm-4.50.0 configuration\n',
+                      u'mech_list: scram-sha-256\n',
                       u'sasldb_path: %s\n' % (_LIBVIRT_SASLDB),
                       u'## end vdsm configuration']
                      )
--- a/tests/tool_test.py
+++ b/tests/tool_test.py
@@ -107,21 +107,28 @@ def patchConfigurators(mockConfigurers):
 
 class PasswdConfiguratorTest(VdsmTestCase):
     def testCheckIsConfiguredNo(self):
-        tmpfile = tempfile.mktemp()
-        with open(tmpfile, 'w') as f:
+        with tempfile.NamedTemporaryFile(mode='w+') as f:
             f.write("\n"
                     "\n"
                     "mech_list: gssapi\n")
+            f.flush()
 
-        passwd._SASL2_CONF = tmpfile
-        self.assertEqual(passwd.libvirt_sasl_isconfigured(), NO)
+            passwd._SASL2_CONF = f.name
+            self.assertEqual(passwd.libvirt_sasl_isconfigured(), NO)
+
+    def testCheckIsConfiguredNoWithWrongSha(self):
+        with tempfile.NamedTemporaryFile(mode='w+') as f:
+            f.write("\nmech_list: scram-sha-1\n")
+            f.flush()
+            passwd._SASL2_CONF = f.name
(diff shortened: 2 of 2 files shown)


lib/vdsm/tool/configurators/passwd.py
```diff
@@ -95,7 +95,7 @@ def removeConf():
 def configure_libvirt_sasl():
     with io.open(_SASL2_CONF, 'w', encoding='utf8') as f:
         f.writelines([u'## start vdsm-4.20.0 configuration\n',
-                      u'mech_list: scram-sha-1\n',
+                      u'mech_list: scram-sha-256\n',
```
nirs, line 98: The title of this commit should be:

   vdsm-tool: Update libvirt sasl password to scram-sha-256

lib/vdsm/tool/configurators/passwd.py
```diff
@@ -95,7 +95,7 @@ def removeConf():
 def configure_libvirt_sasl():
     with io.open(_SASL2_CONF, 'w', encoding='utf8') as f:
         f.writelines([u'## start vdsm-4.20.0 configuration\n',
```
nirs, line 97: Using 4.20.0 does not make sense now. If this configuration was added in vdsm-4.5.x we should use right value.

Note this string may be used by other parts to check if the file is configured for vdsm. It looks like this implementation is broken (broken your patch), hard coding vdsm version that is not relevant for ages.

Please check how other configurators handle this.

lib/vdsm/tool/configurators/passwd.py
tinez: +1

lib/vdsm/tool/configurators/passwd.py
tinez: The version strings added by the configurators were always "arbitrary", but we tried to keep them in sync with actual vdsm versions when changes were introduced. Given that we're actually changing the config in this patch it would be nice to bump the version to `4.50.0` too.

tinez (commented):
We still need to change the `isconfigured` logic because it won't work properly.

lib/vdsm/tool/configurators/passwd.py
hbraha: Done

lib/vdsm/tool/configurators/passwd.py
hbraha: Done

tinez (changes requested):
We still need to fix the test that broke with the change

tests/tool_test.py
```diff
@@ -107,21 +107,31 @@ def patchConfigurators(mockConfigurers):
 
 class PasswdConfiguratorTest(VdsmTestCase):
     def testCheckIsConfiguredNo(self):
-        tmpfile = tempfile.mktemp()
-        with open(tmpfile, 'w') as f:
+        tmpfile = tempfile.NamedTemporaryFile(mode='w+')
```
tinez, line 110: It would be better to collapse those 2 lines into 1: `with tempfile.Named... as f:`

tests/tool_test.py
```diff
@@ -107,21 +107,31 @@ def patchConfigurators(mockConfigurers):
...
+        with tmpfile as f:
             f.write("\n"
                     "\n"
                     "mech_list: gssapi\n")
+            f.flush()
 
-        passwd._SASL2_CONF = tmpfile
-        self.assertEqual(passwd.libvirt_sasl_isconfigured(), NO)
+            passwd._SASL2_CONF = tmpfile.name
```
tinez, line 117: I know that the original test used the same constructs, but this is actually horrible - we modify the `passwd._SASL2_CONF` for the _whole lifetime of the test module_. These kind of changes should be scoped - you change the attribute before the tests starts and then you set it back to the original value once the single test is completed.
This PR is rather urgent so I wouldn't block on it, but we should improve this test module in subsequent PRs.

tinez (dismissed):
The test module part could be improved, but we can do it in next PRs

tests/tool_test.py
```diff
@@ -107,21 +107,31 @@ def patchConfigurators(mockConfigurers):
...
+            self.assertEqual(passwd.libvirt_sasl_isconfigured(), NO)
+
+    def testCheckIsConfiguredNoWithWorngSha(self):
```
mz-pdm, line 120: ...Wrong...

mz-pdm (commented):
Please fix the test name, otherwise good for me.

tests/tool_test.py
hbraha: Done

tests/tool_test.py
hbraha: will be handle in another PR

tests/tool_test.py
hbraha: Done

tinez (approved):

mz-pdm (approved):

PR #68, opened 2022-02-07
Don't use hardcoded value for lvm devices file

vjuranek:
Currently path to lvm devices files is hardcoded in lvmdevices module and uses default lvm value. This would break if the user for some reason configures lvm to use custom devices file. This PR:
* add new function which allows to determine lvm config value which will be actually used by lvm
* replace hardcoded path to lvm devices file by one actually used by lvm
* does some minor cleanup, especially reuse existing constant pointing to lvm executable instead of defining it again in each module using lvm command

### DIFF
--- a/lib/vdsm/storage/lvmconf.py
+++ b/lib/vdsm/storage/lvmconf.py
@@ -72,8 +72,6 @@ TODO
 
 """
 
-from __future__ import absolute_import
-
 import logging
 
 from augeas import Augeas
--- a/lib/vdsm/storage/lvmconf.py
+++ b/lib/vdsm/storage/lvmconf.py
@@ -76,7 +76,21 @@ import logging
 
 from augeas import Augeas
 
+from vdsm import constants
+from vdsm.common import commands
+
+
 log = logging.getLogger("storage.lvmconf")
+
+
+class UnexpectedLvmConfigOutput(Exception):
+    msg = "Unexpected LVM config output"
+
+    def __init__(self, reason):
+        self.value = "reason=%s" % reason
+
+    def __str__(self):
+        return "%s: %s" % (self.msg, repr(self.value))
 
 
 class LVMConfig(object):
@@ -184,3 +198,38 @@ class LVMConfig(object):
     def close(self):
         log.debug("Closing LVM configuration %s", self.path)
         self.aug.close()
+
+
+def configured_value(section, option):
+    """
+    Return value configured for the option, taken into account all config
+    files and default configurations.
+    """
+    if not section:
+        raise ValueError("Section must not be empty.")
+    if not option:
+        raise ValueError("Option must not be empty.")
+
+    cmd = [constants.EXT_LVM, "lvmconfig",
+           "--typeconfig", "full",
+           f"{section}/{option}"]
+    out = commands.run(cmd).decode("utf-8").strip()
+
+    # Validate, that the output is in expected format key=value and doesn't
+    # span multiple lines.
+    if "=" not in out or "\n" in out:
+        raise UnexpectedLvmConfigOutput(
+            f"Unexpected output: option={section}/{option}, output={out}")
+
+    key, value = out.split("=", 1)
+
+    # Validate, that returned key matches requested option.
+    if key.strip() != option:
+        raise UnexpectedLvmConfigOutput(
+            f"Returned key doesn't match option: option={option}, key={key}")
+
+    # String value.
+    if value[0] == '"' and value[-1] == '"':
+        value = value[1:-1]
+
+    return value
--- a/lib/vdsm/storage/lvmdevices.py
+++ b/lib/vdsm/storage/lvmdevices.py
@@ -34,12 +34,13 @@ import logging
 import os
 import subprocess
 
+from vdsm import constants
 from vdsm.common import cmdutils
 from vdsm.common import commands
 from vdsm.storage import lvmconf
 from vdsm.storage import lvmfilter
 
-LVM = "/usr/sbin/lvm"
+
 _LVM_SYSTEM_DEVICES_PATH = "/etc/lvm/devices/system.devices"
 
 
@@ -126,7 +127,7 @@ def _run_vgimportdevices(vg):
     devices won't be imported. If the filter is wrong, we may miss some
     devices. To avoid such situation, set the filter to enable all the devices.
     """
-    cmd = [LVM,
+    cmd = [constants.EXT_LVM,
            'vgimportdevices',
            vg,
            '--config',
@@ -154,7 +155,7 @@ def _run_check():
     raise any exception if the check finds issues in devices file, but only
     log a waring with found issues.
     """
-    cmd = [LVM, 'lvmdevices', "--check"]
+    cmd = [constants.EXT_LVM, 'lvmdevices', "--check"]
 
     p = commands.start(
         cmd,
--- a/lib/vdsm/storage/lvmdevices.py
+++ b/lib/vdsm/storage/lvmdevices.py
@@ -41,7 +41,7 @@ from vdsm.storage import lvmconf
 from vdsm.storage import lvmfilter
 
 
-_LVM_SYSTEM_DEVICES_PATH = "/etc/lvm/devices/system.devices"
+_LVM_DEVICES_DIR = "/etc/lvm/devices"
 
 
 log = logging.getLogger("lvmdevices")
@@ -98,7 +98,13 @@ def _devices_file_exists():
     Returns True if default lvm devices file exists. If the file doesn't
     exists, lvm disables whole devices file functionality.
     """
-    return os.path.exists(_LVM_SYSTEM_DEVICES_PATH)
+    try:
+        devicesfile = lvmconf.configured_value("devices", "devicesfile")
+    except (cmdutils.Error, lvmconf.UnexpectedLvmConfigOutput):
+        return False
+
+    devices_file = os.path.join(_LVM_DEVICES_DIR, devicesfile)
+    return os.path.exists(devices_file)
 
 
 def _configure_devices_file(enable=True):
--- a/tests/storage/lvmconf_test.py
+++ b/tests/storage/lvmconf_test.py
@@ -25,6 +25,8 @@ from __future__ import division
 import os
 import pytest
 
+from vdsm.common import cmdutils
+
 from vdsm.storage import lvmconf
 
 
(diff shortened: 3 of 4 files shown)


vjuranek:
Verified manaully:

Running config on already configured host:
```
[root@localhost ~]# lvmconfig --typeconfig full devices/devicesfile
devicesfile="system.devices"
[root@localhost ~]# vdsm-tool config-lvm-filter -y
Analyzing host...
LVM devices already configured for vdsm
```
Changing lvm config to use some other devices file and running `vdsm-tool` again:
```
[root@localhost ~]# lvmconfig --typeconfig full devices/devicesfile
devicesfile="test"
[root@localhost ~]# vdsm-tool config-lvm-filter -y
Analyzing host...
Found these mounted logical volumes on this host:

  logical volume:  /dev/mapper/cs-root
  mountpoint:      /
  devices:         /dev/vda2

  logical volume:  /dev/mapper/cs-swap
  mountpoint:      [SWAP]
  devices:         /dev/vda2

Configuring LVM system.devices.
Devices for following VGs will be imported:
    
 cs

Configuration completed successfully!

Please reboot to verify the configuration.
    
[root@localhost ~]# cat /etc/lvm/devices/test 
# LVM uses devices listed in this file.
# Created by LVM command vgimportdevices pid 3120 at Mon Feb  7 11:40:03 2022
VERSION=1.1.1
IDTYPE=devname IDNAME=/dev/vda2 DEVNAME=/dev/vda2 PVID=ITbnR9fozXVmq5e6iz8WhQxfGqR4pdrd PART=2
```

Running config again:
```
[root@localhost ~]# vdsm-tool config-lvm-filter -y
Analyzing host...
LVM devices already configured for vdsm
```

Changing back to lvm default:
```
[root@localhost ~]# lvmconfig --typeconfig full devices/devicesfile
devicesfile="system.devices"
[root@localhost ~]# vdsm-tool config-lvm-filter -y
Analyzing host...
LVM devices already configured for vdsm
```

lib/vdsm/storage/lvmconf.py
```diff
@@ -184,3 +189,27 @@ def save(self):
     def close(self):
         log.debug("Closing LVM configuration %s", self.path)
         self.aug.close()
+
+
+def configured_value(option):
```
nirs, line 194: I guess the user needs to specify now the string "section/option"?

Should we use section, option arguments?

One advantage is that we don't have to handle the case of reading the section. We can do:

    def configured_value(section, option):
        if not section:
            raise ValueError("Empty section")
        if not option:
            raise ValueError("Empty option")

And developer errors will be reported very clearly.

Another option is to use a string, and validate that the string always has
one "/", but this may assume too much about options name, maybe lvm will support
options with "/".

lib/vdsm/storage/lvmconf.py
```diff
@@ -184,3 +189,27 @@ def save(self):
...
+    cmd = [constants.EXT_LVM, "lvmconfig", "--typeconfig", "full", option]
+    out = commands.run(cmd).decode("utf-8").strip()
+
+    if "=" not in out or "\n" in out:
+        raise se.UnexpectedLvmConfigOutput(option, out)
+    else:
+        value = out.split("=")[1]
+        # String value.
+        if value.startswith('"'):
+            return value[1:len(value) - 1]
```
nirs, line 212: Maybe:

    if value[0] == '"' and value[-1] == '"]:
        value = value[1:-1]

tests/storage/lvmconf_test.py
```diff
@@ -220,6 +223,26 @@ def test_real_lvm_conf():
...
+def test_non_existing_config_value():
+    with pytest.raises(cmdutils.Error) as e:
+        lvmconf.configured_value("does/not/exists")
+    assert e.value.rc == 5
+
+
+def test_query_config_section():
+    with pytest.raises(se.UnexpectedLvmConfigOutput) as e:
+        val = lvmconf.configured_value("devices")
+        print("val=%s" % val)
```
nirs, line 242: This print will never run and it it runs because the code did not raise, the outupt
is little bit confusing. Maybe:

   print(f"unexpected response: {val}")

lib/vdsm/storage/lvmconf.py
```diff
@@ -184,3 +189,27 @@ def save(self):
...
+    returns multiple configuration values, an exception will be thrown.
+    When option is not found, command fails and exception is thrown directly
+    from commands module.
+    """
+    cmd = [constants.EXT_LVM, "lvmconfig", "--typeconfig", "full", option]
+    out = commands.run(cmd).decode("utf-8").strip()
+
+    if "=" not in out or "\n" in out:
+        raise se.UnexpectedLvmConfigOutput(option, out)
+    else:
```
nirs, line 208: else not needed, we just raised, so we can continue with the happy path.

lib/vdsm/storage/lvmconf.py
```diff
@@ -184,3 +189,27 @@ def save(self):
...
+        value = out.split("=")[1]
```
nirs, line 209: should we validate the key name? it should match our query, right?

lib/vdsm/storage/lvmdevices.py
```diff
@@ -98,7 +99,13 @@ def _devices_file_exists():
     Returns True if default lvm devices file exists. If the file doesn't
     exists, lvm disables whole devices file functionality.
     """
-    return os.path.exists(_LVM_SYSTEM_DEVICES_PATH)
+    try:
+        conf_devices_file = lvmconf.configured_value("devices/devicesfile")
```
nirs, line 103: Minor issue - it would be little bit nicer if we keep the name of the option:

    devicesfile = lvmconf.configured_value("devices/devicesfile")

    path = os.path.join(_LVM_SYSTEM_DEVICES_DIR, devicesfile)

nirs (changes requested):
Nice!

tests/storage/lvmconf_test.py
vjuranek: ops, this is forgotten debug print out, should be removed

lib/vdsm/storage/lvmdevices.py
vjuranek: done

lib/vdsm/storage/lvmdevices.py
```diff
@@ -37,11 +37,12 @@
 from vdsm import constants
 from vdsm.common import cmdutils
 from vdsm.common import commands
+from vdsm.storage import exception as se
 from vdsm.storage import lvmconf
 from vdsm.storage import lvmfilter
 
 
-_LVM_SYSTEM_DEVICES_PATH = "/etc/lvm/devices/system.devices"
+_LVM_SYSTEM_DEVICES_DIR = "/etc/lvm/devices"
```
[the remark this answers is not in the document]
vjuranek: `lvmconfig` returns just the configured value, and AFAIK you can configure only devicesfile name relative to `/etc/lvm/devices`. So I guess therefore there's no option to get full path.

lib/vdsm/storage/exception.py
```diff
@@ -1676,6 +1676,15 @@ def __init__(self, pvs, vgname):
 # NOTE: code is 621 used by LVMCommandError above.
 
 
+class UnexpectedLvmConfigOutput(StorageException):
+    code = 622
+    msg = "Unexpected LVM config output. Only single config value can be " \
+          "queried."
```
nirs, line 1682: Looking at the code and the places you raise this error, this message:

    Only single config value can be queried

Does not look relevant. Maybe this is a leftover from previous version?

Also, this is raised from lvmconf, running always as root from vdsm-tool. Do we really need a public error for this or we can use a module error, used only for
logging error in vdsm-tool?

lib/vdsm/storage/lvmconf.py
```diff
@@ -184,3 +189,38 @@ def save(self):
...
+           f"{section}/{option}"]
+    out = commands.run(cmd).decode("utf-8").strip()
+
+    # Validate, that the output is in expected format key=value and doesn't
+    # span multiple lines.
+    if "=" not in out or "\n" in out:
+        raise se.UnexpectedLvmConfigOutput(
+            f"Unexpected output: option={section}/{option}, output={out}")
+
+    key, value = out.split("=")
```
nirs, line 215: We want to split exactly once:

    split("=", 1)

Otherwise if the value includes "=" this call will fail:

    >>> key, value = "key=val=ue".split("=") 
      File "<stdin>", line 1, in <module>
    ValueError: too many values to unpack (expected 2)

lib/vdsm/storage/lvmconf.py
```diff
@@ -184,3 +189,38 @@ def save(self):
     def close(self):
         log.debug("Closing LVM configuration %s", self.path)
         self.aug.close()
+
+
+def configured_value(section, option):
+    """
+    Return value configured for the option, taken into account all config
+    files or default configurations.
```
nirs, line 197: I think "and" is more precise, because lvm must consider both config files and the defaults.

tests/storage/lvmconf_test.py
```diff
@@ -220,6 +222,29 @@ def test_real_lvm_conf():
...
+
+def test_configured_value_empty_section():
+    with pytest.raises(ValueError):
+        lvmconf.configured_value("", "dir")
+
+
+def test_configured_value_empty_option():
+    with pytest.raises(ValueError):
+        lvmconf.configured_value("devices", "")
+
```
nirs, line 246: The fact that you don't test the special new error tells that it is not 
that important and we don't need a new public error.

lib/vdsm/storage/lvmdevices.py
nirs, line 45: A better name is _LVM_DEVICES_DIR.

nirs (changes requested):
Looks good, but there are some rough edges.

lib/vdsm/storage/lvmdevices.py
vjuranek: done

lib/vdsm/storage/exception.py
vjuranek: Sure, it can be module exception. Done.

lib/vdsm/storage/lvmconf.py
vjuranek: done

lib/vdsm/storage/lvmconf.py
vjuranek: done

tests/storage/lvmconf_test.py
vjuranek: done

nirs (approved):

PR #71, opened 2022-02-09
updated url for vdsm-api.yml

kottapar:
The current url is giving a 404 as this file is moved to lib/vdsm/api. This PR is to correct the URL only


vjuranek (approved):

vjuranek:
Merged, thanks!

PR #74, opened 2022-02-09
Allow selecting OST suite and distro

tinez:
This PR adds the logic for parsing the parameters to '/ost'
comments that allow specifying which suite and distro to use
in the OST run, i.e.:

 /ost network-suite-master rhel8

### DIFF
--- a/.github/workflows/ost.yml
+++ b/.github/workflows/ost.yml
@@ -5,7 +5,19 @@ on:
     types: [created]
 
 jobs:
+  parse-params:
+    outputs:
+      ost_suite: ${{ steps.parse-params.outputs.ost_suite }}
+      ost_images_distro: ${{ steps.parse-params.outputs.ost_images_distro }}
+    steps:
+    - id: parse-params
+      run: |
+        echo "::set-output name=ost_suite::$(echo ${{ github.event.comment.body }} | egrep -o '\w+-suite-master' || echo basic-suite-master)"
+      run: |
+        echo "::set-output name=ost_images_distro::$(echo ${{ github.event.comment.body }} | egrep -o '(el8stream|el9stream|rhel8)' || echo el8stream)"
+
   trigger-ost:
+    needs: parse-params
     if: |
       github.event.issue.pull_request &&
       startsWith(github.event.comment.body, '/ost') &&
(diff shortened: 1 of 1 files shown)


.github/workflows/ost.yml
```diff
@@ -5,7 +5,19 @@ on:
...
+  parse-params:
+    outputs:
+      ost_suite: ${{ steps.parse-params.outputs.ost_suite }}
+      ost_images_distro: ${{ steps.parse-params.outputs.ost_images_distro }}
+    steps:
+    - id: parse-params
+      run: |
+        echo "::set-output name=ost_suite::$(echo ${{ github.event.comment.body }} | egrep -o '\w+-suite-master' || echo basic-suite-master)"
+      run: |
+        echo "::set-output name=ost_images_distro::$(echo ${{ github.event.comment.body }} | egrep -o '(el8stream|el9stream|rhel8)' || echo el8stream)"
```
nirs, line 17: github yaml syntax is a good programming environment. If you want to have some
logic in ci script, write proper python script to parse it.

Also this is needed for all projects, no? why vdsm should keep special code for ost?

nirs (changes requested):
This user interface is confusing and hard to use. We need named arguments like:

    /ost suite=xxx distro=yyy

When used incorrectly the script should write a usage message with
all the available options.

.github/workflows/ost.yml
tinez: If the logic was kept in a python script we'd need to create it inline in the job definition.

We still don't have a solution for reusing this logic across oVirt projects. I'll be happy to move this code once someone figures out how to do this.

vjuranek (approved):

michalskrivanek (approved):
not perfect, sure, but lgtm

i think we can do the parsing and whatnot in the actual trigger code (in here - https://github.com/oVirt/ovirt-system-tests/blob/master/.github/workflows/ost.yaml#L45). It will run on the runner, so potentially slower response in case of issues...

.github/workflows/ost.yml
mz-pdm: > We still don't have a solution for reusing this logic across oVirt projects.

As announced by Michal, we have one now, so this PR is not needed anymore. And we can even simplify Vdsm ost.yml workflow, based on https://github.com/oVirt/ovirt-system-tests/blob/master/.github/example-trigger.yaml, right?

nirs:
@tinez Is this still needed?

tinez:
Implemented externally

PR #75, opened 2022-02-14
vdsm-tool: Create libvirtd conf file for el9

hbraha:
In el9, we use the monolithic daemon libvirtd,
so we need to create a libvirtd config manually
since it's not shipped by deafult.


lib/vdsm/tool/configurators/libvirt.py
```diff
@@ -42,6 +42,7 @@
 from vdsm import constants
 
 
+_LIBVIRT_CONF_PATH = "/etc/sysconfig/libvirtd"
```
tinez, line 45: We already have that path defined under the `LDCONF` key (line ~360) in a better way - it uses the `constants` module. Let's move that definition here, and reuse the new variable in `LDCONF`

lib/vdsm/tool/configurators/libvirt.py
```diff
@@ -99,6 +100,9 @@ def isconfigured():
     """
     Check if libvirt is already configured for vdsm
     """
+    # Creates libvirt conf file if not exists
```
tinez, line 103: Strictly speaking, this is not really "libvirt configration" - this one can be found in `/etc/libvirt/libvirt.conf`. Let's call it "system-level libvirt config". This applies to the commit message too.

tinez (changes requested):

tinez (approved):

almusil (approved):

PR #79, opened 2022-02-17
spec: point to src URL

sandrobonazzola:
Point to the tarball included in the release archive

### DIFF
--- a/vdsm.spec.in
+++ b/vdsm.spec.in
@@ -63,7 +63,7 @@ Summary:        Virtual Desktop Server Manager
 Group:          Applications/System
 License:        GPLv2+
 Url:            http://www.ovirt.org/develop/developer-guide/vdsm/vdsm/
-Source0:        %{vdsm_name}-%{version}.tar.gz
+Source0:        https://github.com/oVirt/vdsm/releases/download/v%{version}/%{vdsm_name}-%{version}.tar.gz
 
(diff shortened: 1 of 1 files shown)


mz-pdm (approved):

vjuranek (approved):

tinez (dismissed):

nirs (changes requested):
Please rebase on master,  you have a merge commit in your pr, and we don't want them in git history.

michalskrivanek:
so why is squash&merge disabled? that would be easy for the maintainer to do without more ping pong

nirs:
We don’t want to squash patches. We can rebase and merge if needed, but this pr was wrongly updated so it must be pushed again without the merge commit.

sandrobonazzola:
> Please rebase on master, you have a merge commit in your pr, and we don't want them in git history.

Done

nirs:
Thanks @sandrobonazzola!

nirs (approved):

nirs:
@michalskrivanek we cannot merge this again. Previously vdsm was
configured so we could rebase and merge if needed, but now the
it does not work and we don't have access to the project settings.
This is not productive. We need full access to the project settings.

nirs:
@michalskrivanek found that "rebase" option was hidden in the "update" menu, so we
can handle the rebase issues.

PR #80, opened 2022-02-21
Improve extend timing

nirs:
Include the time since the the block threshold event was received until
we handled it in the extend timing log.

Based on the timing info we can tune the volume extension parameters to
minimize vm pauses, both in vdsm defaults and in the fields in the defaults
need to be changed for a particular environment.

To improve the timing info, vdsm.comon.time.Clock was modified to allow
timing since a time in the past.

Bug-Url: https://bugzilla.redhat.com/2051997

PR #81, opened 2022-02-21
tests: Test drive exceeded time behavior

nirs:
The way we set and clear exceeded time is little too delicate. Lets
protect with it with a test.

The new test added as simple function since the relevant class was was
not converted properly to pytest, and we cannot use pytest fixtures in
the class. Converting the module to pytest properly will have to be done
later.

Completes #80.

### DIFF
--- a/tests/virt/vmstorage_test.py
+++ b/tests/virt/vmstorage_test.py
@@ -21,6 +21,7 @@
 from __future__ import absolute_import
 from __future__ import division
 
+import logging
 import os
 import xml.etree.ElementTree as etree
 
@@ -36,6 +37,7 @@ from testlib import namedTemporaryDir
 from testlib import permutations, expandPermutations
 
 from vdsm.common import exception
+from vdsm.common import time
 from vdsm.common.units import MiB, GiB
 from vdsm.common import xmlutils
 from vdsm import utils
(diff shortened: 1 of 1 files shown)


tests/virt/vmstorage_test.py
```diff
@@ -617,6 +621,27 @@ def test_block_threshold_stale_path(self):
...
+    drive = Drive(log, **conf)
+
+    # Exceeded time not set yet.
+    assert drive.exceeded_time is None
+
+    # Setting threshold state does not set exceeded time.
+    drive.threshold_state = BLOCK_THRESHOLD.SET
+    assert drive.exceeded_time is None
+
+    # Getting threshold event set exceeded time.
```
barpavel, line 635: `set` --> `sets`

tests/virt/vmstorage_test.py
```diff
@@ -617,6 +621,27 @@ def test_block_threshold_stale_path(self):
...
+    monkeypatch.setattr(time, "monotonic_time", lambda: 123.0)
+    drive.on_block_threshold("/path")
+    assert drive.exceeded_time == 123.0
+
+    # Chanding threshold clears exceeded time.
```
barpavel, line 640: `Chanding` --> `Changing`.

vjuranek (commented):
LGTM but would be nice to fix the tipos found by Pavel

nirs:
@barpavel @vjuranek thanks for reviewing, typos fixed in new version.

PR #82, opened 2022-02-22
config: Increase thin provisioning threshold

nirs:
Collecting extend stats shows that extend takes between 2.2 to 6.2
seconds, with average of 3.7 seconds. With the default thresholds:

[irs]
volume_utilization_chunk_mb = 1024
volume_utilization_percent = 50

This means that we extend the volume when free space is 512 MiB. Writing
more than 512 MiB in 3.7 seconds (138.4 MiB/s) will cause the VM to
pause with ENOSPC.

This configuration was too low 10 years ago, and we need to update it
for modern storage. Update the values to allow 4 times faster writes
before we pause with ENOSPC.

With the new configuration:

[irs]
volume_utilization_chunk_mb = 2560
volume_utilization_percent = 20

We extend the volume when free space is 2048 MiB.

Testing with old and new configuration show that we can cope now with 4x
times faster write rate before we VMs pause during extend.

Before:

write rate  extends   pauses
----------------------------
 75 MiB/s        50        0
100 MiB/s        50        4
125 MiB/s        50        4
150 MiB/s        53       24

After:

write rate  extends   pauses
----------------------------
200 MiB/s        20        0
250 MiB/s        20        0
300 MiB/s        20        0
350 MiB/s        21        0
400 MiB/s        20        1
450 MiB/s        20        2
500 MiB/s        22        7
550 MiB/s        23        7

The downside of this change is allocating more space in the storage
domain. New empty disk will consume 2.5 GiB instead of 1 GiB.

Bug-Url: https://bugzilla.redhat.com/2051997

### DIFF
--- a/lib/vdsm/common/config.py.in
+++ b/lib/vdsm/common/config.py.in
@@ -315,12 +315,14 @@ parameters = [
 
         ('irsd', '%(images)s/irsd', None),
 
-        ('volume_utilization_percent', '50',
+        ('volume_utilization_percent', '20',
             'Together with volume_utilization_chunk_mb, set the minimal free '
             'space before a thin provisioned block volume is extended. Use '
-            'lower values to extend earlier.'),
+            'lower values to extend earlier. With the default value (20) the'
+            'system will extend the volume when free space is 80% of '
+            'volume_utilization_chunk_mb (2 GiB).'),
 
-        ('volume_utilization_chunk_mb', '1024',
+        ('volume_utilization_chunk_mb', '2560',
             'Size of extension chunk in megabytes, and together with '
             'volume_utilization_percent, set the free space limit. Use higher '
             'values to extend in bigger chunks.'),
--- a/tests/storage/blockvolume_test.py
+++ b/tests/storage/blockvolume_test.py
@@ -26,7 +26,6 @@ from contextlib import contextmanager
 import pytest
 
 from vdsm.common.units import MiB, GiB
-from vdsm.config import config
 from vdsm.storage import blockVolume
 from vdsm.storage import constants as sc
 from vdsm.storage import exception as se
@@ -49,7 +48,10 @@ from testlib import permutations, expandPermutations
 from testlib import VdsmTestCase
 
 
-CONFIG = make_config([('irs', 'volume_utilization_chunk_mb', '1024')])
+CONFIG = make_config([
+    ('irs', 'volume_utilization_chunk_mb', '1024'),
+    ('irs', 'volume_utilizzation_precent', '50'),
+])
 
 
 @expandPermutations
@@ -73,7 +75,7 @@ class TestBlockVolumeSize(VdsmTestCase):
         # Sparse, cow, capacity config.volume_utilization_chunk_mb - 1,
         # No initial size.
         # Expected 1024 MiB allocated (config.volume_utilization_chunk_mb)
-        [(sc.SPARSE_VOL, sc.COW_FORMAT, (config.getint(
+        [(sc.SPARSE_VOL, sc.COW_FORMAT, (CONFIG.getint(
             "irs", "volume_utilization_chunk_mb") - 1) * MiB, None), GiB],
         # Sparse, cow, capacity 4 GiB, initial size 952320 B.
         [(sc.SPARSE_VOL, sc.COW_FORMAT, 4 * GiB, 952320),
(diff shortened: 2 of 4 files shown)


bennyz (approved):

nirs:
Some storage tests fail, I guess some wrong tests assumes the old chunk size instead of 
using mocked config.

michalskrivanek:
can you elaborate a little bit of what are the factors affecting the time it takes to extend? I mean besides the actual storage speed. Are there any waits in the process, communication with SPM, etc?

ovirt-infra:
> can you elaborate a little bit of what are the factors affecting the time it takes to extend? I mean besides the actual storage speed. Are there any waits in the process, communication with SPM, etc?

also, is the write speed as perceived by the guest directly corresponding to the actual physical write speed? w're using O_DIRECT everywhere so I would assume yes? So we can roughly estimate the minimal values for a concrete underlying write speed? e.g. if we try to measure the write speed outside of oVirt with fio or even dd or something...

michalskrivanek:
also, is the write speed as perceived by the guest directly corresponding to the actual physical write speed? w're using O_DIRECT everywhere so I would assume yes? So we can roughly estimate the minimal values for a concrete underlying write speed? e.g. if we try to measure the write speed outside of oVirt with fio or even dd or something...

nirs:
> can you elaborate a little bit of what are the factors affecting the time it takes to extend? I mean besides the actual storage speed. Are there any waits in the process, communication with SPM, etc?

The time include lot of waiting since we use polling.

- libvirt event thread get a block threshold event, and mark the drive 
  for extension
- The periodic watermark monitor checks vms every 2 seconds. When it find
  that a drive needs extension, it sends a request to the spm by writing
  to the storage mailbox
- The SPM check storage every 2 seconds. When it find the request, it
  run the extend using the spm mailbox thread pool
- The host is polling its mailbox for replies every 2 seconds. When it detects
  the reply, it complete the extend on the host side and resume the vm if
  needed.

We cannot optimize much the storage mailbox since checking the mailbox requires
reading from storage, and many hosts may check the mailbox at the same time.
Maybe we can check every 1 second instead of 2.

We can optimize the block threshold event handling - it should really post an
event that will wake up the watermark monitor immediately, and start the extend
flow. This will save 0-2 seconds from the total time. This requires rewrite
of the watermark monitor and separating it from the periodic executor which
is a change I wanted to do for long time.

vjuranek (approved):

vjuranek (changes requested):
linter says that `from vdsm.config import config` in `blockvolume_test` is unsed now and fails

nirs:
> linter says that `from vdsm.config import config` in `blockvolume_test` is unsed now and fails

flake8 is correct, fixed in current version.

nirs:
> > Measuring write throughput should be done in the guest, there is big difference
> > between guest performance and host performance. Measuring is complected and it
> > is impossible to predict how fast the disk will grow on a system.
> 
> yes, but isn't it just that it's always somewhat slower in guest, never faster? In that case it is still useful as then we can measure easily on the host and estimate for the worst case of near-host performance in guest

I don't see how it can be faster in the guest. But it is not possible to measure the host
and assume that the measurement is correct for the future, because the load on the storage,
the network and the host at the time of the measurement can be different at the time the
guest use the storage.

nirs:
@vjuranek should be ready now.

michalskrivanek:
4x is a big improvement.
The extra overhead is small enough, 1.5GB extra for today's SD sizes is worth it.

Still, with a bit over 6 seconds it means we will still eventually pause with >300MiB/s writes even in quiet conditions. Is it high enough? Should we go for 500 increasing the overhead? OTOH it's really not a big deal to pause occasionally, we just want to avoid frequent pauses every time e.g. etcd is extended.

@mykaul FYI

mykaul:
> 4x is a big improvement. The extra overhead is small enough, 1.5GB extra for today's SD sizes is worth it.
> 
> Still, with a bit over 6 seconds it means we will still eventually pause with >300MiB/s writes even in quiet conditions. Is it high enough? Should we go for 500 increasing the overhead? OTOH it's really not a big deal to pause occasionally, we just want to avoid frequent pauses every time e.g. etcd is extended.
> 
> @mykaul FYI

Looks good - time to change the defaults indeed.

vjuranek (approved):

PR #83, opened 2022-02-22
network: Run network tests every week after new containers are built

almusil:
By running tests after the container is built we can find any possible
regression caused by newer packages in the container.


nirs (approved):
Nice

PR #86, opened 2022-02-28
Support adding a bitmap to new volume

nirs:
When creating a new volume or a snapshot, caller can specify
bitmap={bitmap-uuid} to create a new empty bitmap in the new volume.
    
This will enable hybrid backup, when every backup is started by creating
a temporary snapshot. The new bitmap records the changes to the volume
since the snapshot was created. When the temporary snapshot will be
deleted, the new bitmap will be copied to the parent volume, and will be
used for the next incremental backup.
    
Like add_bitmaps, we treat creating a new bitmap as best effort. If the
operation fails we don't fail the creation of the volume, but the next
backup using this bitmap will have to be a full backup.

Not tested yet, but tests pass.

Bug-Url: https://bugzilla.redhat.com/2053669

### DIFF
--- a/lib/vdsm/storage/fileVolume.py
+++ b/lib/vdsm/storage/fileVolume.py
@@ -446,7 +446,7 @@ class FileVolume(volume.Volume):
     @classmethod
     def _create(cls, dom, imgUUID, volUUID, capacity, volFormat, preallocate,
                 volParent, srcImgUUID, srcVolUUID, volPath, initial_size=None,
-                add_bitmaps=False):
+                add_bitmaps=False, bitmap=None):
         """
         Class specific implementation of volumeCreate.
         """
@@ -456,7 +456,8 @@ class FileVolume(volume.Volume):
         else:
             return cls._create_cow_volume(
                 dom, volUUID, capacity, volPath, initial_size, volParent,
-                imgUUID, srcImgUUID, srcVolUUID, add_bitmaps)
+                imgUUID, srcImgUUID, srcVolUUID, add_bitmaps=add_bitmaps,
+                bitmap=bitmap)
 
     @classmethod
     def _create_raw_volume(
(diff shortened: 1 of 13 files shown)


tests/storage/blocksd_test.py
```diff
@@ -1079,7 +1079,148 @@ def test_create_with_bitmaps(
...
+
+    sdCache.knownSDs[sd_id] = blockSD.findDomain
+    sdCache.manuallyAddDomain(dom)
+
+    dom.refresh()
+    dom.attach(tmp_repo.pool_id)
+
+    img_id = str(uuid.uuid4())
+    base_id = str(uuid.uuid4())
+    base_capacity = GiB
```
barpavel, line 1108: Arguably better (here & lines 1196, 1309):
`1 * GiB`

tests/storage/localfssd_test.py
```diff
@@ -1267,7 +1267,99 @@ def test_create_volume_with_bitmaps(user_domain, local_fallocate):
...
-def test_failed_to_add_bitmaps_to_v3_domain(user_domain, local_fallocate):
+def test_create_snapshot_with_new_bitmap(user_domain, local_fallocate):
+    if user_domain.getVersion() == 3:
+        pytest.skip("Bitmaps operations not supported in v3 domains")
+
+    img_id = str(uuid.uuid4())
+    base_id = str(uuid.uuid4())
+    top_id = str(uuid.uuid4())
+
+    # Create base volume
```
barpavel, line 1278: Minor: missing dots at the end of the comments in this an the next tests.

barpavel (commented):
LGTM except some small stuff.

lib/vdsm/storage/fileVolume.py
```diff
@@ -500,7 +501,7 @@ def _create_raw_volume(
     @classmethod
     def _create_cow_volume(
             cls, dom, vol_id, capacity, vol_path, initial_size, vol_parent,
-            img_id, src_img_id, src_vol_id, add_bitmaps):
+            img_id, src_img_id, src_vol_id, add_bitmaps=False, bitmap=None):
```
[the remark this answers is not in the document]
nirs: Yes, this fixes the code adding this option, but could be done separately.

tests/storage/blocksd_test.py
nirs: If we had several size and using 1 * GiB helps to align the values, yes.

mz-pdm (approved):

vjuranek (dismissed):

bennyz (approved):
works with the engine patches

vjuranek (approved):

barpavel (approved):

PR #89, opened 2022-03-02
managedvolume: create link for MBS devices

bennyz:
The current code for attaching MBS devices tries to construct stable
path. This works for RBD and mpath devices, however doesn't scale for
drivers that do not fall in this category, for instance, NVMe/TCP
drivers.

This patch create a link for the path returned by os-brick, this link
will be managed by vdsm and will be used by engine as the path to run
VMs.

Link example:
```shell
ll /var/run/vdsm/managedvolumes/fa728e08-2db4-4016-bfbb-bd75f2ee1735_876fa465-d92b-4070-833d-c5ec4331ceb5
lrwxrwxrwx. 1 vdsm kvm 60 May  3 10:21 /var/run/vdsm/managedvolumes/fa728e08-2db4-4016-bfbb-bd75f2ee1735_876fa465-d92b-4070-833d-c5ec4331ceb5 -> /dev/rbd/volumes/volume-876fa465-d92b-4070-833d-c5ec4331ceb5
```

Snippet for VM XML:
```xml
<disk type='block' device='disk' snapshot='no'>
      <driver name='qemu' type='raw' cache='none'/>
      <source dev='/run/vdsm/managedvolumes/fa728e08-2db4-4016-bfbb-bd75f2ee1735_876fa465-d92b-4070-833d-c5ec4331ceb5' index='1'>
        <seclabel model='dac' relabel='no'/>
      </source>
      <backingStore/>
      <target dev='sda' bus='scsi'/>
      <serial>876fa465-d92b-4070-833d-c5ec4331ceb5</serial>
      <boot order='1'/>
      <alias name='ua-876fa465-d92b-4070-833d-c5ec4331ceb5'/>
      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
</disk>
```

Rule content
```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_fa728e08-2db4-4016-bfbb-bd75f2ee1735_d1530bb1-8b92-40e8-9d5b-1adbcbc4eedc.rules
SYMLINK=="rbd/volumes/volume-d1530bb1-8b92-40e8-9d5b-1adbcbc4eedc", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
```

```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_fc3b802e-cbc5-4896-995d-87f4b25a1a68_0d850fc9-9dfe-44ee-a4f6-a209d1f4bf7b.rules
SYMLINK=="mapper/20024f4005854000a", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
```

### DIFF
--- a/lib/vdsm/API.py
+++ b/lib/vdsm/API.py
@@ -1820,8 +1820,8 @@ class ManagedVolume(APIBase):
 
     @api.logged(on="api.storage")
     @api.method
-    def detach_volume(self, vol_id):
-        return managedvolume.detach_volume(vol_id)
+    def detach_volume(self, vol_id, sd_id):
+        return managedvolume.detach_volume(sd_id, vol_id)
 
     @api.logged(on="api.storage")
     @api.method
--- a/lib/vdsm/API.py
+++ b/lib/vdsm/API.py
@@ -1809,11 +1809,14 @@ class ManagedVolume(APIBase):
 
     @api.logged(on="api.storage")
     @api.method
-    def attach_volume(self, vol_id, connection_info):
+    def attach_volume(self, vol_id, connection_info, sd_id):
         """
         attach volume and return attached device information
         """
-        return managedvolume.attach_volume(vol_id, connection_info)
+        return managedvolume.attach_volume(
+            sd_id,
+            vol_id,
+            connection_info)
 
     @api.logged(on="api.storage")
     @api.method
--- a/lib/vdsm/api/vdsm-api.yml
+++ b/lib/vdsm/api/vdsm-api.yml
@@ -9752,6 +9752,11 @@ ManagedVolume.attach_volume:
     -   description: The information needed to attach a volume
         name: connection_info
         type: *ManagedVolumeConnection
+    -   defaultvalue: '00000000-0000-0000-0000-000000000000'
+        added: '4.5.1'
+        description: The storage domain ID
+        name: sd_id
+        type: *UUID
     return:
         description: Information about the attached volume.
         type: *ManagedVolumeInformation
--- a/lib/vdsm/api/vdsm-api.yml
+++ b/lib/vdsm/api/vdsm-api.yml
@@ -1187,9 +1187,18 @@ types:
             name: connection_info
             type: *ManagedVolumeConnection
         -   defaultvalue: null
-            description: The volume path to be used in the VM XML
+            description: A stable volume path constructed by vdsm, used by
+                by engine to run VMs. Superseded by managed_path.
             name: path
             type: string
+            deprecated: '4.5'
+        -   defaultvalue: null
+            description: The volume path to be used for running VMs,
+                         a vdsm controlled link to the path mapped by
+                         os-brick.
+            name: managed_path
+            type: string
+            added: '4.5.1'
         -   defaultvalue: null
             description: The attached volume information
             name: attachment
--- a/lib/vdsm/api/vdsm-api.yml
+++ b/lib/vdsm/api/vdsm-api.yml
@@ -9769,6 +9769,11 @@ ManagedVolume.detach_volume:
     -   description: The volume ID
         name: vol_id
         type: *UUID
+    -   defaultvalue: '00000000-0000-0000-0000-000000000000'
+        added: '4.5.1'
+        description: The storage domain ID
+        name: sd_id
+        type: *UUID
 
 ManagedVolume.volumes_info:
     added: '4.3'
--- a/lib/vdsm/clientIF.py
+++ b/lib/vdsm/clientIF.py
@@ -419,9 +419,14 @@ class clientIF(object):
         """
         if type(drive) is dict:
             device = drive['device']
+
+            # Managed drives are prepared in ManagedVolume.attach_volume
+            if drive.get("managed", False):
+                volPath = drive['path']
+
             # PDIV drive format
             # Since version 4.2 cdrom may use a PDIV format
-            if device in ("cdrom", "disk") and isVdsmImage(drive):
+            elif device in ("cdrom", "disk") and isVdsmImage(drive):
                 res = self.irs.prepareImage(
                     drive['domainID'], drive['poolID'],
                     drive['imageID'], drive['volumeID'])
--- a/lib/vdsm/common/udevadm.py
+++ b/lib/vdsm/common/udevadm.py
@@ -53,7 +53,12 @@ def settle(timeout, exit_if_exists=None):
         logging.error("%s", e)
 
 
-def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
+def trigger(
+    attr_matches=(),
+    property_matches=(),
+    subsystem_matches=(),
+    path=None,
+):
     '''
     Request device events from the kernel.
 
@@ -83,6 +88,10 @@ def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
 
                         Causes only events related to specified subsystem to
                         be triggered.
+
+    path                Path to trigger events for. For example:
+                        /dev/mapper/20024f4005854000a
+
     '''
     _run_command(['control', '--reload'])
 
@@ -96,6 +105,9 @@ def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
 
     for name in subsystem_matches:
         cmd.append('--subsystem-match={}'.format(name))
+
+    if path:
+        cmd.append(path)
 
     _run_command(cmd)
 
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -127,7 +127,7 @@ def attach_volume(sd_id, vol_id, connection_info):
 
 
 @requires_os_brick
-def detach_volume(vol_id):
+def detach_volume(sd_id, vol_id):
     """
     Detach volume with os-brick.
     """
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -85,7 +85,7 @@ def connector_info():
 
 
 @requires_os_brick
-def attach_volume(vol_id, connection_info):
+def attach_volume(sd_id, vol_id, connection_info):
     """
     Attach volume with os-brick.
     """
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -114,7 +114,7 @@ def attach_volume(sd_id, vol_id, connection_info):
                         "rbd, iscsi")
 
                 run_link = _add_run_link(sd_id, vol_id, path)
-                _add_udev_rule(vol_id, path)
+                _add_udev_rule(sd_id, vol_id, path)
             except:
                 _silent_detach(connection_info, attachment)
                 raise
@@ -144,7 +144,7 @@ def detach_volume(sd_id, vol_id):
         if "path" in vol_info and os.path.exists(vol_info["path"]):
             run_helper("detach", vol_info)
 
-        _remove_udev_rule(vol_info['vol_id'])
+        _remove_udev_rule(sd_id, vol_info['vol_id'])
         _remove_run_link(sd_id, vol_id)
         db.remove_volume(vol_id)
 
@@ -270,7 +270,7 @@ def _silent_remove(db, sd_id, vol_id):
     except Exception:
         log.exception("Failed to remove managed volume %s from DB", vol_id)
 
-    _remove_udev_rule(vol_id)
+    _remove_udev_rule(sd_id, vol_id)
     _remove_run_link(sd_id, vol_id)
 
 
@@ -286,20 +286,20 @@ def _silent_detach(connection_info, attachment):
         log.exception("Failed to detach managed volume %s", vol_info)
 
 
-def _add_udev_rule(vol_id, path):
+def _add_udev_rule(sd_id, vol_id, path):
     proxy = supervdsm.getProxy()
-    proxy.add_managed_udev_rule(vol_id, path)
+    proxy.add_managed_udev_rule(sd_id, vol_id, path)
     try:
         proxy.trigger_managed_udev_rule(path)
     except:
-        _remove_udev_rule(vol_id)
+        _remove_udev_rule(sd_id, vol_id)
         raise
 
 
-def _remove_udev_rule(vol_id):
+def _remove_udev_rule(sd_id, vol_id):
     try:
         proxy = supervdsm.getProxy()
-        proxy.remove_managed_udev_rule(vol_id)
+        proxy.remove_managed_udev_rule(sd_id, vol_id)
     except Exception:
         log.exception(
             "Failed to remove udev rule for volume %s", vol_id)
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -56,6 +56,7 @@ from vdsm.storage import managedvolumedb
 HELPER = '/usr/libexec/vdsm/managedvolume-helper'
 DEV_MAPPER = "/dev/mapper"
 DEV_RBD = "/dev/rbd"
+VOLUME_LINK_DIR = "/run/vdsm/managedvolumes/"
 
 log = logging.getLogger("storage.managedvolume")
 
@@ -112,18 +113,18 @@ def attach_volume(sd_id, vol_id, connection_info):
                         "Unsupported volume type, supported types are: "
                         "rbd, iscsi")
 
+                run_link = _add_run_link(sd_id, vol_id, path)
                 _add_udev_rule(vol_id, path)
             except:
                 _silent_detach(connection_info, attachment)
                 raise
         except:
-            _silent_remove(db, vol_id)
+            _silent_remove(db, sd_id, vol_id)
             raise
-
     log.debug("Attached volume %s attachment=%s", vol_id, attachment)
 
     return {"result": {'attachment': attachment, 'path': path,
-                       'vol_id': vol_id}}
+                       'vol_id': vol_id, 'managed_path': run_link}}
 
 
 @requires_os_brick
@@ -144,6 +145,7 @@ def detach_volume(sd_id, vol_id):
             run_helper("detach", vol_info)
 
         _remove_udev_rule(vol_info['vol_id'])
+        _remove_run_link(sd_id, vol_id)
         db.remove_volume(vol_id)
 
 
@@ -258,14 +260,18 @@ def _invalidate_lvm_devices(attachment):
         lvm.invalidate_devices()
 
 
-def _silent_remove(db, vol_id):
+def _silent_remove(db, sd_id, vol_id):
     """
-    Remove volume from db during cleanup flow, logging errors.
+    Remove volume from db, udev rule and link during cleanup flow, logging
+    errors.
     """
     try:
         db.remove_volume(vol_id)
     except Exception:
         log.exception("Failed to remove managed volume %s from DB", vol_id)
+
+    _remove_udev_rule(vol_id)
+    _remove_run_link(sd_id, vol_id)
 
 
 def _silent_detach(connection_info, attachment):
@@ -297,3 +303,28 @@ def _remove_udev_rule(vol_id):
     except Exception:
         log.exception(
             "Failed to remove udev rule for volume %s", vol_id)
+
+
+def _add_run_link(sd_id, vol_id, path):
+    _create_run_dir()
+    run_path = _run_link(sd_id, vol_id)
+    os.symlink(path, run_path)
+    return run_path
+
+
+def _remove_run_link(sd_id, vol_id):
+    try:
+        os.remove(_run_link(sd_id, vol_id))
+    except Exception:
+        log.exception("Failed to remove run link for volume %s", vol_id)
+
+
+def _create_run_dir():
+    try:
+        os.mkdir(VOLUME_LINK_DIR)
+    except FileExistsError:
+        pass
+
+
+def _run_link(sd_id, vol_id):
+    return os.path.join(VOLUME_LINK_DIR, f"{sd_id}_{vol_id}")
--- a/lib/vdsm/supervdsm_api/udev.py
+++ b/lib/vdsm/supervdsm_api/udev.py
@@ -17,6 +17,15 @@
 # Refer to the README and COPYING files for full details of the license
 #
 
+"""
+General udev utilities
+
+Example for Managed Block Storage paths that are handled:
+# /dev/mapper/20024f4005854000b
+# /dev/rbd/volumes/volume-d1530bb1-8b92-40e8-9d5b-1adbcbc4eedc
+# /dev/nvme0n3
+"""
+
 from __future__ import absolute_import
 from __future__ import division
 
@@ -48,7 +57,8 @@ _UDEV_RULE_FILE_NAME = os.path.join(
     _UDEV_RULE_FILE_EXT)
 
 _UDEV_RULE_FILE_NAME_MANAGED = os.path.join(
-    _UDEV_RULE_FILE_DIR, _UDEV_RULE_FILE_PREFIX + 'managed_' + '%s' +
+    _UDEV_RULE_FILE_DIR,
+    _UDEV_RULE_FILE_PREFIX + 'managed_' + '%s' + '_' + '%s' +
     _UDEV_RULE_FILE_EXT)
 _UDEV_RULE_FILE_NAME_VFIO = os.path.join(
     _UDEV_RULE_FILE_DIR, _UDEV_RULE_FILE_PREFIX + "iommu_group_%s" +
@@ -115,11 +125,15 @@ def rmAppropriateMultipathRules(thiefId):
 
 
 @expose
-def add_managed_udev_rule(vol_id, path):
-    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % vol_id
-    symlink = os.path.relpath(path, '/dev/')
-    rule = 'SYMLINK=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
-        symlink, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
+def add_managed_udev_rule(sd_id, vol_id, path):
+    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % (sd_id, vol_id)
+    device = os.path.relpath(path, '/dev/')
+    if os.path.islink(path):
+        rule = 'SYMLINK=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
+            device, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
+    else:
+        rule = 'KERNEL=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
+            device, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
     with open(rule_file, "w") as rf:
         _log.debug("Creating rule %s: %r", rule_file, rule)
         rf.write(rule)
@@ -127,12 +141,12 @@ def add_managed_udev_rule(vol_id, path):
 
 @expose
 def trigger_managed_udev_rule(path):
-    _udevTrigger(property_matches=(('DEVLINKS', path),))
+    _udevTrigger(path=path)
 
 
 @expose
-def remove_managed_udev_rule(vol_id):
-    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % vol_id
+def remove_managed_udev_rule(sd_id, vol_id):
+    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % (sd_id, vol_id)
     try:
         os.remove(rule_file)
     except FileNotFoundError:
--- a/tests/storage/managedvolume_test.py
+++ b/tests/storage/managedvolume_test.py
@@ -161,7 +161,7 @@ def test_attach_volume_not_installed_attach(monkeypatch):
     # Simulate missing os_brick.
     monkeypatch.setattr(managedvolume, "os_brick", None)
     with pytest.raises(se.ManagedVolumeNotSupported):
-        managedvolume.attach_volume("vol_id", {})
+        managedvolume.attach_volume("sd_id", "vol_id", {})
 
 
 @requires_root
@@ -172,7 +172,10 @@ def test_attach_volume_ok_iscsi(monkeypatch, fake_os_brick, tmp_db, fake_lvm,
         "driver_volume_type": "iscsi",
         "data": {"some_info": 26}
     }
-    ret = managedvolume.attach_volume("fake_vol_id", connection_info)
+    ret = managedvolume.attach_volume(
+        "fake_sd_id",
+        "fake_vol_id",
+        connection_info)
     path = "/dev/mapper/fakemultipathid"
 
     assert ret["result"]["path"] == path
@@ -211,7 +214,10 @@ def test_attach_volume_ok_rbd(monkeypatch, fake_os_brick, tmp_db, fake_lvm,
         "data": {
             "name": "volumes/volume-fake"
         }}
-    ret = managedvolume.attach_volume("fake_vol_id", connection_info)
+    ret = managedvolume.attach_volume(
+        "fake_sd_id",
+        "fake_vol_id",
+        connection_info)
     path = "/dev/rbd/volumes/volume-fake"
 
     assert ret["result"]["path"] == path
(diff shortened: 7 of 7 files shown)


bennyz:
need to fix the tests

lib/vdsm/storage/managedvolume.py
```diff
@@ -113,17 +115,19 @@ def attach_volume(vol_id, connection_info):
                         "rbd, iscsi")
 
                 _add_udev_rule(vol_id, path)
+                _create_managedvolume_link_dir()
+                run_path = os.path.join(VOLUME_LINK_DIR, vol_id)
+                os.symlink(attachment['path'], run_path)
```
nirs, line 120: I think we need _add_run_link() and _remove_run_link(), same as _add_udev_rule() and _remove_udev_rule().

We also need to handle cleaning up after creating run path failed - currently we leave leftover udev rule.

nirs (changes requested):

lib/vdsm/api/vdsm-api.yml
```diff
@@ -1190,6 +1190,13 @@ types:
             description: The volume path to be used in the VM XML
             name: path
             type: string
+        -   defaultvalue: null
+            description: The volume path to be used for running VMs,
```
vjuranek, line 1194: I'd explain how it's related to `path` property - saying this will be used for running VMs, while `path` descriptions says it's used in VM XML is IMHO confusing. Or probably just update description for `path` property?

bennyz: yes, I'll update `path`'s description

nirs: Can we deprecate "path"? Idealy we have only one path reported, and it is uniform
for all kinds of disks.

I'm also not sure about "run_path", this was the first thing I thought about when
proposing this change, but sounds bit dirty.

Alternative terms:
- disk_path
- volume_path

bennyz: yes, I thought about deprecating it, we still use it in copy though, where it doesn't matter what the path is

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -117,9 +117,9 @@ def rmAppropriateMultipathRules(thiefId):
 @expose
 def add_managed_udev_rule(vol_id, path):
     rule_file = _UDEV_RULE_FILE_NAME_MANAGED % vol_id
-    symlink = os.path.relpath(path, '/dev/')
-    rule = 'SYMLINK=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
-        symlink, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
+    kernel_device = os.path.relpath(path, '/dev/')
+    rule = 'KERNEL=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
```
bennyz, line 121: not sure about this

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -128,6 +128,7 @@ def add_managed_udev_rule(vol_id, path):
 @expose
 def trigger_managed_udev_rule(path):
     _udevTrigger(property_matches=(('DEVLINKS', path),))
+    _udevTrigger(property_matches=(('DEVNAME', path),))
```
bennyz, line 131: not sure about this

lib/vdsm/storage/managedvolume.py
```diff
@@ -144,6 +146,7 @@ def detach_volume(vol_id):
             run_helper("detach", vol_info)
 
         _remove_udev_rule(vol_info['vol_id'])
+        os.remove(os.path.join(VOLUME_LINK_DIR, vol_id))
```
nirs, line 149: Why not use _remove_run_link()?

lib/vdsm/storage/managedvolume.py
```diff
@@ -297,3 +304,28 @@ def _remove_udev_rule(vol_id):
...
+    except:
+        log.exception("Failed to remove run link for volume %s", vol_id)
+
+
+def _create_managedvolume_link_dir():
+    try:
+        os.mkdir(VOLUME_LINK_DIR)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
+            raise
```
nirs, line 329: This was the best way in python 2, but in python 3 we can use:

    try:
        os.mkdir(path)
    except FileExistsError:
        pass

lib/vdsm/storage/managedvolume.py
```diff
@@ -297,3 +304,28 @@ def _remove_udev_rule(vol_id):
...
+    _create_managedvolume_link_dir()
+    run_path = os.path.join(VOLUME_LINK_DIR, vol_id)
+    os.symlink(path, run_path)
+
+    return run_path
+
+
+def _remove_run_link(vol_id):
+    try:
+        os.symlink(os.path.join(VOLUME_LINK_DIR, vol_id))
```
nirs, line 319: Are you sure about this?

lib/vdsm/storage/managedvolume.py
```diff
@@ -297,3 +304,28 @@ def _remove_udev_rule(vol_id):
...
+    except:
+        log.exception("Failed to remove run link for volume %s", vol_id)
+
+
+def _create_managedvolume_link_dir():
```
nirs, line 324: This name is little confusing, why not _create_run_dir()?

nirs (changes requested):

lib/vdsm/supervdsm_api/udev.py
bennyz: no luck with rbd:
```shell
>>> os.readlink("/run/vdsm/managedvolumes/67240876-c837-4b70-8742-604fbecabae1")
'/dev/rbd2'

>>> os.path.relpath('/dev/rbd2', '/dev/')
'rbd2'
```
```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_67240876-c837-4b70-8742-604fbecabae1.rules
SYMLINK=="rbd2", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
$ udevadm trigger
$ ll /dev/rbd2
brw-rw----. 1 root disk 252, 32 Mar  3 15:12 /dev/rbd2
```

With `KERNEL`:
```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_67240876-c837-4b70-8742-604fbecabae1.rules
KERNEL=="rbd2", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
$ udevadm trigger
$ ll /dev/rbd2
brw-rw----. 1 vdsm qemu 252, 32 Mar  3 15:13 /dev/rbd2
```
Needs additional research, I need to test an iscsi driver as well, only checked rbd

lib/vdsm/storage/managedvolume.py
```diff
@@ -297,3 +303,25 @@ def _remove_udev_rule(vol_id):
...
+
+def _add_run_link(vol_id, path):
+    _create_run_dir()
+    run_path = os.path.join(VOLUME_LINK_DIR, vol_id)
+    os.symlink(path, run_path)
+
+    return run_path
+
+
+def _remove_run_link(vol_id):
```
nirs, line 316: Looks like _silent_remove_run_link(), since it consume the error.

Same for _remove_udev_rule().

Using bare "except:" is good when you want to do something in error
flow and raise the original error, as we do in _add_udev_rule().

lib/vdsm/storage/managedvolume.py
```diff
@@ -297,3 +303,25 @@ def _remove_udev_rule(vol_id):
...
+    try:
+        os.remove(os.path.join(VOLUME_LINK_DIR, vol_id))
+    except:
```
nirs, line 319: Better handle only Exception:

    except Exception:

Like _remove_udev_rule()

lib/vdsm/storage/managedvolume.py
```diff
@@ -297,3 +303,25 @@ def _remove_udev_rule(vol_id):
     except Exception:
         log.exception(
             "Failed to remove udev rule for volume %s", vol_id)
+
+
+def _add_run_link(vol_id, path):
+    _create_run_dir()
+    run_path = os.path.join(VOLUME_LINK_DIR, vol_id)
+    os.symlink(path, run_path)
+
```
nirs, line 312: Empty line here is not needed.

lib/vdsm/clientIF.py
```diff
@@ -451,6 +451,8 @@ def prepareVolumePath(self, drive, vmId=None, path=None):
                             # not speicfy the type, we must use 'file'.
                             drive['diskType'] = DISK_TYPE.FILE
                     volPath = res['path']
+            elif drive.get("managed", False):
```
nirs, line 454: Would be nice to add a comment lime we have for GUID

muliby-lb:
Hey, just wondering if there are any updates on this one?

lib/vdsm/supervdsm_api/udev.py
bennyz: Looks like `KERNEL` works for iSCSI as well (tested with kaminario as I didn't have other HW)
```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_10e2fb62-7cf9-4d28-a5d8-1267cc650aba.rules
KERNEL=="dm-2", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
```

bennyz:
> Hey, just wondering if there are any updates on this one?

If you can confirm this works for lightos we can move forward with this, my tests for rbd and iscsi look good so far

muliby-lb:
> > Hey, just wondering if there are any updates on this one?
> 
> If you can confirm this works for lightos we can move forward with this, my tests for rbd and iscsi look good so far

is it possible to test with vdsm-client? As I noted on the mailing list, getting ovirt running from nightlies has been ... challenging.

nirs:
On Sun, Mar 27, 2022 at 4:06 PM Muli Ben-Yehuda ***@***.***>
wrote:

> Hey, just wondering if there are any updates on this one?
>
> If you can confirm this works for lightos we can move forward with this,
> my tests for rbd and iscsi look good so far
>
> is it possible to test with vdsm-client? As I noted on the mailing list,
> getting ovirt running from nightlies has been ... challenging.
>

So you have trouble installing engine? Can you explain the issue?

muliby-lb:
> So you have trouble installing engine? Can you explain the issue?

Not installing, getting it to work. I will send an email to the mailing list tomorrow with what we did and what doesn't work, need to revisit with close attention to detail to make sure we didn't miss anything silly.

lib/vdsm/storage/managedvolume.py
```diff
@@ -112,18 +113,18 @@ def attach_volume(vol_id, connection_info):
                         "Unsupported volume type, supported types are: "
                         "rbd, iscsi")
```
muliby-lb, line 114: need to support `lightos` here, assuming it's the short name we care about - the full name is `cinder.volume.drivers.lightos.LightOSVolumeDriver`. Otherwise I get this from the engine:

```
2022-04-19 07:53:03,899-04 WARN  [org.ovirt.engine.core.vdsbroker.vdsbroker.AttachManagedBlockStorageVolumeVDSCommand] (EE-ManagedThreadFactory-engine-Thread-85595) [4ddaf7c2] Unexpected return value: Status [code=2009, message=Unsupported operation: 'reason=Unsupported volume type, supported types are: rbd, iscsi, context={}']
2022-04-19 07:53:03,899-04 ERROR [org.ovirt.engine.core.vdsbroker.vdsbroker.AttachManagedBlockStorageVolumeVDSCommand] (EE-ManagedThreadFactory-engine-Thread-85595) [4ddaf7c2] Failed in 'AttachManagedBlockStorageVolumeVDS' method
```

bennyz: yes, this is the `driver_volume_type` property, I think the full name would be `volume_driver` or something, but it doesn't matter in this context.

Adding `lightos` to the tuple is enough to make this work? If it is, I believe we can move forward with this PR

muliby-lb: I wish... adding `lightos` is necessary to make progress, but we still hit a failure in the engine. I think I see maybe why it happens, will commment on the engine patch.

bennyz: Well, the current engine code expects the "path" property, so I suppose this is why it fails

My PR went stale, but I rebased and started a build https://github.com/oVirt/ovirt-engine/actions/runs/2189742137
RPMs should be available in about 20-30 minutes (assuming it doesn't fail :)

muliby-lb: That's great, thanks. I'll give it a spin when it will be ready and let you know how it goes.

muliby-lb:
As I wrote in the corresponding engine PR, this works! One VM one volume, so not exactly comprehensive testing yet, but it's the first time we've been able to attach a LightOS volume to an ovirt VM. As I noted earlier, it does require a one-liner to add LightOS to the allowed drivers for MBS, but other than that, no changes were needed.

image

bennyz:
> As I wrote in the corresponding engine PR, this works! One VM one volume, so not exactly comprehensive testing yet, but it's the first time we've been able to attach a LightOS volume to an ovirt VM. As I noted earlier, it does require a one-liner to add LightOS to the allowed drivers for MBS, but other than that, no changes were needed.
> 
> image

This is great, thanks for checking!
 
@nirs 
So the outstanding issues for this PR are:
* The _udevTrigger change
* We currently save the path in the managedvolume db, which should be fine because this is host-specific and it is used for filtering mpath devices
* Changing `SYMLINK` to `KERNEL` in the udev rule, this works for all tested driver types so it should be ok

In ovirt-engine we'll need to make sure live migration for hosts that don't report `managed_path` yet is handled properly

muliby-lb:
@bennyz @nirs any updates on when this will be merged or what else is needed before it can be merged?

nirs:
> @bennyz @nirs any updates on when this will be merged or what else is needed before it can be merged?

Before we merge this we need few changes in this patch to allow future features in MBS.
- Add storage domain id to the API
- Add storage domain id to managed volume path (/run/vdsm/managedvolume/domain-id_volume-id)
- Keep using the stable paths for multipath and rbd (/dev/mapper/xxx, /dev/rbd/yyy)
- Update the udev rule to support both symlinks (/dev/mapper/xxx, /dev/rbd/yyy) and actual devices (/dev/nvmezzz)

We want to include this in oVirt 4.5.1.

@sandrobonazzola do we have a estimate for the release date?

sandrobonazzola:
> @sandrobonazzola do we have a estimate for the release date?

Not yet

lib/vdsm/api/vdsm-api.yml
```diff
@@ -9752,6 +9761,10 @@ ManagedVolume.attach_volume:
     -   description: The information needed to attach a volume
         name: connection_info
         type: *ManagedVolumeConnection
+    -   defaultvalue: null
```
bennyz, line 9764: needs `added: 4.5.1`

lib/vdsm/api/vdsm-api.yml
```diff
@@ -9764,6 +9777,10 @@ ManagedVolume.detach_volume:
     -   description: The volume ID
         name: vol_id
         type: *UUID
+    -   defaultvalue: null
```
bennyz, line 9780: needs `added: 4.5.1`

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -131,8 +136,8 @@ def trigger_managed_udev_rule(path):
 
 
 @expose
-def remove_managed_udev_rule(vol_id):
-    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % vol_id
+def remove_managed_udev_rule(vol_id, sd_id):
+    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % (sd_id, vol_id)
```
aesteve-rh, line 140: This syntax forces any reader to scroll to the top to be able to understand what is the string in `rule_file`, and is a bit odd, relying on the old string formatting with '%' operator.

Maybe using a function that receives `sd_id` and `vol_id` as parameters, closer to where it is used and employing f-string internally would be a nice alternative.

Again, this is out of the scope of this branch.

aesteve-rh (dismissed):

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -115,11 +116,15 @@ def rmAppropriateMultipathRules(thiefId):
...
-        symlink, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
+def add_managed_udev_rule(vol_id, sd_id, path):
+    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % (sd_id, vol_id)
+    device = os.path.relpath(path, '/dev/')
+    if os.path.islink(path):
+        rule = 'SYMLINK=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
+            device, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
+    else:
+        rule = 'KERNEL=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
+            device, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
```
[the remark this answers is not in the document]
bennyz: A separate issue will be created

lib/vdsm/supervdsm_api/udev.py
bennyz: A separate issue will be created

bennyz:
Tested running VMs and live migration with RBD and iSCSI

These are the rule contents:
Rule content
```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_fa728e08-2db4-4016-bfbb-bd75f2ee1735_d1530bb1-8b92-40e8-9d5b-1adbcbc4eedc.rules
SYMLINK=="rbd/volumes/volume-d1530bb1-8b92-40e8-9d5b-1adbcbc4eedc", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
```

```shell
$ cat /etc/udev/rules.d/99-vdsm-managed_fc3b802e-cbc5-4896-995d-87f4b25a1a68_0d850fc9-9dfe-44ee-a4f6-a209d1f4bf7b.rules
SYMLINK=="mapper/20024f4005854000a", RUN+="/usr/bin/chown vdsm:qemu $env{DEVNAME}"
```

lib/vdsm/api/vdsm-api.yml
```diff
@@ -9752,6 +9752,11 @@ ManagedVolume.attach_volume:
     -   description: The information needed to attach a volume
         name: connection_info
         type: *ManagedVolumeConnection
+    -   defaultvalue: null
```
nirs, line 9764: Using blank uuid (00000000-0000-0000-0000-000000000000) in the schema will simplify all the APIs in the entire system. Regardless of engine version, we always have a sd_id and we never have to handle the case when sd_id=None.

lib/vdsm/API.py
```diff
@@ -1809,11 +1809,14 @@ class ManagedVolume(APIBase):
 
     @api.logged(on="api.storage")
     @api.method
-    def attach_volume(self, vol_id, connection_info):
+    def attach_volume(self, vol_id, connection_info, sd_id=None):
```
nirs, line 1812: If we use blank uuid in the schema, we can use here sd_id without default value, and pass it as positional argument to all APIs.

lib/vdsm/API.py
```diff
@@ -1809,11 +1809,14 @@ class ManagedVolume(APIBase):
...
         """
         attach volume and return attached device information
         """
-        return managedvolume.attach_volume(vol_id, connection_info)
+        return managedvolume.attach_volume(
+            vol_id,
+            connection_info,
+            sd_id=sd_id)
```
nirs, line 1819: The internal managedvolume.attach_volume() can be change to accept sd_id as first arguemnt, in the same way we pass it in other places, so this becomes:

    return managedvolume.attach_volume(sd_id, vol_id, connection_info)

lib/vdsm/API.py
```diff
@@ -1820,8 +1820,8 @@ def attach_volume(self, vol_id, connection_info, sd_id=None):
 
     @api.logged(on="api.storage")
     @api.method
-    def detach_volume(self, vol_id):
-        return managedvolume.detach_volume(vol_id)
+    def detach_volume(self, vol_id, sd_id=None):
+        return managedvolume.detach_volume(vol_id, sd_id=sd_id)
```
nirs, line 1824: Same comment for attach volume - use blank uuid as default, so this will be:

    def detach_volume(self, vol_id, sd_id):
        return managedvolume.detach_volume(sd_id, vol_id)

lib/vdsm/api/vdsm-api.yml
```diff
@@ -9769,6 +9769,11 @@ ManagedVolume.detach_volume:
     -   description: The volume ID
         name: vol_id
         type: *UUID
+    -   defaultvalue: null
```
nirs, line 9781: Same comment for attach volume use default of blank uuid.

lib/vdsm/storage/managedvolume.py
```diff
@@ -258,15 +260,19 @@ def _invalidate_lvm_devices(attachment):
         lvm.invalidate_devices()
 
 
-def _silent_remove(db, vol_id):
+def _silent_remove(db, vol_id, sd_id):
```
nirs, line 263: Lets always use sd_id, vol_id - the only place we cannot do this is the schema. The rest of the code can look like sd_id was always part of the code.

lib/vdsm/storage/managedvolume.py
```diff
@@ -114,7 +114,7 @@ def attach_volume(vol_id, connection_info, sd_id=None):
                         "rbd, iscsi")
```
nirs, line 114: Regarding the commit message - it will help if you actually show the different variants of the udev rule :-)

lib/vdsm/storage/managedvolume.py
```diff
@@ -114,7 +114,7 @@ def attach_volume(vol_id, connection_info, sd_id=None):
...
 
                 run_link = _add_run_link(vol_id, sd_id, path)
-                _add_udev_rule(vol_id, path)
+                _add_udev_rule(vol_id, sd_id, path)
```
nirs, line 117: Lets keep also here sd_id, vol_id, ...

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -115,24 +116,28 @@ def rmAppropriateMultipathRules(thiefId):
...
 @expose
 def trigger_managed_udev_rule(path):
-    _udevTrigger(property_matches=(('DEVLINKS', path),))
+    _udevTrigger(property_matches=(('DEVNAME', os.path.realpath(path)),))
 
 
 @expose
-def remove_managed_udev_rule(vol_id):
-    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % vol_id
+def remove_managed_udev_rule(vol_id, sd_id):
```
nirs, line 139: sd_id, vol_id

lib/vdsm/clientIF.py
```diff
@@ -451,6 +451,9 @@ def prepareVolumePath(self, drive, vmId=None, path=None):
                             # not speicfy the type, we must use 'file'.
                             drive['diskType'] = DISK_TYPE.FILE
                     volPath = res['path']
+            # Managed drives are prepared in ManagedVolume.attach_volume
+            elif drive.get("managed", False):
+                volPath = drive['path']
```
nirs, line 456: We discussed moving this first - before handling images. Any reason to leave it here or you just forgot to move it?

nirs (changes requested):
Nice - but the lightos support is missing in this version?

lib/vdsm/api/vdsm-api.yml
```diff
@@ -1187,9 +1187,18 @@ types:
             name: connection_info
```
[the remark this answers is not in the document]
muliby-lb: Pedantically, NVMe/TCP is an NVMe-oF transport protocol binding, so NVMe-oF is inclusive of NVMe/TCP. I suggest to use NVMe-oF if it was tested with multiple transport bindings (e.g., both NVMe/RoCE and NVMe/TCP), or use NVMe/TCP if it was only tested with NVMe/TCP.

lib/vdsm/clientIF.py
bennyz: I forgot :)

lib/vdsm/storage/managedvolume.py
bennyz: It's in the PR description, I'll add it to the commit message too

bennyz:
> Nice - but the lightos support is missing in this version?

as discussed - lightos support will be added in a separate PR

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -115,24 +116,32 @@ def rmAppropriateMultipathRules(thiefId):
...
         _log.debug("Creating rule %s: %r", rule_file, rule)
         rf.write(rule)
 
 
 @expose
 def trigger_managed_udev_rule(path):
-    _udevTrigger(property_matches=(('DEVLINKS', path),))
+    # Example path: /dev/rbd/volumes/volume-d1530bb1-...-1adbcbc4eedc
+    if os.path.islink(path):
+        _udevTrigger(property_matches=(('DEVLINKS', f"*{path}*"),))
```
nirs, line 137: Why do we need `*{path}*` now? This worked before with path as is. What is the meaning of the `*` in the match?

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -115,24 +116,32 @@ def rmAppropriateMultipathRules(thiefId):
...
+            device, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
     with open(rule_file, "w") as rf:
         _log.debug("Creating rule %s: %r", rule_file, rule)
         rf.write(rule)
 
 
 @expose
 def trigger_managed_udev_rule(path):
-    _udevTrigger(property_matches=(('DEVLINKS', path),))
+    # Example path: /dev/rbd/volumes/volume-d1530bb1-...-1adbcbc4eedc
```
nirs, line 135: We need example for each case in the if block, best 3 example for /dev/mapper/xxx, /dev/rbd/volumes/volume-yyy and /dev/nvme...

Since we need the examples in both add_managed_udev_rules and trigger_managed_udev_rule - maybe add a comment in the module about managed volumes paths?

lib/vdsm/supervdsm_api/udev.py
```diff
@@ -115,24 +116,32 @@ def rmAppropriateMultipathRules(thiefId):
 
 
 @expose
-def add_managed_udev_rule(vol_id, path):
-    rule_file = _UDEV_RULE_FILE_NAME_MANAGED % vol_id
-    symlink = os.path.relpath(path, '/dev/')
-    rule = 'SYMLINK=="%s", RUN+="%s %s:%s $env{DEVNAME}"\n' % (
-        symlink, EXT_CHOWN, DISKIMAGE_USER, DISKIMAGE_GROUP)
+def add_managed_udev_rule(sd_id, vol_id, path):
```
nirs, line 119: Would be nice to add examples for all types of inputs.

nirs (commented):
Looks good, but the trigger pattern looks fishy.

lib/vdsm/supervdsm_api/udev.py
nirs: usedevadm(8) says:

    udevadm trigger [options] [devpath|file|unit]

Since we trigger events for known path, can we use devpath and remove the --property-match?

I tried this:

In one shell:
```
[nsoffer@host4 ~]$ udevadm monitor
monitor will print the received events for:
UDEV - the event which udev sends out after rule processing
KERNEL - the kernel uevent

```

In another shell:
```
[root@host4 vdsm]# udevadm trigger /dev/mapper/0QEMU_QEMU_HARDDISK_data-fc-01
[root@host4 vdsm]# udevadm trigger /dev/mapper/36001405789e8ab39498483ea5858c779
```

In the monitor shell we see:
```
KERNEL[103021.013793] change   /devices/virtual/block/dm-2 (block)
UDEV  [103021.069396] change   /devices/virtual/block/dm-2 (block)
KERNEL[103213.074750] change   /devices/virtual/block/dm-15 (block)
UDEV  [103213.114084] change   /devices/virtual/block/dm-15 (block)
```

Seems to work - can you try with rbd and the iscsi server use for testing?

bennyz: I'll check using the path,
as for asterisks, explained here: https://github.com/oVirt/vdsm/pull/89#discussion_r876812403

bennyz: @nirs 
 
I can use just the path if I apply the following diff:
```diff
diff --git a/lib/vdsm/common/udevadm.py b/lib/vdsm/common/udevadm.py
index d849f0228..b5e7cee2d 100644
--- a/lib/vdsm/common/udevadm.py
+++ b/lib/vdsm/common/udevadm.py
@@ -53,7 +53,7 @@ def settle(timeout, exit_if_exists=None):
         logging.error("%s", e)


-def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
+def trigger(path=None, attr_matches=(), property_matches=(), subsystem_matches=()):
     '''
     Request device events from the kernel.

@@ -87,6 +87,7 @@ def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
     _run_command(['control', '--reload'])

     cmd = ['trigger', '--verbose', '--action', 'change']
+    cmd.append(path)

     for name, value in property_matches:
         cmd.append('--property-match={}={}'.format(name, value))
diff --git a/lib/vdsm/supervdsm_api/udev.py b/lib/vdsm/supervdsm_api/udev.py
index 506d195c3..5af298541 100644
--- a/lib/vdsm/supervdsm_api/udev.py
+++ b/lib/vdsm/supervdsm_api/udev.py
@@ -132,7 +132,7 @@ def add_managed_udev_rule(sd_id, vol_id, path):

 @expose
 def trigger_managed_udev_rule(path):
-    # Example path: /dev/rbd/volumes/volume-d1530bb1-...-1adbcbc4eedc
-    if os.path.islink(path):
-        _udevTrigger(property_matches=(('DEVLINKS', f"*{path}*"),))
-    else:
-       _udevTrigger(property_matches=(('DEVNAME', path),))
-
+    _udevTrigger(path=path)
```

The other options include the asterisks like shown above, or use the original with `os.path.realpath`
```python
    _udevTrigger(property_matches=(('DEVNAME', os.path.realpath(path)),))
```

nirs: > @nirs
> 
> I can use just the path if I apply the following diff:
> 
> ```diff
> diff --git a/lib/vdsm/common/udevadm.py b/lib/vdsm/common/udevadm.py
> index d849f0228..b5e7cee2d 100644
> --- a/lib/vdsm/common/udevadm.py
> +++ b/lib/vdsm/common/udevadm.py
> @@ -53,7 +53,7 @@ def settle(timeout, exit_if_exists=None):
>          logging.error("%s", e)
> 
> 
> -def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
> +def trigger(path=None, attr_matches=(), property_matches=(), subsystem_matches=()):
>      '''
>      Request device events from the kernel.
> 
> @@ -87,6 +87,7 @@ def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
>      _run_command(['control', '--reload'])
> 
>      cmd = ['trigger', '--verbose', '--action', 'change']
> +    cmd.append(path)

Should be:

    if path:
        cmd.append(path)

> 
>      for name, value in property_matches:
>          cmd.append('--property-match={}={}'.format(name, value))

And it should be done *after* adding the other options.

I'm not sure if using both path and properties makes sense, but this is something callers
should handle.

> diff --git a/lib/vdsm/supervdsm_api/udev.py b/lib/vdsm/supervdsm_api/udev.py
> index 506d195c3..5af298541 100644
> --- a/lib/vdsm/supervdsm_api/udev.py
> +++ b/lib/vdsm/supervdsm_api/udev.py
> @@ -132,7 +132,7 @@ def add_managed_udev_rule(sd_id, vol_id, path):
> 
>  @expose
>  def trigger_managed_udev_rule(path):
> -    # Example path: /dev/rbd/volumes/volume-d1530bb1-...-1adbcbc4eedc
> -    if os.path.islink(path):
> -        _udevTrigger(property_matches=(('DEVLINKS', f"*{path}*"),))
> -    else:
> -       _udevTrigger(property_matches=(('DEVNAME', path),))
> -
> +    _udevTrigger(path=path)

This part looks good.

> The other options include the asterisks like shown above, or use the original with `os.path.realpath`
> 
> ```python
>     _udevTrigger(property_matches=(('DEVNAME', os.path.realpath(path)),))
> ```

So the original worked with realative path, and it is broken now becasue we started
to use the full path?

lib/vdsm/common/udevadm.py
```diff
@@ -53,7 +53,12 @@ def settle(timeout, exit_if_exists=None):
         logging.error("%s", e)
 
 
-def trigger(attr_matches=(), property_matches=(), subsystem_matches=()):
+def trigger(
+    path=None,
```
nirs, line 57: Any reason to add the path first? Looks like all the callers are using key=value
so the order does not matter, but the command always add path at the end, so adding the parameter as the last parameter makes the python call mode like the actual command.

lib/vdsm/common/udevadm.py
```diff
@@ -53,7 +53,12 @@ def settle(timeout, exit_if_exists=None):
...
+    attr_matches=(),
+    property_matches=(),
+    subsystem_matches=(),
+):
     '''
     Request device events from the kernel.
```
nirs, line 63: The new argument is not documented under `Arguments:`

lib/vdsm/common/udevadm.py
bennyz: No particular reason, I'll move it to the end

nirs (approved):
Was it tested with current engine (consuming path) and rbd/iscsi MBS?

bennyz:
Tested iSCSI and RBD:
Run VM
Migrate VM
Copy

PR #90, opened 2022-03-03
Add a blacklist for rbd devices

sleviim:
When using Ceph, multipath might prevent the rbd device from being
unmapped and it stays mapped.
It results with getting the following error while operating the disk:
rbd.ImageBusy: [errno 16] RBD image is busy
This patch adds rbd to the multipath blacklist.

Bug-Url: https://bugzilla.redhat.com/1881832
Bug-Url: https://bugzilla.redhat.com/1755801
Signed-off-by: Shani Leviim [sleviim@redhat.com](mailto:sleviim@redhat.com)

### DIFF
--- a/lib/vdsm/tool/configurators/multipath.py
+++ b/lib/vdsm/tool/configurators/multipath.py
@@ -37,9 +37,10 @@ _CONF_FILE = "/etc/multipath.conf"
 # "VDSM REVISION X.Y" tag.  Note that older version used "RHEV REVISION X.Y"
 # format.
 
-_CURRENT_TAG = "# VDSM REVISION 2.0"
+_CURRENT_TAG = "# VDSM REVISION 2.1"
 
 _OLD_TAGS = (
+    "# VDSM REVISION 2.0",
     "# VDSM REVISION 1.9",
     "# VDSM REVISION 1.8",
     "# VDSM REVISION 1.7",
@@ -192,8 +193,8 @@ defaults {
     max_fds                     4096
 }
 
-# Blacklist local and obsolete protocols which run on devices which are not
-# multipathable.
+Blacklist local devices, obsolete protocols, and device nodes which
+should not be used with multipath.
 #
 # Complete list of protocols recognized by multipath:
 # scsi:fcp        Fibre Channel
@@ -253,9 +254,14 @@ defaults {
 #   blacklist_exceptions {
 #       protocol "(scsi:spi|scsi:ssa)"
 #   }
+#
+# We blacklist all RADOS Block Device (RBD) devices.
+# When using Ceph, multipath prevents the rbd devices from being unmapped
+# and the devices remain busy.
 
 blacklist {
     protocol "(scsi:adt|scsi:sbp)"
+    devnode "^(rbd)[0-9]*"
 }
 
 # Options defined here override device specific options embedded into


lib/vdsm/storage/mpathconf.py
```diff
@@ -85,9 +86,11 @@ def format_blacklist(wwids):
 
     Returns:
         A formatted blacklist section string.
+        RBD devices were added to the multipath blacklist (BZ#1881832)
     """
     lines = "\n".join('    wwid "{}"'.format(w) for w in sorted(wwids))
     return """blacklist {{
+    devnode "^(rbd)[0-9]*"
```
vjuranek, line 93: This is create dynamically and is specific for each host, while rbd blacklist is static, always same. I'd consider to move it to [multipath blacklist](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurators/multipath.py#L258)

nirs: I agree with Vojta, this is not the place to put static rules.

lib/vdsm/storage/mpathconf.py
```diff
@@ -85,9 +86,11 @@ def format_blacklist(wwids):
 
     Returns:
         A formatted blacklist section string.
+        RBD devices were added to the multipath blacklist (BZ#1881832)
```
nirs, line 89: This is not the place to document why RBD is included.

nirs (changes requested):

lib/vdsm/storage/mpathconf.py
sleviim: ACK, shoot in the dark for finding the right place for it :)

lib/vdsm/storage/mpathconf.py
sleviim: Mentioned it on multipath.py, is there a better place/phrasing?

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -214,6 +214,7 @@
 # We blacklist following protocols:
```
nirs, line 214: Please update the commit message, this patch does not change
mpathconf.py.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -214,6 +214,7 @@
...
 # scsi:adt  local storage protocol
 # scsi:sbp  local storage protocol
+# RBD devices were also added to the multipath blacklist (BZ#1881832)
```
nirs, line 217: This is a list of protocols we blacklist. This is not the place to add
the note about RBD.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -256,6 +257,7 @@
```
nirs, line 257: Please add new section in the comment about blacklisted devnodes,
similar to the section about blacklisted protocols. Each line
in the blacklist should be explained in the comment.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -256,6 +257,7 @@
 
 blacklist {
     protocol "(scsi:adt|scsi:sbp)"
+    devnode "^(rbd)[0-9]*"
```
nirs, line 260: The actual change looks good, but we need @bennyz ack.

nirs (changes requested):

lib/vdsm/tool/configurators/multipath.py
bennyz: ACK'd, this rule was suggested by ceph developers and has been confirmed by users to work

lib/vdsm/tool/configurators/multipath.py
barpavel: I guess we can agree on some "PR best practices".
Like we can continue using prefixes (`core`, `webadmin`, etc.) in commits that form the PR, but PR title can be a single 1-sentence starting with uppercase describing what it does **without** any prefix.
In this case it can look like: `Add a blacklist for rbd devices`.

sleviim: Ack

lib/vdsm/tool/configurators/multipath.py
sleviim: ACK, looked for the right place for that :)

lib/vdsm/tool/configurators/multipath.py
sleviim: Marking as resolved

lib/vdsm/tool/configurators/multipath.py
sleviim: @nirs Done, please share your feedback

vjuranek (approved):

lib/vdsm/tool/configurators/multipath.py
barpavel: @nirs Maybe worth changing line 256 to be an empty line (rather then a comment), thus making 2 comments' blocks visually separated or maybe even moving a new 3-lines' comment (lines 257-259) to be inside the `blacklist` section before line 263 `devnode "^(rbd)[0-9]*"`.
I mean this long comment's section is more like a general Blacklist description and a new Shani's comment is more like a specific line/change explanation. IMHO.
Anyway, approving.

barpavel (approved):

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -193,7 +193,7 @@
 }
 
 # Blacklist local and obsolete protocols which run on devices which are not
-# multipathable.
+# multipathable and RADOS Block Device (RBD) devices.
```
nirs, line 196: This will be better more generic so we don't have to change it again.

   Blacklist local devices, obsolete protocols, and device nodes which
   should not be used with multipath.

lib/vdsm/tool/configurators/multipath.py
nirs: The comment explains the blacklist section, so it should be one comment.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -253,9 +253,14 @@
...
 #       protocol "(scsi:spi|scsi:ssa)"
 #   }
+#
+# We blacklist all RADOS Block Device (RBD) devices.
+# When using Ceph, multipath prevents the rbd devices from being unmapped
+# and the devices remain busy.
 
 blacklist {
     protocol "(scsi:adt|scsi:sbp)"
+    devnode "^(rbd)[0-9]*"
```
nirs, line 263: This change will not be effective on upgrades without creating a new revision
of the configuration. You have to change this:

```
_CURRENT_TAG = "# VDSM REVISION 2.0"
_OLD_TAGS = (
    "# VDSM REVISION 1.9",
```

The current revision must be 2.1, and revision 2.0
must move to the old tags section.

Please check previous changes in this file, they show how to change this
file. One example is d2dcd5d015463589ea9aad147b75c95d74effa0f.

nirs (changes requested):

lib/vdsm/tool/configurators/multipath.py
sleviim: Ack

lib/vdsm/tool/configurators/multipath.py
sleviim: IIUC, I should leave it as I've changed it, right?
`#
# We blacklist all RADOS Block Device (RBD) devices.
# When using Ceph, multipath prevents the rbd devices from being unmapped
# and the devices remain busy.`

lib/vdsm/tool/configurators/multipath.py
sleviim: Thanks for the example.
ack

nirs (approved):
Thanks, please described how this was tested.
We need to test new install and upgrades.

sleviim:
Verification:
- After artifacts were built (checks > CI > artifacts > https://github.com/oVirt/vdsm/suites/5591507319/artifacts/181334210/rpm-el8.zip), scp it to /etc/yum.repos.d on the relevant host.
- unzipped the file > cd to the directory > dnf upgrade *.rpm
- cat /etc/multipath.conf:
`blacklist {
    protocol "(scsi:adt|scsi:sbp)"
    devnode "^(rbd)[0-9]*"
}
`

nirs:
@sleviim thanks, looks good, lets see if OST is happy (you could start it in the same way).

bennyz (approved):

PR #92, opened 2022-03-10
virt: Add support for vGPU driver parameters

mz-pdm:
Engine is going to add a new metadata item for vGPU devices,
mdevDriverParameters.  It specifies driver parameters to be set when
the mediated device is created, by writing a file in /sys.

The driver consumes the last character written to the file in /sys, so
it is important to write a new line character at the end, as `echo'
from a shell command line would do.

Since mediated devices are created before each use and destroyed after
each use, it's enough to write the driver parameters, if specified,
only when the device is created.  It's not necessary to clear them
when the device is removed, although it is possible to do it by
writing a space to the driver parameters file in /sys.

Currently, only NVidia driver parameters are supported.  If setting
driver parameters is attempted for other cards or if setting the
driver parameters fails for another reason, creation of the mediated
device will fail and the VM won't be started.

Bug-Url: https://bugzilla.redhat.com/1987121

### DIFF
--- a/lib/vdsm/supervdsm_api/virt.py
+++ b/lib/vdsm/supervdsm_api/virt.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2021 Red Hat, Inc.
+# Copyright 2016-2022 Red Hat, Inc.
 #
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -84,13 +84,14 @@ def hugepages_alloc(count, path):
 
 
 @expose
-def mdev_create(device, mdev_type, mdev_uuid=None):
-    """Create the desired mdev type.
+def mdev_create(device, mdev_properties, mdev_uuid=None):
+    """Create the desired mediated device.
 
     Args:
         device: PCI address of the parent device in the format
             (domain:bus:slot.function). Example:  0000:06:00.0.
-        mdev_type: Type to be spawned. Example: nvidia-11.
+        mdev_properties: `vdsm.common.hostdev.MdevProperties` instance
+            specifying the desired device properties
         mdev_uuid: UUID for the spawned device. Keeping None generates a new
             UUID.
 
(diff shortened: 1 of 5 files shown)


ahadas (approved):
lgtm

lib/vdsm/supervdsm_api/virt.py
```diff
@@ -102,16 +102,30 @@ def mdev_create(device, mdev_type, mdev_uuid=None):
...
-
     with open(path, 'w') as f:
         f.write(mdev_uuid)
 
+    driver_parameters = mdev_properties.driver_parameters
+    if driver_parameters is not None:
+        param_path = f'/sys/bus/mdev/devices/{mdev_uuid}/nvidia/vgpu_params'
+        try:
+            with open(param_path, 'w') as f:
+                f.write('%s\n' % (driver_parameters,))
```
almusil, line 118: Maybe worth adding the comment that additional character is required.

almusil (dismissed):

lib/vdsm/supervdsm_api/virt.py
```diff
@@ -102,16 +102,30 @@ def mdev_create(device, mdev_type, mdev_uuid=None):
...
+        except:
+            # We must perform the cleanup here, because we may not know the
+            # device UUID elsewhere.
+            try:
+                mdev_delete(device, mdev_uuid)
+            except Exception as e:
+                logging.error("Failed to remove vGPU device with failed "
```
nyoxi, line 125: Is it always vGPU?

lib/vdsm/supervdsm_api/virt.py
nyoxi: What exactly does "additional character" mean? Is it new line? Is it space? Is it arbitrary character (meaning last character is always lost)? ... The commit message is not making this clear either.

nyoxi (commented):
Just a few nitpicks.

lib/vdsm/supervdsm_api/virt.py
mz-pdm: Yes.

tinez (commented):
Apart from the other comments I was thinking - wouldn't it make sense to suck in `uuid` to `MdevProperties` as well?

mz-pdm:
> wouldn't it make sense to suck in `uuid` to `MdevProperties` as well?

I think it's better to keep them separated.  The UUID is somewhat different, e.g. it's not among vGPU parameters on the Engine side, or it's passed to mdev_delete where MdevProperties is of no use.

lib/vdsm/supervdsm_api/virt.py
```diff
@@ -84,7 +84,7 @@ def hugepages_alloc(count, path):
 
 
 @expose
-def mdev_create(device, mdev_type, mdev_uuid=None):
+def mdev_create(device, mdev_properties, mdev_uuid=None):
     """Create the desired mdev type.
 
     Args:
```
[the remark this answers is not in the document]
mz-pdm: Both the docstrings updated, in separate commits.

lib/vdsm/supervdsm_api/virt.py
```diff
@@ -102,16 +102,30 @@ def mdev_create(device, mdev_type, mdev_uuid=None):
...
     with open(path, 'w') as f:
         f.write(mdev_uuid)
 
+    driver_parameters = mdev_properties.driver_parameters
+    if driver_parameters is not None:
+        param_path = f'/sys/bus/mdev/devices/{mdev_uuid}/nvidia/vgpu_params'
+        try:
+            with open(param_path, 'w') as f:
+                f.write('%s\n' % (driver_parameters,))
+        except:
```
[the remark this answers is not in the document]
mz-pdm: Here we must catch all exceptions, because we must remove the device created above under any circumstances.

tinez (approved):

PR #93, opened 2022-03-12
NBD bitmap validation with single volume

nirs:
When exporting bitmap from single volume, we did not validate that
the bitmap exists and valid. When staring qemu-nbd service, qemu-nbd
starts normally so we report success, but then it fails and exits.
Later, the user fail when trying to download the backup.

Another issue, we created an overlay when exporting a volume with a parent
when the parent volume does not include the bitmap. This case does not
require an overlay.

- Add failing tests for bitmap validation
- Add bitmap validation when exporting single volume
- Create overlay only if needed

### DIFF
--- a/lib/vdsm/storage/nbd.py
+++ b/lib/vdsm/storage/nbd.py
@@ -192,7 +192,9 @@ def _create_overlay(server_id, backing, bitmap):
     """
     filenames = _find_bitmap(backing, bitmap)
     if not filenames:
-        raise se.BitmapDoesNotExist(bitmap=bitmap)
+        raise se.BitmapDoesNotExist(
+            reason=f"Bitmap does not exist in {backing}",
+            bitmap=bitmap)
 
     overlay = transientdisk.create_disk(
         server_id,
--- a/lib/vdsm/storage/nbd.py
+++ b/lib/vdsm/storage/nbd.py
@@ -134,7 +134,9 @@ def start_server(server_id, config):
 
     _create_rundir()
 
-    using_overlay = cfg.bitmap and vol.getParent() != sc.BLANK_UUID
+    # If the bitmap exists in the volume backing chain, we need to
+    # create an overlay exporting the bitmap from the entire chain.
+    using_overlay = cfg.bitmap and len(bitmap_chain) > 1
 
     if using_overlay:
         path = _create_overlay(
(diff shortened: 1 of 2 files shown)


nirs:
Verified with the nbd tests (locally) and backup stress test:
https://gitlab.com/nirs/ovirt-stress/-/tree/master/backup

nirs:
Tested using backup stress test:
https://gitlab.com/nirs/ovirt-stress/-/tree/master/backup

2022-03-13 10:37:49,122 INFO (MainThread) 64 full backups, 896 incremental backups, 0 failed, 960 passed, 0 errored in 333.8 minutes

In this test all the incremental backups use single volume. With this change, we don't 
create an overlay in this case.

bennyz (dismissed):

lib/vdsm/storage/nbd.py
```diff
@@ -119,15 +119,28 @@ def start_server(server_id, config):
...
+        if not bitmap_chain:
+            raise se.BitmapDoesNotExist(
+                reason=f"Bitmap does not exist in {vol.volumePath}",
+                bitmap=cfg.bitmap)
 
     _create_rundir()
 
-    using_overlay = cfg.bitmap and vol.getParent() != sc.BLANK_UUID
+    # If the bitmap exists in the volume backing chain, we need to
+    # create an overlay exporting all bitmap from the entire chain.
```
barpavel, line 138: `bitmap` --> `bitmaps` (plural).

tests/storage/nbd_test.py
```diff
@@ -633,7 +633,33 @@ def test_bitmap_backing_chain(nbd_env):
...
+
+    config = {
+        "sd_id": vol.sdUUID,
+        "img_id": vol.imgUUID,
+        "vol_id": vol.volUUID,
+        "readonly": True,
+        "bitmap": bitmap,
+    }
+
+    # Starting server will fail since the bitmap in vol is in in use.
```
barpavel, line 654: `in in` -- > `in`.

barpavel (approved):

lib/vdsm/storage/nbd.py
nirs: We are exporting a single bitmap - it may exist in multiple images in the chain, so
we need to merge all the copies from the chain into the overlay.

nirs: I replaced "all" with "the". If you want to understand this better look in _create_overlay and the relevant tests.

vjuranek (approved):

barpavel (approved):

PR #99, opened 2022-03-17
Numa migration

nyoxi:
Necessary code and API changes to allow migration of NUMA VMs. It has to be possible to adapt the NUMA pinning during migration because CPUs on destination host may be used differently or destination host may have different hardware. Similarly it is necessary to allow modification to manual CPU pinning to limit virtual NUMA cells to corresponding physical NUMA cells.

### DIFF
--- a/lib/vdsm/virt/migration.py
+++ b/lib/vdsm/virt/migration.py
@@ -29,6 +29,7 @@ import re
 import threading
 import time
 import libvirt
+import xml.etree.ElementTree as ET
 
 from vdsm.common import concurrent
 from vdsm.common import conv
@@ -649,10 +650,27 @@ class SourceThread(object):
                     continue
                 cputune.remove(vcpu)
         if self._destination_cpusets is not None:
+            # Reconfigure CPU pinning based on the call parameter
             if cputune is None:
                 cputune = xml.etree.ElementTree.Element('cputune')
                 dom.append(cputune)
-            for vcpupin in self._destination_cpusets:
+            # First modify existing elements
+            for vcpupin in vmxml.find_all(cputune, 'vcpupin'):
+                vcpu_id = int(vcpupin.get('vcpu'))
+                if vcpu_id >= 0 and vcpu_id < len(self._destination_cpusets):
+                    vcpupin.set('cpuset',
+                                str(self._destination_cpusets[vcpu_id]))
+                    self._destination_cpusets[vcpu_id] = None
+            # Now create elements for pinning that was not there before.
+            # This should happen only for pinning that was removed above. It
+            # should not happen for manual CPU pinning because it would render
+            # the value of manuallyPinedCPUs metadata invalid.
+            for vcpu_id, cpuset in enumerate(self._destination_cpusets):
+                if cpuset is None:
+                    continue
+                vcpupin = ET.Element('vcpupin')
+                vcpupin.set('vcpu', str(vcpu_id))
+                vcpupin.set('cpuset', str(cpuset))
                 cputune.append(vcpupin)
         if self._destination_numa_nodesets is not None:
             numatune = dom.find('numatune')
--- a/lib/vdsm/virt/migration.py
+++ b/lib/vdsm/virt/migration.py
@@ -139,7 +139,7 @@ class SourceThread(object):
                  tunneled=False, dstqemu='', abortOnError=False,
                  consoleAddress=None, compressed=False,
                  autoConverge=False, recovery=False, encrypted=False,
-                 cpusets=None, parallel=None, **kwargs):
+                 cpusets=None, parallel=None, numaNodesets=None, **kwargs):
         self.log = vm.log
         self._vm = vm
         self._dom = DomainAdapter(self._vm)
@@ -202,6 +202,7 @@ class SourceThread(object):
         # the first extend of the disk during migration if finished.
         self._supports_disk_refresh = None
         self._destination_cpusets = cpusets
+        self._destination_numa_nodesets = numaNodesets
 
     def start(self):
         self._thread.start()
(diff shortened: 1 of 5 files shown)


nyoxi:
@liranr23 @ahadas

lib/vdsm/virt/migration.py
```diff
@@ -648,11 +650,38 @@ def _migration_params(self, muri):
...
         if self._destination_cpusets is not None:
+            # Reconfigure CPU pinning based on call parameter
             if cputune is None:
                 cputune = xml.etree.ElementTree.Element('cputune')
                 dom.append(cputune)
-            for vcpupin in self._destination_cpusets:
+            # First modify existing elements
+            for vcpupin in vmxml.find_all(cputune, 'vcpupin'):
+                vcpu_id = int(vcpupin.get('vcpu'))
+                if vcpu_id > 0 and vcpu_id < len(self._destination_cpusets):
```
mz-pdm, line 660: Really vcpu_id > 0 and not vcpu_id >= 0?

And is it expected to see vcpu_id outside the given range or should a warning be logged in such a case?

lib/vdsm/virt/migration.py
```diff
@@ -648,11 +650,38 @@ def _migration_params(self, muri):
...
+                    vcpupin.set('cpuset',
+                                str(self._destination_cpusets[vcpu_id]))
+                    self._destination_cpusets[vcpu_id] = None
+            # Now create elements for pinning that was not there before.
+            #
```
mz-pdm, line 665: The empty comment line looks weird to me, can it be omitted?

lib/vdsm/virt/migration.py
```diff
@@ -648,11 +650,38 @@ def _migration_params(self, muri):
...
+            for vcpu_id, cpuset in enumerate(self._destination_cpusets):
+                if cpuset is None:
+                    continue
+                vcpupin = ET.Element('vcpupin')
+                vcpupin.set('vcpu', str(vcpu_id))
+                vcpupin.set('cpuset', str(cpuset))
                 cputune.append(vcpupin)
+        if self._destination_numa_nodesets is not None:
+            numatune = dom.find('numatune')
+            if numatune is not None:
```
mz-pdm, line 678: Is it valid if numaNodesets is specified but there is no <numatune>?  Does such a situation deserve logging something at least?

mz-pdm (commented):
Looks overall good, some questions etc. inside.

In the commit message: pre-createing -> pre-creating

lib/vdsm/virt/migration.py
```diff
@@ -648,11 +650,38 @@ def _migration_params(self, muri):
                     continue
                 cputune.remove(vcpu)
         if self._destination_cpusets is not None:
+            # Reconfigure CPU pinning based on call parameter
```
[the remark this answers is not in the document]
nyoxi: updated

lib/vdsm/virt/migration.py
nyoxi: updated

lib/vdsm/virt/migration.py
nyoxi: No, such case is not valid. This however should already be handled during validation in vm.py.

lib/vdsm/virt/migration.py
nyoxi: > Really vcpu_id > 0 and not vcpu_id >= 0?

Yes, it should be `>= 0`. I have it fixed locally just didn't push it yet.


> And is it expected to see vcpu_id outside the given range or should a warning be logged in such a case?

It is possible to see such `vcpu_id`. There may be vCPU pinning for vCPUs that are defined but not active (ready for hot-plugging). We don't really care about those here (correct pinning will be specified during hot-plug). Producing a warning sounds superfluous. We could go with info level, but I still think this is not necessary.

nyoxi:
> In the commit message: pre-createing -> pre-creating

updated

nyoxi:
verified

lib/vdsm/virt/migration.py
mz-pdm: I see, sure, no log message needed in such a case.

mz-pdm:
> verified

Please use "verified" label to mark Vdsm PRs as verified.

nyoxi:
> > verified
> 
> Please use "verified" label to mark Vdsm PRs as verified.

I am not allowed to modify label. In fact, I am not allowed to modify anything besides PR title.

mz-pdm:
> I am not allowed to modify label. In fact, I am not allowed to modify anything besides PR title.

Oops, that's not very good.  And nobody from Vdsm maintainers is allowed to view permission settings to check them.  Let's ask on the mailing list then.

lib/vdsm/virt/migration.py
```diff
@@ -648,11 +650,37 @@ def _migration_params(self, muri):
...
+                vcpupin.set('vcpu', str(vcpu_id))
+                vcpupin.set('cpuset', str(cpuset))
                 cputune.append(vcpupin)
+        if self._destination_numa_nodesets is not None:
+            numatune = dom.find('numatune')
+            if numatune is not None:
+                for memnode in vmxml.find_all(numatune, 'memnode'):
+                    cellid = int(memnode.get('cellid'))
+                    if (cellid >= 0 and
+                            cellid < len(self._destination_numa_nodesets)):
```
liranr23, line 681: why this `if` needed? i can't think of a case `cellid` would be < 0. Also since you already validated the length, can it be bigger than `self._destination_numa_nodesets`?

nyoxi: > why this `if` needed? i can't think of a case `cellid` would be < 0.

I suppose you're right. Libvirt should prevent negative ID from appearing in the XML.

> Also since you already validated the length, can it be equal or bigger than `self._destination_numa_nodesets`?

Yes, it can. We expect `numaNodesets` to contain only active CPUs, but there may be pinning for non-active CPUs that are ready for hot-plugging.

nyoxi: > > why this `if` needed? i can't think of a case `cellid` would be < 0.
> 
> I suppose you're right. Libvirt should prevent negative ID from appearing in the XML.

Since this does not do any harm let's leave it as it is to avoid blocking the merge.

liranr23 (approved):

PR #100, opened 2022-03-18
copy_data: only skip lock for div endpoint

bennyz:
As a result of commit 8dc293a (copy_data: skip locking source when same image is used), a regression was introduced, breaking copy_data operations involving endpoints that aren't DIV, currently only affects external endpoints as this is the only other endpoint type.

Because the external endpoint does not have the img_id property, it will fail:
```
  File "/usr/lib/python3.6/site-packages/vdsm/storage/task.py", line 884, in _run
    return fn(*args, **kargs)
  File "</usr/lib/python3.6/site-packages/decorator.py:decorator-gen-199>", line 2, in sdm_copy_data
  File "/usr/lib/python3.6/site-packages/vdsm/common/api.py", line 50, in method
    ret = func(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/vdsm/storage/hsm.py", line 3357, in sdm_copy_data
    copy_bitmaps=copy_bitmaps)
  File "/usr/lib/python3.6/site-packages/vdsm/storage/sdm/api/copy_data.py", line 60, in __init__
    lock_image=source['img_id'] != destination['img_id'])
KeyError: 'img_id'
```

This PR moves the img_id check to CopyDataEndpointDiv

Bug-Url: https://bugzilla.redhat.com/show_bug.cgi?id=2066285

### DIFF
--- a/lib/vdsm/storage/sdm/api/copy_data.py
+++ b/lib/vdsm/storage/sdm/api/copy_data.py
@@ -53,14 +53,11 @@ class Job(base.Job):
     def __init__(self, job_id, host_id, source, destination,
                  copy_bitmaps=False):
         super(Job, self).__init__(job_id, 'copy_data', host_id)
+        self._dest = _create_endpoint(
+            destination, host_id, writable=True, job_id=job_id)
         self._source = _create_endpoint(
             source, host_id, writable=False, job_id=job_id,
-            # Skip locking image for source endpoint as locking would fail
-            # if copy data is performed within the same image.
-            lock_image=source['img_id'] != destination['img_id'])
-        self._dest = _create_endpoint(
-            destination, host_id, writable=True, job_id=job_id,
-            is_destination=True)
+            dest=self._dest)
         self._operation = None
         self._copy_bitmaps = copy_bitmaps
 
@@ -122,16 +119,14 @@ class Job(base.Job):
 
 
 def _create_endpoint(params, host_id, writable, job_id=None,
-                     is_destination=False,
-                     lock_image=True):
+                     dest=None):
     endpoint_type = params.pop('endpoint_type')
     if endpoint_type == 'div':
         return CopyDataDivEndpoint(
             params,
             host_id,
             writable,
-            is_destination=is_destination,
-            lock_image=lock_image)
+            dest=dest)
     elif endpoint_type == 'external':
         return CopyDataExternalEndpoint(params, host_id, job_id)
     else:
@@ -146,15 +141,18 @@ class CopyDataDivEndpoint(properties.Owner):
                                     maxval=sc.MAX_GENERATION)
     prepared = properties.Boolean(default=False)
 
-    def __init__(self, params, host_id, writable, is_destination=False,
-                 lock_image=True):
+    def __init__(self, params, host_id, writable, dest=None):
         self.sd_id = params.get('sd_id')
         self.img_id = params.get('img_id')
         self.vol_id = params.get('vol_id')
         self.generation = params.get('generation')
         self.prepared = params.get('prepared')
-        self.is_destination = is_destination
-        self.lock_image = lock_image
+        self.is_destination = dest is None
+
+        # Always lock the destination image and only lock the source if
+        # the copy is not within the same image.
+        self.lock_image = self.is_destination or self.img_id != dest.img_id
+
         self._host_id = host_id
         self._writable = writable
         self._vol = None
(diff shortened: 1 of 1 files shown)


tests/storage/sdm_copy_data_test.py
```diff
@@ -563,6 +563,49 @@ def test_copy_data_illegal(
...
+        vol_id=source_vol_id)
+    dest = dict(
+        endpoint_type='div',
+        sd_id=str(uuid.uuid4()),
+        img_id=source_img_id,
+        vol_id=dest_vol_id)
+
+    job = copy_data.Job(str(uuid.uuid4()), 0, source, dest)
+
+    assert not job._source.lock_image
```
nirs, line 586: Leta avoid testing private attributes. Instead test that the job succeeded, and the image was copied correctly.

We must have tests for:
- copy div to external
- copy external to div

The external endpoint can be a temporary image created by the test.

nirs (changes requested):

tests/storage/sdm_copy_data_test.py
bennyz: Added test from div to external, this test fails without the change in copy_data with the same KeyError. If the approach is sound I'll the external to div test

lib/vdsm/storage/sdm/api/copy_data.py
```diff
@@ -146,15 +141,17 @@ class CopyDataDivEndpoint(properties.Owner):
...
         self.vol_id = params.get('vol_id')
         self.generation = params.get('generation')
         self.prepared = params.get('prepared')
-        self.is_destination = is_destination
-        self.lock_image = lock_image
+        self.is_destination = dest is None
+        self.lock_image = (
+            self.is_destination
+            or dest.get('img_id') != self.img_id
+        )
```
nirs, line 154: The check is confusing but correct - the destination always lock the image, the source (dest != None) lock the image only if dest and source have different image id. Maybe worth a comment to explain this.

Need to update to use an endpoint object instead of dict

    getattr(dest, "img_id", None) != self.img_id

Or add img_id property to external endpoint - it can return vol_id or None since
external endpoint does not have an image in all cases. It is easier to have the
same interface instead of using getattr().

nirs (changes requested):

tests/storage/sdm_copy_data_test.py
```diff
@@ -563,6 +563,67 @@ def test_copy_data_illegal(
...
+        endpoint_type='div',
+        sd_id=src_vol.sdUUID,
+        img_id=src_vol.imgUUID,
+        vol_id=src_vol.volUUID)
+
+    dest_vol = str(uuid.uuid4())
+    lease_id = str(uuid.uuid4())
+    dest = dict(
+        endpoint_type='external',
+        url=f"file://{tmpdir.join(dest_vol)}",
```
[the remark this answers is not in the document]
bennyz: It failed originally, I'll recheck

tests/storage/sdm_copy_data_test.py
```diff
@@ -563,6 +563,67 @@ def test_copy_data_illegal(
     assert jobs.STATUS.DONE == job.status
 
 
+def test_copy_data_div_to_external(
```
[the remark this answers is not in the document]
bennyz: functional test will be added separately

bennyz:
Manual testing:
Disk conversion
Copy/move between DIV SDs
Copy between DIV and MBS domains (MBS -> DIV fails because of an engine regression, I will file a bug for it)

OST passed (has test for move and copy)

nirs:
@bennyz do we need a bug for this? the regression was released in 4.4.x?

Also is this a still a draft? seems like ready to review and tested to me.

lib/vdsm/storage/sdm/api/copy_data.py
```diff
@@ -146,15 +141,21 @@ class CopyDataDivEndpoint(properties.Owner):
...
         self.prepared = params.get('prepared')
-        self.is_destination = is_destination
-        self.lock_image = lock_image
+        self.is_destination = dest is None
+
+        # Always lock the destination image and only lock the source if
+        # the copy is not within the same image.
+        self.lock_image = (
+            self.is_destination
+            or getattr(dest, "img_id", None) != self.img_id
```
nirs, line 156: Since you ensure that external endpoint always have an img_id, we can simplify this to:

    self.lock_image = self.is_destination or self.img_id != dest.img_id

bennyz:
> @bennyz do we need a bug for this? the regression was released in 4.4.x?
> 
> Also is this a still a draft? seems like ready to review and tested to me.

I created a bug[1], but the regression is in 4.5 only


[1] https://bugzilla.redhat.com/show_bug.cgi?id=2066285

nirs (dismissed):
Would be useful to add the bug url to the commit message, so tools generating relases notes or change logs will pick this commit.

I plan to merge this later today.

nirs (approved):

PR #101, opened 2022-03-18
Support measuring subchain and active volumes

nirs:
This PR add support for measuring a sub-chain without the rest of
the backing chain:

    snap1 <- [snap2 <- snap3] <- active

The sub chain can include the active volume:

    snap1 <- [snap2 <- active]

Or measure the active volume without the backing chain:

    snap1 <- [active]

Measuring a sub chain gives an estimate for the size of the base
volume after the top volume is merged into it.

This will be used during live and cold merge to avoid over allocation
that cannot be fixed in live merge.

Engine will use during live storage migration when creating the destination
chain and for available size validation. Using measure will remove the need
to hard code vdsm internal details such as chunk size and qcow2 metadata 
overhead.

Bug-Url: https://bugzilla.redhat.com/2064907


lib/vdsm/storage/hsm.py
```diff
@@ -2857,13 +2857,19 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
...
+
         result = qemuimg.measure(
             vol.getVolumePath(),
             format=sc.fmt2str(vol.getFormat()),
             output_format=sc.fmt2str(dest_format),
             backing=backing,
             is_block=vol.is_block(),
-            base=base.getVolumePath() if base else None)
+            base=base.getVolumePath() if base else None,
+            unsafe=True)
```
bennyz, line 2872: 🎉

lib/vdsm/api/vdsm-api.yml
```diff
@@ -12082,6 +12082,14 @@ Volume.measure:
           or the entire chain.
         name: backing_chain
         type: boolean
+
+    -   defaultvalue: null
+        description: If specified, the base volume UUID, measuring the
```
vjuranek, line 12087: This is IMHO hard to understand. Maybe something like "If the baseID is specified, only the sub-chain will be measured. The range of sub-chain is from volumeID to baseID. Currently the only allowed value of baseID is parent volume UUID."?

lib/vdsm/storage/hsm.py
```diff
@@ -2828,19 +2829,41 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True):
             volUUID (str): The UUID of the volume you want to get the info on.
             dest_format (str): The output format we want to measure for
             backing (bool): True if we want to measure the volume with its
-                        backing chain, False otherwise. (Default: True)
+                backing chain, False otherwise. (Default: True)
+            baseUUID (str): If specified, the base volume UUID, measuring the
```
vjuranek, line 2833: same here

lib/vdsm/api/vdsm-api.yml
nirs: I will use your text, thanks!

lib/vdsm/storage/hsm.py
nirs: :+1:

nirs:
This version makes qemuio.open() more robust, it will wait 2.0 seconds before timing out.

Github UI does not have a way to compare custom version, but we can build the URL manually:
https://github.com/oVirt/vdsm/compare/771b9ced43035af8c64e13c34dece6149e8f385e..64db659c83059ef84f75cbdbed383590d4bc4c35

bennyz (approved):

tests/storage/qemuimg_test.py
```diff
@@ -1617,6 +1803,31 @@ def check_measure(self, filename, compat, format, compressed):
...
+                format=qemuimg.FORMAT.QCOW2,
+                output_format=qemuimg.FORMAT.QCOW2,
+                unsafe=True)
+
+        inactive = qemuimg.measure(
+            file_chain.top,
+            format=qemuimg.FORMAT.QCOW2,
+            output_format=qemuimg.FORMAT.QCOW2)
+
+        assert active == inactive
```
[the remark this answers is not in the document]
nirs: In general yes, you don't control the guest. But in this case there is no guest,
and we control the program with the write lock (qemu-io), so we know that the
no data is written during the test.

vjuranek (approved):

nirs:
Last push fixes conflict in host/caps.py after replicate_extend capability was merged.

PR #111, opened 2022-03-29
Improve mailbox tests

nirs:
- Verify roundtrip test results
- Refine mailbox roundtrip tests
- Add roundtrip test when events are disabled

### DIFF
--- a/tests/storage/mailbox_test.py
+++ b/tests/storage/mailbox_test.py
@@ -309,6 +309,34 @@ class TestCommunicate:
         sm.MESSAGES_PER_MAILBOX,
     ])
     def test_roundtrip(self, mboxfiles, delay, messages):
+        """
+        Test roundtrip latency.
+
+        This test is best run like this:
+
+            $ tox -e storage tests/storage/mailbox_test.py -- \
+                -k test_roundtrip \
+                --log-cli-level=info \
+                | grep stats
+
+        Example output (trimmed):
+
+            stats: messages=1 delay=0.000 best=0.215 average=0.215 worst=0.215
+            stats: messages=1 delay=0.050 best=0.234 average=0.234 worst=0.234
+            stats: messages=2 delay=0.000 best=0.231 average=0.233 worst=0.236
+            stats: messages=2 delay=0.050 best=0.252 average=0.285 worst=0.319
+            stats: messages=4 delay=0.000 best=0.257 average=0.258 worst=0.259
+            stats: messages=4 delay=0.050 best=0.267 average=0.311 worst=0.354
+            stats: messages=8 delay=0.000 best=0.231 average=0.344 worst=0.365
+            stats: messages=8 delay=0.050 best=0.246 average=0.314 worst=0.391
+            stats: messages=16 delay=0.000 best=0.244 average=0.379 worst=0.395
+            stats: messages=16 delay=0.050 best=0.148 average=0.312 worst=0.393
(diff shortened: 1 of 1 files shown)


tests/storage/mailbox_test.py
```diff
@@ -364,8 +364,17 @@ def reply_msg_callback(vol_data):
...
+        best = times[0]
+        worst = times[-1]
+        average = sum(times) / len(times)
+
+        log.info(
+            "stats: messages=%d delay=%.3f best=%.3f average=%.3f worst=%.3f",
+            messages, delay, best, average, worst)
+
+        assert best < 4 * EVENT_INTERVAL
+        assert average < 7 * EVENT_INTERVAL
```
nirs, line 376: Fails in github:

    >       assert average < 7 * EVENT_INTERVAL
    E       assert 0.7567226762539666 < (7 * 0.1)

We need larger timeouts.

vjuranek: This shows that these kind of tests are quite problematic. It would be good to add a comment where these numbers (4, 7, 10) comes from or what does it mean if this test fails.

nirs: Yes this test is problematic. If we don't test that the numbers makes sense, we can have performance regressions. If we test, the test will be flaky - when running on slow or overloaded machine the test will fail.

I think I will a comment showing how to run the tests and inspect the "stats" log, and make the asserts more forgiving for slow environment.

vjuranek (approved):

tests/storage/mailbox_test.py
```diff
@@ -357,13 +421,15 @@ def reply_msg_callback(vol_data):
                         # check that SPM inbox was cleared
                         if f.read(sm.MAILBOX_SIZE) == sm.EMPTYMAILBOX:
                             break
-                    assert time.time() < deadline, "Timeout clearing SPM inbox"
-                    time.sleep(0.1)
+                    if time.monotonic() < deadline:
```
barpavel, line 424: Whats' wrong with `assert`?

barpavel (approved):

tests/storage/mailbox_test.py
bennyz: >If a timeout expires, raise RuntimeError instead of assert. This is not a test failure but an error.

bennyz (approved):

tests/storage/mailbox_test.py
barpavel: OK :)
Since those are tests, I thought `assert` are appropriate :)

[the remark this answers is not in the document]
barpavel: Also just for my information. You changed to `monotonic()` to protect from summer --> winter --> summer clock transitions situations?
I thought `time.time()` is used widely and is enough.

nirs: assert has 2 issues:
- It confuse the reader that this is part of the test, but this check is not the purpose of the test
- The syntax does not allow long assert message. If you want to add a longer message it is much easier to do this by raising an exception.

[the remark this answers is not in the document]
nirs: time.time() is not affected by daylight saving changes, but it can go backwards in the system time changes. It is ok for tests but there is no reason to use it in python 3 since we have time.monotonic(). All code checking for timeouts should use time.monotonic().

Regarding the check - yes, it should should be:

    if time.monotonic() > deadline:
        fail...

barpavel (approved):

PR #113, opened 2022-03-31
Fix comments typos in '`storage.py`'

barpavel:
From going over the code of the `storage.py` file, I saw a few typos in the comments.
No-effort fix, improves searching (i.e., search by "chunked" word).

Signed-off-by: Pavel Bar <pbar@redhat.com>

### DIFF
--- a/lib/vdsm/virt/vmdevices/storage.py
+++ b/lib/vdsm/virt/vmdevices/storage.py
@@ -378,8 +378,8 @@ class Drive(core.Base):
     @property
     def replicaChunked(self):
         """
-        Return True if drive is replicating to chuked storage and the replica
-        volume may require extending. See Drive.chunkd for more info.
+        Return True if drive is replicating to chunked storage and the replica
+        volume may require extending. See Drive.chunked for more info.
         """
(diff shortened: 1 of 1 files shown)


vjuranek (dismissed):

lib/vdsm/virt/vmdevices/storage.py
```diff
@@ -453,7 +453,7 @@ def threshold_state(self, state):
     def needs_monitoring(self):
         """
         Return True if the drive needs to be picked by
-        a Drivemonitor periodic check, False otherwise.
+        a Drive monitor periodic check, False otherwise.
```
nirs, line 456: This should actually be:

    Return True if the drive needs to be picked by the VolumeMonitor
    periodic check, False otherwise.

nirs (changes requested):
Thanks!

lib/vdsm/virt/vmdevices/storage.py
barpavel: Done :)

nirs (approved):

PR #114, opened 2022-04-03
tests: Refine thinp stress test

nirs:
Make the thinp test script easier to use for testing, and nicer as
example for RHV demo.

- Add command line arguments so there is no need to edit the script.
- Fix short write - previously short write would be ignored, reporting
  incorrect actual rate and total size.
- Fix actual rate - previously did not consider throttling.
- Add correction time for throttling, considering the time to sleep and
  get the next timestamp in the next sleep.
- Round the actual rate for more stable results.

Example usage with this change:

    $ python tests/storage/stress/thinp.py --rate 1024 --size 10 thinp.data
    10.00 GiB, 10.00 s, 1024.1 MiB/s

### DIFF
--- a/tests/storage/stress/thinp.py
+++ b/tests/storage/stress/thinp.py
@@ -5,52 +5,73 @@ Usage:
 
 1. Create a 50g thin disk on block storage domain
 
-2. Update RATE to the desired write rate
-
-3. Update PATH for the actual device in the guest.
-
-4. Start watching extension completion logs on the host running the vm:
+2. Start watching extension completion logs on the host running the vm:
 
     $ tail -f /var/log/vdsm/vdsm.log | grep 'completed <Clock'
 
-5. Run inside the guest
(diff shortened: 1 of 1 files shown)


tests/storage/stress/thinp.py
```diff
@@ -5,52 +5,73 @@
```
vjuranek, line 5: > Add command line arguments to there is no need to edit the script.

"to" -> "so"

> Add correction time for throttling, fixing considering the time to  sleep and get the next timestamp in the next sleep.

either "fixing" or "considering"

vjuranek (commented):
typos in commit message, otherwise LGTM

tests/storage/stress/thinp.py
nirs: Thanks, fixed.

vjuranek (approved):

PR #115, opened 2022-04-04
fileVolume: Fix creation of tiny volumes

nirs:
When creating preallocated volume with size <= 4096, we try to allocate
0 bytes:

    /usr/libexec/vdsm/fallocate --offset 4096 0 /path...

Which fail in fallocate() with EINVAL as expected, and then we raise

    VolumesZeroingError: Cannot zero out volume: /path...

It looks like engine does not handle this error correctly, and the disk
remain locked forever.

Creating a volume of 4k bytes or less is not a real use case, but there
is no technical reason why it should fail. Fixed by skipping the
allocation if we don't have anything to allocate.

Reported-by: Janos Bonic <jpasztor@redhat.com>

### DIFF
--- a/lib/vdsm/storage/fileVolume.py
+++ b/lib/vdsm/storage/fileVolume.py
@@ -568,7 +568,7 @@ class FileVolume(volume.Volume):
             # If the image is preallocated, allocate the rest of the image
             # using fallocate helper. qemu-img create always writes zeroes to
             # the first block so we should skip it during preallocation.
-            if preallocate == sc.PREALLOCATED_VOL:
+            if size > 4096 and preallocate == sc.PREALLOCATED_VOL:
                 op = fallocate.allocate(vol_path, size - 4096, offset=4096)
 
                 # This is fast on NFS 4.2, GlusterFS, XFS and ext4, but can be


nirs:
Reproduced with upload disk:

Created a tiny disk:
```
truncate -s 512 /var/tmp/empty-512.raw
```

Upload
```
$ ./upload_disk.py -c engine --sd-name alpine-nfs-01 /var/tmp/empty-512.raw 
[   0.0 ] Checking image...
[   0.0 ] Image format: raw
[   0.0 ] Disk format: raw
[   0.0 ] Disk content type: data
[   0.0 ] Disk provisioned size: 512
[   0.0 ] Disk initial size: 512
[   0.0 ] Disk name: empty-512.raw
[   0.0 ] Disk backup: False
[   0.0 ] Connecting...
[   0.0 ] Creating disk...
^CTraceback (most recent call last):
  File "/home/nsoffer/src/python-ovirt-engine-sdk4/examples/./upload_disk.py", line 254, in <module>
    time.sleep(1)
KeyboardInterrupt
```

Upload stuck for ever, interrupted after 15 minutes or so.

With the fix we can upload this disk:
```
$ ./upload_disk.py -c engine --sd-name alpine-nfs-01 /var/tmp/empty-512.raw 
[   0.0 ] Checking image...
[   0.0 ] Image format: raw
[   0.0 ] Disk format: raw
[   0.0 ] Disk content type: data
[   0.0 ] Disk provisioned size: 512
[   0.0 ] Disk initial size: 512
[   0.0 ] Disk name: empty-512.raw
[   0.0 ] Disk backup: False
[   0.0 ] Connecting...
[   0.0 ] Creating disk...
[  15.5 ] Disk ID: 4cfaa59b-7b22-4ace-9fce-835124e91db0
[  15.5 ] Creating image transfer...
[  16.9 ] Transfer ID: d4ad3372-38b7-4a43-a12a-4a1bf44b8a77
[  16.9 ] Transfer host name: host4
[  16.9 ] Uploading image...
[ 100.00% ] 512 bytes, 0.08 seconds, 6.06 KiB/s                                
[  17.0 ] Finalizing image transfer...
[  20.0 ] Upload completed successfully
```

And download the disk:
```
$ ./download_disk.py -c engine 4cfaa59b-7b22-4ace-9fce-835124e91db0 --format raw /var/tmp/download.raw[   0.0 ] Connecting...
[   0.1 ] Creating image transfer...
[   1.4 ] Transfer ID: a30c38be-70cb-48e3-a05a-0d4ece0814d9
[   1.4 ] Transfer host name: host4
[   1.4 ] Downloading disk...
[ 100.00% ] 4.00 KiB, 0.09 seconds, 43.14 KiB/s                                
[   1.5 ] Finalizing image transfer...
[   4.5 ] Download completed successfully
```

As expected, the disk size was increased to 4k. We don't support creating
volumes smaller than 4k.

lib/vdsm/storage/fileVolume.py
```diff
@@ -568,7 +568,7 @@ def _allocate_volume(cls, vol_path, size, preallocate):
             # If the image is preallocated, allocate the rest of the image
             # using fallocate helper. qemu-img create always writes zeroes to
             # the first block so we should skip it during preallocation.
-            if preallocate == sc.PREALLOCATED_VOL:
+            if size > 4096 and preallocate == sc.PREALLOCATED_VOL:
```
barpavel, line 571: IMHO using `BLOCK_SIZE_4K` instead of 4096 is better (3 times).

barpavel (approved):

lib/vdsm/storage/fileVolume.py
nirs: Not in this case, it just make the code more verbose and it can cause a real bug
since this code depends on qemu-img behavior to allocated 4096 bytes when creating
a new image.

barpavel: No problem.
From the comment `writes zeroes to the first block` I assumed this is always a 4096 bytes' block.
Maybe a more correct solution (even if not in this patch, if you don't want refactoring here) could be to define other constant like `QEMU_IMG_ZERO_BLOCK_SIZE` or something, and to use it.
Anyway, you already have my approval :)

vjuranek (approved):

bennyz (approved):

lib/vdsm/storage/fileVolume.py
nirs: Constant like QEMU_IMG_ZERO_BLOCK_SIZE or QEMU_IMG_MIN_ALLOC is the right way to handle this. Feel free to post a PR cleaning this up.

PR #118, opened 2022-04-07
network: use nmstate for bridge options

hbraha:
the bridge options are currently implemented via sysfs,
We want to use nmstate for bridge options.

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,9 +1,14 @@
-.PHONY: installdeps srpm
+.PHONY: installdeps git_cfg_safe srpm
 
 installdeps:
 	dnf -y install autoconf automake gcc gettext-devel git libtool make openssl python3-dateutil python3-devel python3-libvirt python3-pyyaml python3-six systemd-units util-linux
 
-srpm: installdeps
+git_cfg_safe:
+	# From git 2.35.2 we need to mark temporary directory, where the project is cloned to, as safe, otherwise
+	# git commands won't work because of the fix for CVE-2022-24765
+	git config --global --add safe.directory "$(shell pwd)"
+	
+srpm: installdeps git_cfg_safe
 	$(eval SUFFIX=$(shell sh -c " echo '.$$(date -u +%Y%m%d%H%M%S).git$$(git rev-parse --short HEAD)'"))
 	mkdir -p tmp.repos
 	./autogen.sh \
--- a/.github/workflows/network.yml
+++ b/.github/workflows/network.yml
@@ -26,7 +26,12 @@ jobs:
     strategy:
       matrix:
         type: [ unit, integration, functional ]
-        tag: [ centos-8, centos-9 ]
+        tag: [ centos-8 ]
+        include:
+          - type: unit
+            tag: centos-9
+          - type: integration
+            tag: centos-9
     steps:
       - uses: actions/checkout@v2
       - name: Install dependencies
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ Makefile
 Makefile.in
 aclocal.m4
 autom4te.cache
+build
 build-aux/compile
 build-aux/config.guess
 build-aux/config.sub
--- a/Makefile.am
+++ b/Makefile.am
@@ -166,6 +166,10 @@ all-local: \
 
 clean-local: clean-build
 
+.PHONY: upgrade
+upgrade:
+	sudo dnf upgrade $(RPM_TOPDIR)/RPMS/*/*.rpm
+
 .PHONY: srpm rpm clean-build
 
 srpm: clean-build dist
--- a/Makefile.am
+++ b/Makefile.am
@@ -35,6 +35,9 @@ SUBDIRS = \
 # prepared before running.
 SUBDIRS += tests
 
+VENV_DIR = $(HOME)/.venv/vdsm
+VENV = $(VENV_DIR)/bin
+
 include $(top_srcdir)/build-aux/Makefile.subs
 
 # This is an *exception*, we ship also vdsm.spec so it's possible to build the
@@ -134,6 +137,12 @@ check-unit:
 
 .PHONY: lint
 lint: gitignore execcmd black flake8 pylint
+
+.PHONY: venv
+venv:
+	python3 -m venv $(VENV_DIR) && \
+	$(VENV)/python3 -m pip install --upgrade pip && \
+	$(VENV)/python3 -m pip install -r docker/requirements.txt
 
 .PHONY: tests
 tests: tox
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,7 @@ SUBDIRS += tests
 
 VENV_DIR = $(HOME)/.venv/vdsm
 VENV = $(VENV_DIR)/bin
+RPM_TOPDIR = $(PWD)/build
 
 include $(top_srcdir)/build-aux/Makefile.subs
 
@@ -163,14 +164,18 @@ clean-storage:
 all-local: \
 	vdsm.spec
 
-.PHONY: srpm rpm
+clean-local: clean-build
 
-srpm: dist
+.PHONY: srpm rpm clean-build
+
+srpm: clean-build dist
 	rpmbuild -ts $(if $(BUILDID),--define="extra_release .$(BUILDID)") \
+		--define="_topdir $(RPM_TOPDIR)" \
 		$(DIST_ARCHIVES)
 
-rpm: dist
+rpm: clean-build dist
 	rpmbuild -ta $(if $(BUILDID),--define="extra_release .$(BUILDID)") \
+		--define="_topdir $(RPM_TOPDIR)" \
 		--define="qemu_user $(QEMUUSER)" \
 		--define="qemu_group $(QEMUGROUP)" \
 		--define="with_ovirt_vmconsole $(OVIRT_VMCONSOLE)" \
@@ -180,6 +185,9 @@ rpm: dist
 		--define="vdsm_version $(PACKAGE_VERSION)" \
 		--define="vdsm_release $(PACKAGE_RELEASE)" \
 		$(DIST_ARCHIVES)
+
+clean-build:
+	rm -rf $(RPM_TOPDIR)
 
 dist-hook: gen-VERSION gen-ChangeLog
 .PHONY: gen-VERSION gen-ChangeLog
--- a/automation/rpm.sh
+++ b/automation/rpm.sh
@@ -20,7 +20,7 @@ mkdir -p ${EXPORT_DIR}
 
 cp $PWD/lib/vdsm/api/vdsm-api.html "${EXPORT_DIR}"
 
-find ~/rpmbuild \
+find $PWD/build \
     -iname '*.rpm' \
     -exec mv {} "${EXPORT_DIR}/" \;
 
--- a/configure.ac
+++ b/configure.ac
@@ -401,6 +401,7 @@ AC_OUTPUT([
 	vdsm_hooks/fileinject/Makefile
 	vdsm_hooks/httpsisoboot/Makefile
 	vdsm_hooks/localdisk/Makefile
+	vdsm_hooks/log_console/Makefile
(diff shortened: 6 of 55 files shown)


hbraha:
two current issues with the patch are:
1. nmstate doesnt pick up the "multicast_router" option when passed.
2. all options that uses bool failing on `_assert_kernel_vs_running_config` since the expected value from the kernel is int but the value passed by nmstate is bool. for example:
```E            'test-network': {'bootproto': 'none',
E                             'bridged': True,
E         -                   'custom': {'bridge_opts': 'multicast_snooping=0'},
E         ?                                                                 ^
E         +                   'custom': {'bridge_opts': 'multicast_snooping=False'},
E         ?                                                                 ^^^^^
```

@almusil have any ideas how should we continue ?

nirs:
@hbraha You need to make black happy. I think formatting the code with black before
committing will solve the lint failure.

lib/vdsm/network/link/setup.py
```diff
@@ -55,3 +74,15 @@ def parse_nets_bridge_opts(nets):
...
             )
             yield name, bridge_opts
+
+
+def parse_nets_bridge_opts_for_nmstate(opts):
+    if opts:
+        opts_str = opts.replace('_', '-')
+        bridge_opts = dict(
+            opt.split('=', 1) for opt in opts_str.split(' ')
+        )
```
almusil, line 84: This part is very similar to `_string_to_dict` above. Please resuse that function.

lib/vdsm/network/link/setup.py
```diff
@@ -55,3 +74,15 @@ def parse_nets_bridge_opts(nets):
                 opt.split('=', 1) for opt in opts_str.split(' ')
             )
             yield name, bridge_opts
+
+
```
almusil, line 78: Eh cannot comment that directly on that code. Please remove `setup_custom_bridge_opts`.

almusil, line 78: The `parse_nets_bridge_opts` can be updated and used instead of creating a new function.

lib/vdsm/network/nmstate/linux_bridge.py
```diff
@@ -34,6 +34,7 @@
 from .schema import LinuxBridge
 from .schema import Vlan
 from .state import NetworkingState
+from ..link.setup import parse_nets_bridge_opts_for_nmstate
```
almusil, line 37: For imports beyond `nmstate` module please use absolute imports.

tests/network/functional/bridge_test.py
```diff
@@ -71,7 +71,7 @@ def test_add_bridge_with_custom_opts(self, adapter, switch, nic0):
             'nic': nic0,
             'switch': switch,
             'custom': {
-                'bridge_opts': 'multicast_snooping=0 multicast_router=0'
+                'bridge_opts': 'multicast_snooping=False'
```
almusil, line 74: This must be compatible with the old way, that's also the reason why it fails the validation step in tests.

lib/vdsm/network/link/setup.py
```diff
@@ -55,3 +74,15 @@ def parse_nets_bridge_opts(nets):
...
+def parse_nets_bridge_opts_for_nmstate(opts):
+    if opts:
+        opts_str = opts.replace('_', '-')
+        bridge_opts = dict(
+            opt.split('=', 1) for opt in opts_str.split(' ')
+        )
+        for key in bridge_opts.keys():
+            bridge_opts[key] = nmstate_options_types[key](eval(bridge_opts[key]))
```
almusil, line 86: Why is the `eval` needed?

almusil (changes requested):
We should probably consider what to do with options that are not supported from nmstate side.

tests/network/functional/bridge_test.py
hbraha: you mean we should pass `0` or `1` and convert it to True or False if needed?

lib/vdsm/network/link/setup.py
hbraha: because `bridge_opts` values are strings and nmstate seem to accept int/bool

tests/network/functional/bridge_test.py
almusil: Yes

lib/vdsm/network/link/setup.py
almusil: Having eval on user input is huge security risk, if `bool` is not working we need to create our own function that will do the conversion.

lib/vdsm/network/link/setup.py
```diff
@@ -23,35 +23,49 @@
...
+    'multicast-querier-interval': int,
+    'multicast-query-use-ifaddr': bool,
+    'multicast-query-interval': int,
+    'multicast-query-response-interval': int,
+    'multicast-startup-query-count': int,
+    'multicast-startup-query-interval': int,
+}
+
+
+def _string_to_dict(str, div, eq):
```
almusil, line 45: It's good practice to have helpers after all public functions + for better readiness having the function defined after all callers. That way you can read the file from top to bottom.

lib/vdsm/network/link/setup.py
```diff
@@ -23,35 +23,49 @@
...
-
-
-def parse_nets_bridge_opts(nets):
-    for name, opts in nets.items():
-        opts_str = opts.get('custom', {}).get('bridge_opts')
-        if opts_str:
-            bridge_opts = dict(
-                opt.split('=', 1) for opt in opts_str.split(' ')
+def parse_nets_bridge_opts(opts):
+    if opts:
```
almusil, line 63: nit: To avoid multiple nesting you could do `if not opts: return {}`

lib/vdsm/network/nmstate/linux_bridge.py
```diff
@@ -34,6 +34,7 @@
 from .schema import LinuxBridge
 from .schema import Vlan
 from .state import NetworkingState
+from vdsm.network.link.setup import parse_nets_bridge_opts
```
almusil, line 37: Please put the import above the relative ones with blank line in between.

lib/vdsm/network/link/setup.py
```diff
@@ -23,35 +23,49 @@
 from vdsm.network.link.bridge import Bridge
 
 
+nmstate_options_types = {
```
almusil, line 26: Please make the name capital

almusil (changes requested):
Also there are couple of tests that will need some adjustments, also this deserves some new unit test on it's own, somewhere in the `linux_bridge_test.py` module.

lib/vdsm/network/link/setup.py
```diff
@@ -20,38 +20,64 @@
...
     else:
         return {}
 
 
-def setup_custom_bridge_opts(nets):
-    for name, opts in parse_nets_bridge_opts(nets):
-        Bridge(name, opts)
+def parse_nets_bridge_opts(opts):
 
+    nmstate_options_types = {
```
almusil, line 41: This is still a constant, it should have capital name and better place is under imports.

lib/vdsm/network/link/setup.py
```diff
@@ -20,38 +20,64 @@
...
+    if not opts:
+        return {}
+    opts_str = opts.replace('_', '-')
+    bridge_opts = _string_to_dict(opts_str, ' ', '=')
+    try:
+        for key in bridge_opts.keys():
+            bridge_opts[key] = nmstate_options_types[key](bridge_opts[key])
+    except KeyError:
+        raise NmstateBridgeOptionNotSupported(
+            f'{key} is not a valid nmstate birdge option'
```
almusil, line 68: nit: `bridge`

lib/vdsm/network/link/setup.py
```diff
@@ -20,38 +20,64 @@
...
+        )
+    return bridge_opts
+
+
+def _str_to_bool(str_value):
+    false_values = ['false', 'False', '0']
+    if str_value in false_values:
```
almusil, line 75: `return str_value not in false_values`

lib/vdsm/network/link/setup.py
```diff
@@ -20,38 +20,64 @@
 from __future__ import absolute_import
 from __future__ import division
 
-from vdsm.network.link.bridge import Bridge
```
almusil, line 23: It looks like the `link.bridge` module is not used anywhere, it can be completely removed. Probably as separate commit.

lib/vdsm/network/nmstate/linux_bridge.py
```diff
@@ -18,6 +18,7 @@
 #
 
 from collections import defaultdict
```
almusil, line 20: nit: Missing empty line.

tests/network/unit/link_setup_test.py
```diff
@@ -37,9 +37,12 @@ def test_parse_nets_bridge_opts():
...
-        'br2': {'multicast_router': '1', 'multicast_snooping': '1'},
+        'br1': {'multicast-router': 0, 'multicast-snooping': False},
+        'br2': {'multicast-router': 1, 'multicast-snooping': True},
     }
 
-    for name, opts in linksetup.parse_nets_bridge_opts(nets):
-        assert expected[name] == opts
+    for name, opts in nets.items():
+        parsed_opts = linksetup.parse_nets_bridge_opts(
+            opts['custom']['bridge_opts']
```
almusil, line 46: Is there a reason to keep the `custom` and `bridge_opts` nesting around?

lib/vdsm/network/link/setup.py
```diff
@@ -20,38 +20,64 @@
...
+            bridge_opts[key] = nmstate_options_types[key](bridge_opts[key])
+    except KeyError:
+        raise NmstateBridgeOptionNotSupported(
+            f'{key} is not a valid nmstate birdge option'
+        )
+    return bridge_opts
+
+
+def _str_to_bool(str_value):
+    false_values = ['false', 'False', '0']
```
almusil, line 74: We could save one value in the array by using `str_value.lower()`, but not that important.

almusil (changes requested):
Some minor things other than that it looks good, thanks

lib/vdsm/network/link/setup.py
hbraha: let's handle this after this PR will be merged.

lib/vdsm/network/link/setup.py
hbraha: the problem was when it was is under imports it didn't recognize `str_to_bool` func. anyway, i think now it's more readable and well encapsulated.

tests/network/unit/link_setup_test.py
hbraha: not sure what you mean here, can you elaborate?

almusil: Since we are parsing only the string of the options now, we don't need to create the whole nets dictionary. The `nets` can be reduced to array same with expected.

lib/vdsm/network/link/setup.py
```diff
@@ -20,38 +20,74 @@
...
-        Bridge(name, opts)
+def parse_nets_bridge_opts(opts):
+
+    if not opts:
+        return {}
+    opts_str = opts.replace('_', '-')
+    bridge_opts = _string_to_dict(opts_str, ' ', '=')
+    try:
+        for key, value in bridge_opts.items():
+            bridge_opts[key] = NmstateBridgeOpts().convert_value(key, value)
```
almusil, line 82: Now we are calling init for every loop run, please create it before the for as it is constant that can be reused.

almusil (changes requested):

almusil (approved):
LGTM, thanks

hbraha:
unfortunately, this PR is blocked for now by nmsatate bug that doesn't allow disable multicast-router on linux-bridge options. 

https://bugzilla.redhat.com/show_bug.cgi?id=2076590

erav:
wait for nmstate-1.2.1-3 to be published on RHEL 8.6 Z-stream repos and then see which RHV version is aligned with it and merged this PR into this RHV version

PR #133, opened 2022-04-13
Fix over allocation when preparing merge

### LINKED ISSUE #134: Merge over extends the base volume during the merge
Before merge, the base volume is extended too much, using a dumb way:

    new_size = top_size + base_size + chunk_size

This is the worst possible case, when both top and base have no free space, and all clusters in top are not in base. But in the common case, some clusters in top are already in base, so the
needed allocation is much smaller.

After the merge, the base volume is shrinked to optimal size, release the unneeded space to the system.

Because too much space was allocated, we can have these issues during the merge:

- The storage domain can become full - VMs will pause when they don't have enough space to extend thin disks
- Other storage operations to fail, because there is not enough free space
- Preparing the merge will fail, because there is not enough space to extend base

## How to reproduce

1. Create VM with 30g thin disk
2. In the guest, write the first 10g

       # dd if=/dev/zero bs=8M count=1280 of=/dev/sdb oflag=direct

4. Create a snapshot including the 30g disk. This creates a new image with actual size of 2.5g.
5. In the guest, write again the first 10g

       # dd if=/dev/zero bs=8M count=1280 | tr "\0" "\1" > /dev/sdb; sync

   This extends the top image to 12.5g

6. Shut down the vm
7. Delete snapshot in engine UI

On the SPM, the base volume is extended to 27.5g:

    2022-04-13 18:49:49,004+0300 INFO  (tasks/7) [storage.lvm] Extending LV bafd0f16-9aba-4f9f-ba90-46d3b8a29157/9106b4de-bb36-4d3f-86ea-76bedbab842f to 28160 megabytes (lvm:1616)

After the merge completes, the image is reduced to optimal size: 10243 MiB (10.002 GiB)

    2022-04-13 18:50:20,767+0300 INFO  (tasks/6) [storage.lvm] Reducing LV bafd0f16-9aba-4f9f-ba90-46d3b8a29157/9106b4de-bb36-4d3f-86ea-76bedbab842f to 10243 megabytes (force=False) (lvm:1654)

This request is aligned to lvm extent size (128 MiB):

    # lvs --devicesfile= -o vg_name,lv_name,tags,size | grep f1fd2ae0-3217-4ef7-b870-fe5469dd3fa2
  bafd0f16-9aba-4f9f-ba90-46d3b8a29157 9106b4de-bb36-4d3f-86ea-76bedbab842f IU_f1fd2ae0-3217-4ef7-b870-fe5469dd3fa2,MD_7,PU_00000000-0000-0000-0000-000000000000  10.12g

Summary:

1. The size required for merging was 10.12g
2. The LV was resized to 27.5g during the merge.

nirs:
When preparing for merge, we extend the base volume in a dumb way, allocating too much space. We clean up when finalizing the merge, so the volume size is optimal when the merge is done. However during the merge the over allocation can cause the storage domain to become full which can cause vms to fail to extend their disks, or to other storage operation to because storage space is low.

This change fix the issue by measuring the required size for the merge, and calculating the optimal size of the volume based on the required size. When the merge is finalized the volume should be in the optimal size so we don't need to shrink the volume.

Fixes #134

### DIFF
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -428,6 +428,10 @@ class BlockVolumeManifest(volume.VolumeManifest):
             # minimal volume size.
             optimal_size = max(actual_size + MIN_PADDING, sc.MIN_CHUNK)
 
+        # Align to align size so callers can compare optimal size with the
+        # actual size.
+        optimal_size = utils.round(optimal_size, self.align_size)
+
         # Limit by maximum size.
         max_size = self.max_size(self.getCapacity(), self.getFormat())
         optimal_size = min(optimal_size, max_size)
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -421,8 +421,10 @@ class BlockVolumeManifest(volume.VolumeManifest):
                 "irs", "volume_utilization_chunk_mb") * MiB
             optimal_size = actual_size + free_space
         else:
-            # For internal volumes, extend to minimal volume size.
-            optimal_size = max(actual_size, sc.MIN_CHUNK)
+            # For internal volumes, use the actual size. It cannot be zero
+            # since it includes qcow2 metadata.
+            assert actual_size > 0
+            optimal_size = actual_size
 
         # Align to align size so callers can compare optimal size with the
         # actual size.
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -397,6 +397,43 @@ class BlockVolumeManifest(volume.VolumeManifest):
             if pvolUUID != sc.BLANK_UUID:
                 cls.teardown(sdUUID=sdUUID, volUUID=pvolUUID, justme=False)
 
+    @classmethod
+    def optimal_cow_size(cls, required_size, capacity, is_leaf):
+        """
+        Return the optimal size for a volume based on the required allocation,
+        volume capacity, and volume type.
+
+        This is a class method so we can use this calculation for the base
+        volume during merge, using the mesured size for the merged instead of
+        the actual volume size.
+        """
+        # Add free space.
+        if is_leaf:
+            # For leaf volumes, add one chunk of free space so the VM
+            # will not pause quickly when it is started.
+            free_space = config.getint(
+                "irs", "volume_utilization_chunk_mb") * MiB
+            optimal_size = required_size + free_space
+        else:
+            # For internal volumes, use the required size. It cannot be zero
+            # since it includes qcow2 metadata.
+            assert required_size > 0
+            optimal_size = required_size
+
+        # Align to align size so callers can compare optimal size with the
+        # actual size.
+        optimal_size = utils.round(optimal_size, cls.align_size)
+
+        # Limit by maximum size.
+        max_size = cls.max_size(capacity, sc.COW_FORMAT)
+        optimal_size = min(optimal_size, max_size)
+
+        cls.log.debug(
+            "COW format, required_size=%s max_size=%s optimal_size=%s",
+            required_size, max_size, optimal_size)
+
+        return optimal_size
+
     def optimal_size(self):
         """
         Return the optimal size of the volume, based on actual allocation.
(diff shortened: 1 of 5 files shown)


vjuranek (dismissed):

lib/vdsm/storage/merge.py
```diff
@@ -207,18 +205,29 @@ def _extend_base_allocation(base_vol, top_vol):
...
+        base=base_vol.getVolumePath())
+    log.debug("Measure result: %s", measure)
+
+    # When merging we always copy the bitmaps from the top to base. Measure
+    # gives us the size of the bitmaps in top *and* base, so this may allocate
+    # more than needed, but bitmaps are small so it should be good enough.
+    required_size = measure["required"] + measure.get("bitmaps", 0)
+
+    optimal_size = base_vol.optimal_cow_size(
+        required_size, base_vol.getCapacity(), base_vol.isLeaf())
```
nirs, line 226: Testing reveal that this is wrong, should be top_vol.isLeaf().

We need to improve the test to test both internal and active merge.

vjuranek (approved):

tests/storage/blockvolume_test.py
```diff
@@ -160,7 +162,7 @@ def test_optimal_size_cow_leaf_max(self):
         (512 * MiB, 200 * MiB, 640 * MiB),
         # Add minimum padding.
         (2 * GiB, 1023 * MiB, 1024 * MiB),
-        (2 * GiB, 1024 * MiB, 1024 * MiB + blockVolume.MIN_PADDING),
+        (2 * GiB, 1024 * MiB, 1152 * MiB),
```
aesteve-rh: I'm not a big fan of inline comments, but here maybe helps to have an explanation of what is tested with the optimal size of this permutation?

nirs: Yes, we need a comment for each case.

lib/vdsm/storage/blockVolume.py
```diff
@@ -421,8 +421,8 @@ def optimal_size(self):
                 "irs", "volume_utilization_chunk_mb") * MiB
             optimal_size = actual_size + free_space
         else:
-            # For internal volumes, extend to minimal volume size.
-            optimal_size = max(actual_size, sc.MIN_CHUNK)
+            # For internal volumes, use the actual size.
+            optimal_size = actual_size
```
nirs: Need to ensure size >= self.align_size here.

tests/storage/blockvolume_test.py
```diff
@@ -159,7 +159,7 @@ def test_optimal_size_cow_leaf_max(self):
     @permutations([
         # virtual_size, actual_size, optimal_size
         # Limited by max size.
-        (512 * MiB, 200 * MiB, 640 * MiB),
+        (512 * MiB, 200 * MiB, 256 * MiB),
```
nirs, line 169: Missing test for the the case of empty source image. This should not happen since with qcow2 image empty image needs about 200k (for qcow2 header), but may happen with empty raw sparse image.

tests/storage/blockvolume_test.py
```diff
@@ -144,9 +144,12 @@ def test_optimal_size_cow_leaf(self):
         with self.make_volume(size=2 * GiB, format=sc.COW_FORMAT) as vol:
             chunk_size = GiB
             check = qemuimg.check(vol.getVolumePath(), qemuimg.FORMAT.QCOW2)
-            actual_size = utils.round(
+            optimal_size = utils.round(
                 check['offset'] + chunk_size, vol.align_size)
-            self.assertEqual(vol.optimal_size(), actual_size)
+            self.assertEqual(vol.optimal_size(), optimal_size)
+            self.assertEqual(
+                vol.optimal_cow_size(check["offset"], 2 * GiB, True),
```
aesteve-rh, line 151: Something like this would be clearer?
```
vol.optimal_cow_size(check["offset"], capacity=2 * GiB, is_lead=True),
```
At least having a name for the boolean literal parameter at the end would help, so that we don't have to scroll up to the definition to see what it means in this position.

nirs: Yes, is_leaf= is certainly needed here.

lib/vdsm/storage/blockVolume.py
nirs: The next line rounds optimal size to align size, and optimal size cannot be zero since actual_size includes qcow2 metadata size.

tests/storage/blockvolume_test.py
nirs: For block raw, actual size is the virtual size, so the only missing test is empty qcow2 image.

```
$ qemu-img create -f qcow2 empty-100g.qcow2 100g
Formatting 'empty-100g.qcow2', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=107374182400 lazy_refcounts=off refcount_bits=16

$ qemu-img check empty-100g.qcow2 
Image end offset: 262144
```

tests/storage/blockvolume_test.py
nirs: While is_leaf=True is more clear, it treats the argument as keyword argument, and we try to always call positional arguments as:

    func(arg1, arg2)

and keyword arguments as:

    func(key=value)

This way code never breaks when keyword argument order is changed (e.g. adding new keyword argument), or when renaming positional argument. Both changes are valid when callers use these rules.

So base on our current conventions, if we want to use is_leaf=True, we need to add default value, which make it more fragile - easy to call with the missing argument and get the wrong result.

nirs: Improved by using `vol.isLeaf()` instead of True which is also more similar to the actual code.

vjuranek (approved):

tests/storage/blockvolume_test.py
```diff
@@ -169,7 +172,8 @@ def fake_check(path, format):
             # fake qemuimg check to return big volumes size, instead of writing
             # big data to volumes, an operation that takes long time.
             with MonkeyPatchScope([(qemuimg, 'check', fake_check)]):
-                env.chain = make_qemu_chain(env, actual_size, sc.COW_FORMAT, 3)
+                env.chain = make_qemu_chain(
+                    env, virtual_size, sc.COW_FORMAT, 3)
```
aesteve-rh, line 187: Maybe give a key name for the magic number argument?

```python
                env.chain = make_qemu_chain(
                    env, virtual_size, sc.COW_FORMAT, chain_len=3)
```

aesteve-rh (approved):

PR #142, opened 2022-04-19
Support Ceph mount using DNS name

KarlJorgensen:
Case in point: For Ceph file systems, the mount command can look like this:

    mount -t ceph -o mds_namespace=ns1,name=ovirt,secret=xxx monitors.storage.example.com:/ /mnt

but the resulting entry in /proc/mounts will NOT look identical, as ceph will resolve DNS names into their IP addresses:

    192.168.1.5,192.168.1.6,192.168.1.7:/ /mnt ceph rw,seclabel,relatime,name=ovirt,secret=<hidden>,acl,mds_namespace=ovirt 0 0

(in this example `monitors.storage.example.com` was a DNS entry
resolving to 3 IP addresses).

Lesson learned: We cannot rely on mount options being carried
unmolested into /proc/mounts...

Bug-Url: https://bugzilla.redhat.com/1995610

### DIFF
--- a/lib/vdsm/storage/mount.py
+++ b/lib/vdsm/storage/mount.py
@@ -231,7 +231,13 @@ class Mount(object):
             fs_specs = self.fs_spec, None
 
         for record in _iterMountRecords():
-            if self.fs_file == record.fs_file and record.fs_spec in fs_specs:
+            if self.fs_file == record.fs_file:
+                # Note: We cannot compare fs_spec, as some file
+                # systems will record _different_ information in
+                # /proc/mounts than was given to the mount command...
+                #
+                # E.g. ceph will resolve DNS names given into IP
+                # addresses to be presented in /proc/mounts...
                 return record
 
         raise OSError(errno.ENOENT,


lib/vdsm/storage/mount.py
```diff
@@ -248,7 +248,13 @@ def getRecord(self):
...
 
         for record in _iterMountRecords():
-            if self.fs_file == record.fs_file and record.fs_spec in fs_specs:
+            if self.fs_file == record.fs_file:
+                # Note: We cannot compare fs_spec, as some file
+                # systems will record _different_ information in
+                # /proc/mounts than was given to the mount command...
+                #
+                # E.g. ceph will resolve DNS names given into IP
+                # addresses to be presented in /proc/mounts...
```
nirs, line 257: There are 2 issues with this change:
1. It affects all kinds of mounts to solve a ceph specific issue. We need a fix that does not change the system behavior with other mounts.
2. Even for ceph mount, this is incorrect. With this change if you have 2 ceph mounts, one mounted and one not, mounting the second mount will fail since we will consider it as mounted, comparing only the remote file part "/", which is same in both mounts.

I think that what we learn from this is not that we cannot expect to find the mount options in /proc/mounts, but that we cannot work with ceph mounts since they behave in a non-standard way.

We need a way to detect if a mount is mounted. The current code accept a "server:path" format and mount it at /rhev/data-center/mnt/srever:path. Since mounts can become stale, and can be unmounted, we must make sure that /rhev/data-center/mnt/srever:path is mounted, and that the remove server and path match.

In summary, a fix must not change anything for other storage, and provide a good way to match the mount with the server:path. A possible solution can be resolving the DNS name before the mount so the local mount path will match /proc/mounts. Or maybe a change in ceph to disable resolving before the mount.

This change also break the tests:
https://github.com/oVirt/vdsm/actions/runs/2192442081

nirs (changes requested):

dupondje:
No follow-up since long time.
Closing, feel free to re-open if you want to continue working on this!

Issue #143, opened 2022-04-21
Improve TestExecCmd in misc_test for output handling


aesteve-rh:
### Description

In misc_tests, the executed commands in `TestExecCmd` do not consider the return code nor all the outputs to determine if the test should fail. Considering them not only will increase the test coverage, but will provide useful output for debugging.

Also, `checkSudo` function is not working as expected, resulting in runs where the test should be skipped because the user is not root, but it is instead executed and fails as a result.

### Traceback

```bash
$ tox -e storage -- -vv storage/misc_test.py::TestExecCmd::testSudo
...
self = <storage.misc_test.TestExecCmd testMethod=testSudo> 
    def testSudo(self):
        """
        Tests that when running with sudo the user really is root (or other
        desired user). 
        """ 
        cmd = [EXT_WHOAMI]
        checkSudo(cmd)
        ret, stdout, stderr = commands.execCmd(cmd, sudo=True) 
>       self.assertEqual(stdout[0].decode("ascii"), SUDO_USER) 
E       IndexError: list index out of range
storage/misc_test.py:600: IndexError
```

### Investigation

The problem seems to be related with the `-l` option of the sudo command in the `checkSudo` function: https://www.sudo.ws/docs/man/1.8.13/sudo.man/

```
If a command is specified and is permitted by the security policy, the fully-qualified path to the 
command is displayed along with any command line arguments. If command is specified but 
not allowed, sudo will exit with a status value of 1.
```

So doing:

```bash
$ rpm -q sudo   
sudo-1.9.7p2-2.fc35.x86_64

$ sudo -l -n whoami
/usr/bin/whoami
```

Just prints the full path, no error, and `checkSudo` fails to recognize the need for sudo, which ends up causing the execution of the command to fail.

### Proposed Solution

- For the additional checks of the return code, when rc is `!=0` the test should fail and print the error output for a more descriptive log.
- For the `checkSudo`, replace the function with the `@requires_root` decorator. Eventually might be good idea to remove `checkSudo` code completely.

nirs:
@aesteve-rh See comments on #144, they apply to this issue as well.

PR #145, opened 2022-04-21
misc_test: Check ret code and improve sudo check 

aesteve-rh:
Fixes: #143
Changes:
- Add additional checks to verify the return value on tests that execute commands in misc_tests
- Remove copyrighted texts from misc_tests
- Avoid using checkSudo in misc_tests, use `@requires_root` decorator instead

### DIFF
--- a/tests/storage/misc_test.py
+++ b/tests/storage/misc_test.py
@@ -47,7 +47,7 @@ from vdsm.storage import fileUtils
 from vdsm.storage import misc
 
 from monkeypatch import MonkeyPatch
-from testValidation import checkSudo
+from . marks import requires_root
 
 
 EXT_DD = "/bin/dd"
@@ -585,13 +585,13 @@ class TestExecCmd(VdsmTestCase):
         self.assertEqual(ret, 0, f"Command failed: {stderr}")
         self.assertEqual(stderr[0].decode("ascii"), "it works!")
 
+    @requires_root
     def testSudo(self):
         """
         Tests that when running with sudo the user really is root (or other
         desired user).
         """
         cmd = [EXT_WHOAMI]
-        checkSudo(cmd)
         ret, stdout, stderr = commands.execCmd(cmd, sudo=True)
         self.assertEqual(ret, 0, f"Command failed: {stderr}")
         self.assertEqual(stdout[0].decode("ascii"), SUDO_USER)
--- a/tests/storage/misc_test.py
+++ b/tests/storage/misc_test.py
@@ -415,9 +415,7 @@ class TestReadBlock(VdsmTestCase):
         """
         Test that when all arguments are correct the method works smoothly.
         """
-        writeData = (b"DON'T THINK OF IT AS DYING, said Death. "
-                     b"JUST THINK OF IT AS LEAVING EARLY TO AVOID THE RUSH.")
-        # (C) Terry Pratchet - Good Omens
+        writeData = b"Some data"
         dataLength = len(writeData)
 
         offset = 512
@@ -459,9 +457,7 @@ class TestReadBlock(VdsmTestCase):
         See that correct exception is raised when trying to read more then the
         file has to offer.
         """
-        writeData = (b"History, contrary to popular theories, "
-                     b"is kings and dates and battles.")
-        # (C) Terry Pratchet - Small Gods
+        writeData = b"Some data"
 
         offset = 512
         size = 512
@@ -574,9 +570,7 @@ class TestExecCmd(VdsmTestCase):
         Tests that execCmd correctly returns the standard output of the prog it
         executes.
         """
-        line = ("All I wanted was to have some pizza, hang out with dad, "
-                "and not let your weirdness mess up my day")
-        # (C) Nickolodeon - Invader Zim
+        line = "Some text line"
         ret, stdout, stderr = commands.execCmd((EXT_ECHO, line))
         self.assertEqual(ret, 0, f"Command failed: {stderr}")
         self.assertEqual(stdout[0].decode("ascii"), line)
--- a/tests/storage/misc_test.py
+++ b/tests/storage/misc_test.py
@@ -578,6 +578,7 @@ class TestExecCmd(VdsmTestCase):
                 "and not let your weirdness mess up my day")
         # (C) Nickolodeon - Invader Zim
         ret, stdout, stderr = commands.execCmd((EXT_ECHO, line))
+        self.assertEqual(ret, 0, f"Command failed: {stderr}")
         self.assertEqual(stdout[0].decode("ascii"), line)
 
     def testStdErr(self):
@@ -587,6 +588,7 @@ class TestExecCmd(VdsmTestCase):
         """
         cmd = ["sh", "-c", "echo it works! >&2"]
         ret, stdout, stderr = commands.execCmd(cmd)
+        self.assertEqual(ret, 0, f"Command failed: {stderr}")
         self.assertEqual(stderr[0].decode("ascii"), "it works!")
 
     def testSudo(self):
@@ -597,6 +599,7 @@ class TestExecCmd(VdsmTestCase):
         cmd = [EXT_WHOAMI]
         checkSudo(cmd)
         ret, stdout, stderr = commands.execCmd(cmd, sudo=True)
+        self.assertEqual(ret, 0, f"Command failed: {stderr}")
         self.assertEqual(stdout[0].decode("ascii"), SUDO_USER)
 
     def testNice(self):


tests/storage/outofprocess_test.py
```diff
@@ -49,6 +50,13 @@ def oop_cleanup():
     oop.stop()
 
 
+@pytest.fixture(scope='session')
+def usr_file_512():
+    userstorage.PATHS["file-512"].setup()
+    yield userstorage.PATHS["file-512"].path
+    userstorage.PATHS["file-512"].teardown()
```
aesteve-rh, line 57: Checking the READMEs, it states that teardown is rarely required. So if no teardown is required and setup can be assumed, I probably could get rid of the fixture altogether. I'll wait for some feedback though.

tests/storage/misc_test.py
```diff
@@ -576,8 +576,9 @@ def testStdOut(self):
         """
         line = ("All I wanted was to have some pizza, hang out with dad, "
                 "and not let your weirdness mess up my day")
-        # (C) Nickolodeon - Invader Zim
+        # (C) Nickelodeon - Invader Zim
```
nirs, line 579: This change is not related. The real fix is to remove the copyrighted text from the tests, instead of fixing the copyright. The tests license is GPL and they cannot contain text with another copyright.

tests/testValidation.py
```diff
@@ -354,7 +354,7 @@ def wrapper(*args, **kwargs):
 
 def checkSudo(cmd):
     try:
-        p = subprocess.Popen(['sudo', '-l', '-n'] + cmd,
+        p = subprocess.Popen(['sudo', '-n'] + cmd,
```
nirs, line 357: I'm not sure about this change, and in general tests using this check are wrong.

Failure to run command using "sudo -n" or "sudo -l -n" does not prove that sudo is not configured for the command. The command can fail because of a real failure running the command.

We have only few tests using this check:

```
$ git grep checkSudo
tests/mkimage_test.py:from testValidation import checkSudo, ValidateRunningAsRoot
tests/mkimage_test.py:        checkSudo(["mount", "-o", "loop", "somefile", "target"])
tests/mkimage_test.py:        checkSudo(["umount", "target"])
tests/storage/misc_test.py:from testValidation import checkSudo
tests/storage/misc_test.py:        checkSudo(cmd)
tests/testValidation.py:def checkSudo(cmd):
```

The way we use this check is wrong:

    checkSudo(["mount", "-o", "loop", "somefile", "target"])
    checkSudo(["umount", "target"])

These can be fail to mount, unrelated to sudo.

Even when the call works, it works because we run as root. Only certain commands are configured to work with sudo (see static/etc/sudoers.d/50_vdsm.in), and using sudo works only if you run the tests as vdsm, and the machine is configured for vdsm.

This is never true on a development machine or a container. So I think we should remove this code, and skip the sudo tests unless we run as user "vdsm", and can run "lvm" readonly command that will never fail for another reason, like:

$ lvm vgs; echo $?
  WARNING: Running as a non-root user. Functionality may be unavailable.
  /run/lock/lvm/P_global:aux: open failed: Permission denied
5

Or, simply remove checkSudo, and replace it with @requires_root. When we switch storage tests to run as vdsm, we can fix this issue in a better way.

tests/storage/outofprocess_test.py
nirs: I think you are confusing the upstream userstorage code with vdsm userstorage code.

This fixture repeat "file-512" 3 times, and cannot be changed to support other types of storage, like "file-4k". I'm not sure we need to test this code with 4k block size, but being able to do this later is important.

The way we use this in other files is:

```
@pytest.fixture(
    scope="module",
    params=[
        userstorage.PATHS["file-512"],
        userstorage.PATHS["file-4k"],
    ],   
    ids=str,
)
def user_storage(request):
    storage = request.param
    if not storage.exists():
        pytest.xfail("{} storage not available".format(storage.name))
    return storage
```

We should use the same fixture here, with one parameter "file-512". We an add more block sizes later if needed by adding one line to the fixture params.

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +975,55 @@ def test_glob(oop_cleanup, tmpdir):
...
-def test_read_lines(oop_cleanup, tmpdir):
+def test_read_lines(oop_cleanup, usr_file_512):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
 
-    with open(path, "wb") as f:
+    with open(usr_file_512, "wb") as f:
         f.write(b"1\n2\n3\n")
-    assert iop.readLines(path) == [b"1", b"2", b"3"]
+    assert iop.readLines(usr_file_512) == [b"1", b"2", b"3"]
```
nirs, line 983: Looks fine, and will be better when using a more generic name (user_file).

nirs (changes requested):
This PR mix fixes for 3 unrelated issues.
 
1. issue with checkSudo
2. issue with outofprocess tests on btrfs, 
3. refactoring copy_data 

Let's keep this PR deadling with checkSudo, and open 2 new PRs for the other unrelated issues.

This will be easier to review and we will be able to merge quickly, since the PRs are not related in anyway.

tests/storage/misc_test.py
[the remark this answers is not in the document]
aesteve-rh: True, but I couldn't resist once I noticed the typo. I'll remove the copyright :)

tests/testValidation.py
aesteve-rh: I'll go with the decorator for now, then. Thanks for the explanation and all the context for that check!

aesteve-rh: Tracked in https://github.com/oVirt/vdsm/issues/152

tests/storage/misc_test.py
```diff
@@ -576,8 +576,9 @@ def testStdOut(self):
...
         ret, stdout, stderr = commands.execCmd((EXT_ECHO, line))
+        self.assertTrue(ret == 0, f'rc: {ret} - stderr: "{stderr}"')
```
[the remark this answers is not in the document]
aesteve-rh: Done.

tests/testValidation.py
aesteve-rh: Done. Reverted so that the file is unchanged and tracked in #152 as stated above so that we can change it properly in a different PR.

tests/storage/outofprocess_test.py
aesteve-rh: Addressed in another PR after splitting commits.

tests/storage/outofprocess_test.py
aesteve-rh: Addressed in another PR after splitting commits (ended up using user_mount).

nirs (changes requested):

tests/storage/misc_test.py
```diff
@@ -587,6 +588,8 @@ def testStdErr(self):
         """
         cmd = ["sh", "-c", "echo it works! >&2"]
         ret, stdout, stderr = commands.execCmd(cmd)
+        self.assertEqual(ret, 0, f"Command failed: {stderr}")
+        self.assertEqual(stdout, [], f"Unexpected output: {stdout}")
```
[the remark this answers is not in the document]
aesteve-rh: Yes, I'll remove the check.

tests/storage/misc_test.py
```diff
@@ -590,13 +590,13 @@ def testStdErr(self):
...
         self.assertEqual(stderr[0].decode("ascii"), "it works!")
 
+    @requires_root
     def testSudo(self):
         """
         Tests that when running with sudo the user really is root (or other
         desired user).
         """
         cmd = [EXT_WHOAMI]
-        checkSudo(cmd)
```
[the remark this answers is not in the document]
aesteve-rh: I think this is addressed already in the current state?

tests/storage/misc_test.py
```diff
@@ -415,9 +415,9 @@ def testValidInput(self):
         """
         Test that when all arguments are correct the method works smoothly.
         """
-        writeData = (b"DON'T THINK OF IT AS DYING, said Death. "
-                     b"JUST THINK OF IT AS LEAVING EARLY TO AVOID THE RUSH.")
-        # (C) Terry Pratchet - Good Omens
+        writeData = (b"Lorem ipsum dolor sit amet, consectetur adipiscing "
+                     b"elit. Curabitur risus turpis, gravida vel ligula in, "
+                     b"rhoncus maximus sem. Curabitur ac.")
```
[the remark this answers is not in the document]
nirs: Some data, why do we need random data?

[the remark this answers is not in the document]
aesteve-rh: I was just trying to figure out what did you mean by "some data" where "Lorem ipsum" is not good enough. I understand that it does not have to be a text-like input, but I still need to write something and assign it to `writeData`. So, I could smash my keyboard, generate data with a (deterministic) data generator in the test itself, or generate/look for random data and then paste it in the code.

Or maybe you are happy enough by just doing:
```python
        writeData = b"Some data"
```
I am not sure, and that's what I was trying to ask.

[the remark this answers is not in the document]
nirs: Yes, this looks good.

Check if the test is expecting certain size - it is not clear what this test is trying to do, and the copyrighted text did not help to understand what the tests is checking.

If the test is expecting data with certain size, the best way to create it is:

    b"x" * expected_size

If the test just need some data with any size:


    b"some data"

is makes the tests very simple to understand and debug in case the test fails.

If the test is about blocks, it can be useful to create blocks of text:

    b"A" * 512 + b"B" * 512

Finally, it is useful to see how vdsm is using this code, and test in the same
way. For example when we write metadata, we create a block of 512 bytes like this:

    b"data".

aesteve-rh:
@nirs still doesn't let me re-request a review in this PR for some reason. Anyway, should be ready now to give it another look.

nirs (approved):

Issue #144, opened 2022-04-21
Improve test_write_file_direct_true_aligned 


aesteve-rh:
### Description

`outofprocess_test` uses temporary files created by pytest fixture. However, when using these files, `test_write_file_direct_true_aligned` will not raise with file systems without minimum block size for direct I/O (e.g., btrfs).

### Traceback

```bash
$ tox -e storage -- -vv storage/outofprocess_test.py::test_write_file_direct_true_unaligned
...
oop_cleanup = None, tmpdir = local('/var/tmp/vdsm/test_write_file_direct_true_un0')   
    def test_write_file_direct_true_unaligned(oop_cleanup, tmpdir): 
        iop = oop.getProcessPool("test")
        path = str(tmpdir.join("file"))
>       with pytest.raises(OSError) as e:
E       Failed: DID NOT RAISE <class 'OSError'> 

storage/outofprocess_test.py:1010: Failed  
```
while:
```bash
$ df -hT /var/tmp
Filesystem                                            Type   Size  Used Avail Use% Mounted on
/dev/mapper/luks-08653fbb-0d5f-4a76-be60-30be4377cf5d btrfs  476G  119G  357G  25% /
```
### Proposed solution

To avoid such issues, the tests should use user storage instead of random storage from the host running the tests.

nirs:
@aesteve-rh can you change this issue to describe the problem?

- Describe the issue
- Include the traceback from the test
- Provide additional data if you have, explaining the issue or hinting on the solution

The issue should have enough data so someone else can pick it up easily.

PR #148, opened 2022-04-22
outofprocess_test: 'External API' block to use userstorage

aesteve-rh:
Fixes: #144 

### DIFF
--- a/tests/storage/outofprocess_test.py
+++ b/tests/storage/outofprocess_test.py
@@ -26,6 +26,7 @@ import gc
 import logging
 import os
 import re
+import shutil
 import stat
 import time
 import weakref
@@ -39,6 +40,7 @@ from vdsm.storage import constants as sc
 from vdsm.storage import outOfProcess as oop
 from vdsm.storage.exception import MiscDirCleanupFailure
 
+from . import userstorage
 from . marks import requires_root, requires_unprivileged_user
 from . storagetestlib import chmod
 
@@ -47,6 +49,23 @@ from . storagetestlib import chmod
 def oop_cleanup():
     yield
     oop.stop()
+
+
+@pytest.fixture(
+    params=[
+        userstorage.PATHS["mount-512"],
+        userstorage.PATHS["mount-4k"],
+    ],
+    ids=str,
+)
+def user_mount(request):
+    storage = request.param
+    if not storage.exists():
+        pytest.xfail("{} storage not available".format(storage.name))
+    tmpdir = os.path.join(storage.path, "tmp")
+    os.mkdir(tmpdir)
+    yield tmpdir
+    shutil.rmtree(tmpdir)
 
 
 # TODO: the following 2 tests use private instance variables that
@@ -967,54 +986,54 @@ def test_glob(oop_cleanup, tmpdir):
 
 # External APIs
 
-def test_read_lines(oop_cleanup, tmpdir):
+def test_read_lines(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
+    path = os.path.join(user_mount, "file")
 
     with open(path, "wb") as f:
         f.write(b"1\n2\n3\n")
     assert iop.readLines(path) == [b"1", b"2", b"3"]
 
 
-def test_write_lines(oop_cleanup, tmpdir):
+def test_write_lines(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
+    path = os.path.join(user_mount, "file")
 
     iop.writeLines(path, [b"1\n", b"2\n", b"3\n"])
     with open(path, "rb") as f:
         assert f.read() == b"1\n2\n3\n"
 
 
-def test_write_file_direct_false(oop_cleanup, tmpdir):
+def test_write_file_direct_false(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
+    path = os.path.join(user_mount, "file")
 
     iop.writeFile(path, b"1\n2\n3\n", direct=False)
     with open(path, "rb") as f:
         assert f.read() == b"1\n2\n3\n"
 
 
-def test_write_file_direct_true_aligned(oop_cleanup, tmpdir):
+def test_write_file_direct_true_aligned(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
+    path = os.path.join(user_mount, "file")
 
     iop.writeFile(path, b"x" * 4096, direct=True)
     with open(path, "rb") as f:
         assert f.read() == b"x" * 4096
 
 
-def test_write_file_direct_true_unaligned(oop_cleanup, tmpdir):
+def test_write_file_direct_true_unaligned(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
+    path = os.path.join(user_mount, "file")
 
     with pytest.raises(OSError) as e:
         iop.writeFile(path, b"1\n2\n3\n", direct=True)
     assert e.value.errno == errno.EINVAL
 
 
-def test_truncate_file_default_mode(oop_cleanup, tmpdir):
+def test_truncate_file_default_mode(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
+    path = os.path.join(user_mount, "file")
 
     iop.truncateFile(path, 10)
 


tests/storage/outofprocess_test.py
```diff
@@ -49,6 +50,20 @@ def oop_cleanup():
     oop.stop()
 
 
+@pytest.fixture(
+    scope="module",
+    params=[
```
vjuranek, line 55: Is there any need to parametrize it? The tests don't seem to use it.

aesteve-rh: It is used below:

```python
def user_file(request):
    storage = request.param
```

Having it parametrized makes it simple to update/modify in the future if we need to change/add any user file used here. Also, it is easier to identify the change with git diff. And finally, it keeps the fixture consistent with other similar fixtures for `userstorage` in other tests in this folder.

It could be a good idea, though, to keep common fixtures in a global library file to avoid code repetition.

vjuranek: yes, but I guess you can use 
```
storage = userstorage.PATHS["file-512"]
```
Other fixtures are parametrize because they run the test for more parameter values (typically for both 512 and 4k), but this not the case here. Let's wait for @nirs what he thinks about it, but personally I'd not make code more complicated than it needs to be and add parameters once it's really needed.

aesteve-rh: I agree, but it still make it nicer to extend in the future if we keep it parametrized. And the project already have tests with only one parameter. E.g.:

https://github.com/oVirt/vdsm/blob/feb336ae643e0342df74877819a3dec10ed10bb4/tests/storage/sdm_copy_data_test.py#L72-L83

tests/storage/outofprocess_test.py
```diff
@@ -41,6 +41,7 @@
```
nirs, line 41: The commit message needs a body explaining this change.

The "Fixes: #NNN" header should be in the end of the commit mesage, before the Signed-off-by: header.

tests/storage/outofprocess_test.py
nirs: This code should support 4k storage, so we are very likely to add also "file-4k" parameter. Even if we are not sure using parameters make the code easy to extend.

tests/storage/outofprocess_test.py
```diff
@@ -49,6 +50,20 @@ def oop_cleanup():
...
+    params=[
+        userstorage.PATHS["file-512"],
+    ],
+    ids=str,
+)
+def user_file(request):
+    storage = request.param
+    if not storage.exists():
+        pytest.xfail("{} storage not available".format(storage.name))
+    return storage
```
nirs, line 64: We need to reset this file before each test so tests cannot be affected by previous tests. In other modules using this storage we truncate the file before each test.

Example:
https://github.com/oVirt/vdsm/blob/feb336ae643e0342df74877819a3dec10ed10bb4/tests/storage/filesd_metadata_test.py#L114

But looking at the tests, we really want a temporary directory and not a file, so we should use "mount-512" instead.

The current tests use a temporary directory, so the easier and smallest change is to create a fixture returning a new temporary directory, and delete the temporary directory no the end of the test.

Example: https://github.com/oVirt/vdsm/blob/feb336ae643e0342df74877819a3dec10ed10bb4/tests/storage/localfssd_test.py#L78

Since the fixture return a path instead of pytedt tmpdir, the tests will have to change code like:

    path = str(tmpdir.join("file"))

To:

    path = os.path.join(user_tmpdir, "file")

But no other change is needed in the tests.

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +982,56 @@ def test_glob(oop_cleanup, tmpdir):
...
-def test_write_lines(oop_cleanup, tmpdir):
+def test_write_lines(oop_cleanup, user_file):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
 
-    iop.writeLines(path, [b"1\n", b"2\n", b"3\n"])
-    with open(path, "rb") as f:
+    iop.writeLines(user_file.path, [b"1\n", b"2\n", b"3\n"])
+    with open(user_file.path, "rb") as f:
         assert f.read() == b"1\n2\n3\n"
```
nirs, line 998: Previously we tested that io.writeLines() creates the file if the file does not exists. Now we use an existing file, so we don't test the behavior when file does not exist.

This test needs a directory, not a file, so we better use the "mount-512" and create a temporary directory inside this mount. See localfssd_test.py for example of using "mount-512" in a safe way.

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +982,56 @@ def test_glob(oop_cleanup, tmpdir):
...
-def test_write_file_direct_false(oop_cleanup, tmpdir):
+def test_write_file_direct_false(oop_cleanup, user_file):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
 
-    iop.writeFile(path, b"1\n2\n3\n", direct=False)
-    with open(path, "rb") as f:
+    iop.writeFile(user_file.path, b"1\n2\n3\n", direct=False)
+    with open(user_file.path, "rb") as f:
         assert f.read() == b"1\n2\n3\n"
```
nirs, line 1006: Same issue, testing now existing file, before the call created the file.

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +982,56 @@ def test_glob(oop_cleanup, tmpdir):
...
-def test_write_file_direct_true_aligned(oop_cleanup, tmpdir):
+def test_write_file_direct_true_aligned(oop_cleanup, user_file):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
 
-    iop.writeFile(path, b"x" * 4096, direct=True)
-    with open(path, "rb") as f:
+    iop.writeFile(user_file.path, b"x" * 4096, direct=True)
+    with open(user_file.path, "rb") as f:
         assert f.read() == b"x" * 4096
```
nirs, line 1014: Same

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +982,56 @@ def test_glob(oop_cleanup, tmpdir):
...
 
-def test_write_file_direct_true_unaligned(oop_cleanup, tmpdir):
+def test_write_file_direct_true_unaligned(oop_cleanup, user_file):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
 
     with pytest.raises(OSError) as e:
-        iop.writeFile(path, b"1\n2\n3\n", direct=True)
+        iop.writeFile(user_file.path, b"1\n2\n3\n", direct=True)
     assert e.value.errno == errno.EINVAL
```
nirs, line 1022: Same

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +982,56 @@ def test_glob(oop_cleanup, tmpdir):
...
 
 
-def test_truncate_file_default_mode(oop_cleanup, tmpdir):
+def test_truncate_file_default_mode(oop_cleanup, user_file):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
 
-    iop.truncateFile(path, 10)
+    iop.truncateFile(user_file.path, 10)
```
nirs, line 1028: Same, previously we tested that iop.truncateFile() create the file.

nirs (changes requested):

tests/storage/outofprocess_test.py
aesteve-rh: OK.

I have used [pathlib](https://docs.python.org/3/library/pathlib.html) instead of `os.path`, since it is much nicer (in this case is not such a big difference, and one could argue that `os` was already imported in the module, but start introducing a library little by little is an approach, and `pathlib` is there since v3.4).

Also, I changed the fixture a bit. I checked `localfssd_test` as an example, but changed it differently. I have changed the scope to function, so that it is executed for every test:
- It creates the file Path from inside the fixture, to avoid code repetition
- Then yields it instead of the raw mounting point
- Finally, it removes the file so that each test has to create it anew every time

But I can just change it to follow the example as in `localfssd_test`, I was experimenting a bit :)

tests/storage/outofprocess_test.py
aesteve-rh: Done, see comment above. I will close all these threads and we can follow the discussion in https://github.com/oVirt/vdsm/pull/148#discussion_r857135030 .

tests/storage/outofprocess_test.py
aesteve-rh: Already added the support for 4k mounts. I'll resolve this conversation.

tests/storage/outofprocess_test.py
```diff
@@ -31,6 +31,7 @@
 import weakref
 
 from contextlib import contextmanager
+from pathlib import Path
```
nirs, line 34: We don't use pathlib in vdsm for 2 reasons:
- It did not exist in python 2.7, and and vdsm was ported to python 3 only in 2020
- I don't see any advantage in using it, and syntax like:

      path / "file"

  for joining "file" to a path is actually worse. "/" should be used only for 
  division.

tests/storage/outofprocess_test.py
```diff
@@ -31,6 +31,7 @@
 import weakref
```
nirs, line 31: Regarding the commit message - it mostly describes how the code works, but we can see this in the code.

What we cannot see in the code is:
- What is the issue with pytest tmpdir?
- What is the advantage of using userstorage?

The commit message should answer these questions.

The commit message can add some details about the implementation, if the details
are important to the change. I'm not sure the details about how the file is created are important.

The blank line between Fixes: and Signed-off-by: should be removed. We should treat "Fixes:" as a header to make it easier for process git logs with tools.

tests/storage/outofprocess_test.py
```diff
@@ -41,6 +42,7 @@
 
 from . marks import requires_root, requires_unprivileged_user
 from . storagetestlib import chmod
+from . import userstorage
```
nirs, line 45: Please sort the "from ." imports. Less specific imports:

    from . import foo

should be sorted before more specific imports like:

    form . bar import baz

Check how we import in other modules.

tests/storage/outofprocess_test.py
```diff
@@ -49,6 +51,23 @@ def oop_cleanup():
     oop.stop()
 
 
+@pytest.fixture(
+    scope="function",
```
nirs, line 55: function is the default, why do we need to specify it? No other fixture use scope="function". If we don't have a good reason to change all fixture, we better be consistent with existing code.

tests/storage/outofprocess_test.py
```diff
@@ -49,6 +51,23 @@ def oop_cleanup():
...
+    params=[
+        userstorage.PATHS["mount-512"],
+        userstorage.PATHS["mount-4k"],
```
nirs, line 58: Testing 2 block sizes is important change that should be described in the commit message.

Add 4k block size can also be done later if it make this change harder. If everything works, we can keep the change here.

tests/storage/outofprocess_test.py
```diff
@@ -49,6 +51,23 @@ def oop_cleanup():
...
+    ],
+    ids=str,
+)
+def user_mount(request):
```
nirs, line 62: This is a mount...

tests/storage/outofprocess_test.py
```diff
@@ -49,6 +51,23 @@ def oop_cleanup():
...
+    storage = request.param
+    if not storage.exists():
+        pytest.xfail("{} storage not available".format(storage.name))
+    file = Path(storage.path) / "file"
+    yield file
```
nirs, line 67: So why it returns a filename?

If you want to return a file, call it user_file, same way it is used in other fixtures. But I don't think we should return a file here, since creating a file name
from the directory is trivial, and getting a directory is more flexiable. If we add a new tests that need to create 2 files, this fixture will be too specific and we will have to make it more generic and return the directory. So lets avoid this trouble by returning a directory like other user_mount fixture do.

tests/storage/outofprocess_test.py
```diff
@@ -49,6 +51,23 @@ def oop_cleanup():
...
+    file.unlink()
+
```
nirs, line 69: I think what we should have here is:

    tmpdir = os.mkdir(storage.path, "tmp")
    yield tmpdir
    shutil.rmtree(tmpdir)

The test will use the directory in the same way they use it today.

tests/storage/outofprocess_test.py
```diff
@@ -967,61 +986,56 @@ def test_glob(oop_cleanup, tmpdir):
 
 # External APIs
 
-def test_read_lines(oop_cleanup, tmpdir):
+def test_read_lines(oop_cleanup, user_mount):
     iop = oop.getProcessPool("test")
-    path = str(tmpdir.join("file"))
```
nirs, line 972: This is the only change need in this test:

    path = os.path.join(user_mount, "file"))

The rest of the does not need any change, and the new version is actually worse, since we need to do str(user_ount) each time we use the patch, instead of once.
If ioprocess interface was accepting pathlib.Path, this could be better, but I don't think it worth our time to change the interface to support pathlib. This bring zero value to oVirt users and questionable value to developers.

nirs (changes requested):

tests/storage/outofprocess_test.py
aesteve-rh: Well, the '/' syntax is optional, it could as well be parameters of the Path constructor: `Path(path, "file)`.

In general as I wrote before, for joining there is not much difference between `os.path` and `pathlib`, but there are cases where working with `os.path` relies on nested calls and string manipulation, whereas pathlib deals with it much more elegantly and readably.

I know why it wasn't used before, but introducing a new library when we do new code or when code is revisited is nice, overtime we will see libraries removed/modernized without having to do a massive refactor.

Anyway, I will not push for the change, I'll change it back to `os.path` for now. We can re-discuss this in the future if the need rises.

tests/storage/outofprocess_test.py
aesteve-rh: OK, I see your point to make it more extendable.

I returned a filename mostly because it was what all tests were doing, and I tried to avoid code repetition, having to do that every time in every test. But right, too specific behavior.

tests/storage/outofprocess_test.py
aesteve-rh: I agree that having to cast to string every Path is not so nice, but as you pointed out, my thought on this was that overtime `ioprocess` could accept Path. Value for developers could be to have OOP path handling and more readable code, and well, flexibility. But this I guess is a matter of taste mostly, and `os.path` will not be deprecated any time soon, if ever. So I will change the code back to it.

tests/storage/outofprocess_test.py
nirs: Introducing new library in a big code base creates yet another way to do the same thing. This makes the code harder to work with. It can make sense if the new way is much better and solve actual issues.

Good example is using fstring in error messages - most of the code uses % which is error prone. In newer code we moved to str.format() which is less fragile. Since we moved to python 3, we use fstrings in error messages, which much less error prone, more clear and less verbose.

nirs (approved):

PR #159, opened 2022-05-02
lvm: remove outdated warning

aesteve-rh:
This warning is left over from times we retried the lvm command
before failing. Now we run once with specific lvm devices, and again
with all devices, so this log is not very helpful and can be removed.

Addresses the first issue in https://bugzilla.redhat.com/show_bug.cgi?id=2048545

### DIFF
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -451,8 +451,6 @@ class LVMCache(object):
 
     def run_command(self, cmd, devices=(), use_lvmpolld=True):
         with self._cmd_sem:
-            tries = 1
-
             # 1. Try the command with fast specific filter including the
             # specified devices. If the command succeeded and wanted output was
             # returned we are done.
@@ -472,13 +470,10 @@ class LVMCache(object):
                     "Command with specific filter failed or returned no data, "
                     "retrying with refreshed device list: %s", error)
                 full_cmd = wider_cmd
-                tries += 1
                 try:
                     return self._runner.run(full_cmd)
                 except se.LVMCommandError as e:
                     error = e
-
-            log.warning("All %d tries have failed: %s", tries, error)
 
             raise error
 


lib/vdsm/storage/lvm.py
```diff
@@ -473,14 +471,11 @@ def run_command(self, cmd, devices=(), use_lvmpolld=True):
                     "Command with specific filter failed or returned no data, "
                     "retrying with refreshed device list: %s", error)
                 full_cmd = wider_cmd
-                tries += 1
                 try:
                     return self._runner.run(full_cmd)
                 except se.LVMCommandError as e:
                     error = e
 
-            log.warning("All %d tries have failed: %s", tries, error)
```
nirs, line 482: Removal looks good - nice that you also got rid of the ties counter :+1: 

But we need to make sure that this warning was not the only place we log lvm errors on failure during reloadlvs, reloadpvs and reloadvgs calls.

It would be useful to introduce fake errors in all reload methods by adding an invalid argument to the command, for example:

    if random.random() > 0.5:
        log.info("Faking lvm error")
        cmd.append("--bad-argument")

And run the system for a while. Then check the logs to see all fake errors were log properly, meaning that real lvm error will be logged.

aesteve-rh: I have tried that, and I couldn't see any errors being logged. So I have been digging a bit into it, using the `lvm_test` for it.

In reload methods, `lvm` command is run through `run_command_error`, which captures the `LVMCommandError` exception and returns it. Then, back in the reload method, the error is only checked for a potential reload failure, and a warning is logged, in any other case, `lvm` errors are just ignored.

So there is no other place where `lvm` errors are logged that I can see so far.
I'll see how's the best way to improve error handling in the caller.

nirs: The most relevant code path is _reloadvgs():

        with self._lock:
            if error:
                unreadable_vgs = [] 
                for v in (vgNames or self._vgs):
                    vg = self._vgs.get(v)
                    if vg and vg.is_stale():
                        self._vgs[v] = Unreadable(vg.name)
                        unreadable_vgs.append(vg.name)

                if unreadable_vgs:
                    # This may be a real error (failure to reload existing VG)
                    # or no error at all (failure to reload non-existing VG),
                    # so we cannot make this an error.
                    log.warning(
                        "Marked vgs=%r as Unreadable due to reload failure",
                        logutils.Head(unreadable_vgs, max_items=20))

If we try to reload a non existing vg that is not the cache,
we will not get any error about that once we remove the warning.

This is good for the case when we check if the vg exist, but not for
the case when the vg should exist. In this case we will raise 
se.VolumeGroupDoesNotExist in lvm.getVG(), but this error does not include
any context from the actual error in _reloadvgs().

This is not a new issue but removing the warning removes the only log we had
for this case, and it may make it harder to debug real issue of missing vg.

I think this can improved by changing LVMCache.getVg() to raise the error instead of
raising unhelpful error in lvm.getVG().

    def getVg(self, vgName):
        # Get specific VG
        vg = self._vgs.get(vgName)
        if not vg or vg.is_stale():
            self.stats.miss()
            vgs = self._reloadvgs(vgName)

Since we always call getVg() with one vg name, _reloadvgs() can raise in this case if
the command failed, and there is no output. If we raise se.VolumeGroupDoesNotExist 
in _reloadvgs(), the user will be able to log a helpful error in the case when non
existing vg is a real error, or ignore the exception in the case we check if a vg
exists.

_reloadvgs() is more complicated to handle, since when calling without vg name, the command does fail if some vgs cannot be read, but the command returns an output about the vgs that it can read. I think that when calling with one vg name we never get output
if vgs fail, so we can handle this case in a better way.

The current patch looks good, but we should fix error handling before we merge it.

aesteve-rh: I am working on this in a separate branch, whether I merge them together afterwards or put them in different PRs, we can decide later.

aesteve-rh: Follow up in a different PR: https://github.com/oVirt/vdsm/pull/161

nirs (approved):
Depends on #192

PR #161, opened 2022-05-05
lvm: reload VG error handling

aesteve-rh:
Currently, if we try to reload a non-existing VG that is not in the cache, we will not get any error, it will raise an exception without any context of the actual reason of the fail.

Furthermore, if the LVM command fails, it might go unnoticed, or trigger a single uninformative warning without the reason why it might have failed.

Make _reloadvgs to raise an exception if the LVM command failed and there is no output. Then, the caller will decide whether it is a real error or it was just probing.

Add tests to verify the behavior of all parts of `lvm` module that are affected by the error handling update.

Related https://github.com/oVirt/vdsm/pull/159

### DIFF
--- a/lib/vdsm/storage/exception.py
+++ b/lib/vdsm/storage/exception.py
@@ -1349,6 +1349,26 @@ class VolumeGroupDoesNotExist(StorageException):
     code = 506
     msg = "Volume Group does not exist"
 
+    def __init__(self, vg_name=None, vg_uuid=None, error=None):
+        if vg_name is None and vg_uuid is None:
+            raise ValueError("Require a VG name or UUID")
+
+        if error is not None and not isinstance(error, LVMCommandError):
+            raise TypeError(
+                f"Expecting instance of LVMCommandError, got {type(error)}")
+
+        self.vg_name = vg_name
+        self.vg_uuid = vg_uuid
+        self.error = error
+
+    @property
+    def value(self):
+        return ", ".join(f"{k}={v}" for k, v in self.__dict__.items() if v)
+
+    @classmethod
+    def from_error(cls, vg_name, error):
+        return cls(vg_name=vg_name, error=error).with_exception(error)
+
 
 class VolumeGroupRenameError(StorageException):
     code = 507
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -1487,19 +1487,13 @@ def _checkpvsblksize(pvs, vgBlkSize=None):
 
 
 def checkVGBlockSizes(vg_name, vgBlkSize=None):
-    pvs = listPVNames(vg_name)
-    if not pvs:
-        raise se.VolumeGroupDoesNotExist(vg_name=vg_name)
-    _checkpvsblksize(pvs, vgBlkSize)
+    _checkpvsblksize(listPVNames(vg_name), vgBlkSize)
 
 
 def getVGBlockSizes(vg_name):
-    pvs = listPVNames(vg_name)
-    if not pvs:
-        raise se.VolumeGroupDoesNotExist(vg_name=vg_name)
     # Returning the block size of the first pv is correct since we don't allow
     # devices with different block size to be on the same VG.
-    return _getpvblksize(pvs[0])
+    return _getpvblksize(listPVNames(vg_name)[0])
 
 #
 # Public Logical volume interface
@@ -1844,11 +1838,7 @@ def getVgMetadataPv(vgName):
 
 
 def listPVNames(vgName):
-    try:
-        pvNames = _lvminfo._vgs[vgName].pv_name
-    except (KeyError, AttributeError):
-        pvNames = getVG(vgName).pv_name
-    return pvNames
+    return getVG(vgName).pv_name
 
 
 def setrwLV(vg_name, lv_name, rw=True):
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -1486,17 +1486,17 @@ def _checkpvsblksize(pvs, vgBlkSize=None):
             raise se.VolumeGroupBlockSizeError(vgBlkSize, pvBlkSize)
 
 
-def checkVGBlockSizes(vgUUID, vgBlkSize=None):
-    pvs = listPVNames(vgUUID)
+def checkVGBlockSizes(vg_name, vgBlkSize=None):
+    pvs = listPVNames(vg_name)
     if not pvs:
-        raise se.VolumeGroupDoesNotExist(vg_uuid=vgUUID)
+        raise se.VolumeGroupDoesNotExist(vg_name=vg_name)
     _checkpvsblksize(pvs, vgBlkSize)
 
 
-def getVGBlockSizes(vgUUID):
-    pvs = listPVNames(vgUUID)
+def getVGBlockSizes(vg_name):
+    pvs = listPVNames(vg_name)
     if not pvs:
-        raise se.VolumeGroupDoesNotExist(vg_uuid=vgUUID)
+        raise se.VolumeGroupDoesNotExist(vg_name=vg_name)
     # Returning the block size of the first pv is correct since we don't allow
     # devices with different block size to be on the same VG.
     return _getpvblksize(pvs[0])
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -649,12 +649,29 @@ class LVMCache(object):
                 "Marked vgs=%r as Unreadable due to reload failure",
                 logutils.Head(unreadable_vgs, max_items=20))
 
-    def _reloadvgs(self, vgName=None):
+    def _reload_single_vg(self, vg_name):
+        """
+        Run LVM 'vgs' command and update VG name.
+        """
         cmd = list(VGS_CMD)
+        cmd.append(vg_name)
+        out, error = self.run_command_error(
+            cmd, devices=self._getVGDevs([vg_name]))
 
+        with self._lock:
+            if error:
+                self._update_stale_vgs_locked([vg_name])
+
+            return self._updatevgs_locked(out, [vg_name])
+
+    def _reloadvgs(self, vgName=None):
+        """
+        Run LVM 'vgs' command and update VG names.
+        If no VG name is provided, reload all VGs.
+        """
         vgNames = normalize_args(vgName)
-        if vgNames:
-            cmd.extend(vgNames)
+        cmd = list(VGS_CMD)
+        cmd.extend(vgNames)
 
         out, error = self.run_command_error(
             cmd, devices=self._getVGDevs(vgNames))
@@ -896,7 +913,7 @@ class LVMCache(object):
         vg = self._vgs.get(vgName)
         if not vg or vg.is_stale():
             self.stats.miss()
-            vgs = self._reloadvgs(vgName)
+            vgs = self._reload_single_vg(vgName)
             vg = vgs.get(vgName)
         else:
             self.stats.hit()
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -1275,7 +1275,7 @@ def movePV(vgName, src_device, dst_devices):
 def getVG(vgName):
     vg = _lvminfo.getVg(vgName)  # returns single VG namedtuple
     if not vg:
-        raise se.VolumeGroupDoesNotExist(vgName)
+        raise se.VolumeGroupDoesNotExist(vg_name=vgName)
     else:
         return vg
 
@@ -1301,7 +1301,7 @@ def getVGbyUUID(vgUUID):
             log.debug("%s", e, exc_info=True)
             continue
     # If not cry loudly
-    raise se.VolumeGroupDoesNotExist("vg_uuid: %s" % vgUUID)
+    raise se.VolumeGroupDoesNotExist(vg_uuid=vgUUID)
 
 
 def getLV(vgName, lvName=None):
@@ -1489,14 +1489,14 @@ def _checkpvsblksize(pvs, vgBlkSize=None):
 def checkVGBlockSizes(vgUUID, vgBlkSize=None):
     pvs = listPVNames(vgUUID)
     if not pvs:
-        raise se.VolumeGroupDoesNotExist("vg_uuid: %s" % vgUUID)
+        raise se.VolumeGroupDoesNotExist(vg_uuid=vgUUID)
     _checkpvsblksize(pvs, vgBlkSize)
 
 
 def getVGBlockSizes(vgUUID):
     pvs = listPVNames(vgUUID)
     if not pvs:
-        raise se.VolumeGroupDoesNotExist("vg_uuid: %s" % vgUUID)
+        raise se.VolumeGroupDoesNotExist(vg_uuid=vgUUID)
     # Returning the block size of the first pv is correct since we don't allow
     # devices with different block size to be on the same VG.
     return _getpvblksize(pvs[0])
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -577,6 +577,78 @@ class LVMCache(object):
                 devices = tuple(devices)
         return devices
 
+    def _updatevgs_locked(self, vgs_output, vg_names):
+        """
+        Update cached VGs based on the output of the LVM command:
+        - Add new VGs to the cache.
+        - Replace VGs in the cache with VGs reported by the 'vgs' command,
+          updating the VG attributes.
+        - If called without vg names, remove all VGs from the cache not
+          reported by LVM.
+        - If called with vg names, remove specifed VGs from the cache if they
+          were not reported by LVM.
+        Must be called while holding the lock.
+        Return dict of updated VGs.
+        """
+        updatedVGs = {}
+        vgsFields = {}
+        for line in vgs_output:
+            fields = [field.strip() for field in line.split(SEPARATOR)]
+            if len(fields) != VG_FIELDS_LEN:
+                raise InvalidOutputLine("vgs", line)
+
+            uuid = fields[VG._fields.index("uuid")]
+            pvNameIdx = VG._fields.index("pv_name")
+            pv_name = fields[pvNameIdx]
+            if pv_name == UNKNOWN:
+                # PV is missing, e.g. device lost of target not connected
+                continue
+            if uuid not in vgsFields:
+                fields[pvNameIdx] = [pv_name]  # Make a pv_names list
+                vgsFields[uuid] = fields
+            else:
+                vgsFields[uuid][pvNameIdx].append(pv_name)
+        for fields in vgsFields.values():
+            vg = VG.fromlvm(*fields)
+            if int(vg.pv_count) != len(vg.pv_name):
+                log.error("vg %s has pv_count %s but pv_names %s",
+                          vg.name, vg.pv_count, vg.pv_name)
+            self._vgs[vg.name] = vg
+            updatedVGs[vg.name] = vg
+
+        # Remove stale VGs
+        staleVGs = [name for name in (vg_names or self._vgs)
+                    if name not in updatedVGs]
+        for name in staleVGs:
+            if name in self._vgs:
+                log.warning("Removing stale VG %s", name)
+                del self._vgs[name]
+                # Remove fresh lvs indication of the vg removed from cache.
+                self._freshlv.discard(name)
+
+        return updatedVGs
+
+    def _update_stale_vgs_locked(self, vg_names):
+        """
+        Check if any VG is stale after an LVM command and make it Unreadable.
+        Log a warning if any VG has been made Unreadable.
+        Must be called while holding the lock.
+        """
+        unreadable_vgs = []
+        for v in (vg_names or self._vgs):
+            vg = self._vgs.get(v)
+            if vg and vg.is_stale():
+                self._vgs[v] = Unreadable(vg.name)
+                unreadable_vgs.append(vg.name)
+
+        if unreadable_vgs:
+            # This may be a real error (failure to reload existing VG)
+            # or no error at all (failure to reload non-existing VG),
+            # so we cannot make this an error.
+            log.warning(
+                "Marked vgs=%r as Unreadable due to reload failure",
+                logutils.Head(unreadable_vgs, max_items=20))
+
     def _reloadvgs(self, vgName=None):
         cmd = list(VGS_CMD)
 
@@ -589,59 +661,12 @@ class LVMCache(object):
 
         with self._lock:
             if error:
-                unreadable_vgs = []
-                for v in (vgNames or self._vgs):
-                    vg = self._vgs.get(v)
-                    if vg and vg.is_stale():
-                        self._vgs[v] = Unreadable(vg.name)
-                        unreadable_vgs.append(vg.name)
-
-                if unreadable_vgs:
-                    # This may be a real error (failure to reload existing VG)
-                    # or no error at all (failure to reload non-existing VG),
-                    # so we cannot make this an error.
-                    log.warning(
-                        "Marked vgs=%r as Unreadable due to reload failure",
-                        logutils.Head(unreadable_vgs, max_items=20))
+                self._update_stale_vgs_locked(vgNames)
 
                 # NOTE: vgs may return useful output even on failure, so we
                 # don't return here.
 
-            updatedVGs = {}
-            vgsFields = {}
-            for line in out:
-                fields = [field.strip() for field in line.split(SEPARATOR)]
-                if len(fields) != VG_FIELDS_LEN:
-                    raise InvalidOutputLine("vgs", line)
-
-                uuid = fields[VG._fields.index("uuid")]
-                pvNameIdx = VG._fields.index("pv_name")
-                pv_name = fields[pvNameIdx]
-                if pv_name == UNKNOWN:
-                    # PV is missing, e.g. device lost of target not connected
-                    continue
-                if uuid not in vgsFields:
-                    fields[pvNameIdx] = [pv_name]  # Make a pv_names list
-                    vgsFields[uuid] = fields
-                else:
-                    vgsFields[uuid][pvNameIdx].append(pv_name)
-            for fields in vgsFields.values():
-                vg = VG.fromlvm(*fields)
-                if int(vg.pv_count) != len(vg.pv_name):
-                    log.error("vg %s has pv_count %s but pv_names %s",
-                              vg.name, vg.pv_count, vg.pv_name)
-                self._vgs[vg.name] = vg
-                updatedVGs[vg.name] = vg
-
-            # Remove stale VGs
-            staleVGs = [name for name in (vgNames or self._vgs)
-                        if name not in updatedVGs]
-            for name in staleVGs:
-                if name in self._vgs:
-                    log.warning("Removing stale VG %s", name)
-                    del self._vgs[name]
-                    # Remove fresh lvs indication of the vg removed from cache.
-                    self._freshlv.discard(name)
+            updatedVGs = self._updatevgs_locked(out, vgNames)
 
             # If we updated all the VGs drop stale flag
             if not vgName:
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -652,6 +652,9 @@ class LVMCache(object):
     def _reload_single_vg(self, vg_name):
         """
         Run LVM 'vgs' command and update VG name.
+        Raise VolumeGroupDoesNotExist if an LVM error occurred without output
+        or VG is stale (not updated).
+        Return the updated VG.
         """
         cmd = list(VGS_CMD)
         cmd.append(vg_name)
@@ -662,7 +665,17 @@ class LVMCache(object):
             if error:
                 self._update_stale_vgs_locked([vg_name])
 
-            return self._updatevgs_locked(out, [vg_name])
+                # Reload a specific VG name and failing
+                # might be indicative of a real error.
+                raise se.VolumeGroupDoesNotExist.from_error(vg_name, error)
+
+            updated_vgs = self._updatevgs_locked(out, [vg_name])
+
+            if vg_name not in updated_vgs:
+                # This should not happen.
+                raise se.VolumeGroupDoesNotExist(vg_name=vg_name)
+
+        return updated_vgs[vg_name]
 
     def _reloadvgs(self, vgName=None):
         """
@@ -909,12 +922,14 @@ class LVMCache(object):
         return pvs
 
     def getVg(self, vgName):
-        # Get specific VG
+        """
+        Get specific VG.
+        Raise a VolumeGroupDoesNotExist for LVM command errors or missing VG.
+        """
         vg = self._vgs.get(vgName)
         if not vg or vg.is_stale():
             self.stats.miss()
-            vgs = self._reload_single_vg(vgName)
-            vg = vgs.get(vgName)
+            vg = self._reload_single_vg(vgName)
         else:
             self.stats.hit()
         return vg
@@ -1315,11 +1330,11 @@ def movePV(vgName, src_device, dst_devices):
 
 
 def getVG(vgName):
-    vg = _lvminfo.getVg(vgName)  # returns single VG namedtuple
-    if not vg:
-        raise se.VolumeGroupDoesNotExist(vg_name=vgName)
-    else:
-        return vg
+    """
+    Return VG named tupple. Raise se.VolumeGroupDoesNotExist if the
+    VG does not exist.
+    """
+    return _lvminfo.getVg(vgName)
 
 
 def getVGs(vgNames):
--- a/tests/storage/exception_test.py
+++ b/tests/storage/exception_test.py
@@ -25,6 +25,8 @@ import six
 
 from collections import defaultdict
 
+import pytest
+
 from vdsm.common.compat import get_args_spec
 from vdsm.common.exception import GeneralException
 from vdsm.common.exception import VdsmException
@@ -58,6 +60,11 @@ def test_collisions():
 
 def test_info():
     for obj in find_module_exceptions(storage_exception, VdsmException):
+        if obj == storage_exception.VolumeGroupDoesNotExist:
+            # Skip since a constructor parameter requires a specific type.
+            # This exception should be tested separately.
+            continue
+
         # Inspect exception object initialization parameters.
         args, varargs, kwargs = get_args_spec(obj.__init__)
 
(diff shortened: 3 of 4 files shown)


lib/vdsm/storage/blockSD.py
```diff
@@ -521,14 +521,14 @@ def _validateNotVgMetadataDevice(self, guid):
                 "This PV is used by LVM to store the VG metadata")
 
     def _validatePVsPartOfVG(self, pv, dstPVs=None):
-        vgPvs = {os.path.basename(pv) for pv in lvm.listPVNames(self.sdUUID)}
+        vgPvs = {os.path.basename(_pv) for _pv in lvm.listPVNames(pv.vg_name)}
         if pv not in vgPvs:
-            raise se.NoSuchPhysicalVolume(pv, self.sdUUID)
+            raise se.NoSuchPhysicalVolume(pv, pv.vg_name)
```
vjuranek, line 526: sdUUID should be the same as vgUUID. Good to make it more clear and use vgUUID, but this is not a real issue.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,42 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+
+    # VG is returned with valid VG name.
+    vg = lc.getVg(vg_name)
+    assert vg.name == vg_name
+
+    # Non-existing VG name raises an exception.
+    with pytest.raises(se.VolumeGroupDoesNotExist) as e:
+        lc.getVg(non_vg_name)
+    # The name of the non-found VG must appear in the exception.
+    assert non_vg_name in e.value.value
```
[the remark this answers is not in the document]
aesteve-rh: Because the error message may change in the future, be reformulated a bit, and it would be a bit of a burden having to update the check in all asserts that look into the error message.

The relevant part for me was that the vg_name that caused the exception to rise appears in the message, and so this check.

But I'm ok with changing so that the whole error message is checked with `==` if you think that adds verification value.

lib/vdsm/storage/lvm.py
```diff
@@ -887,15 +887,19 @@ def getVg(self, vgName):
...
-        Fills the cache but not uses it.
-        Only returns found VGs.
+        Fill the cache but do not use it.
+        Return only found VGs.
         """
         self.stats.miss()
-        return [vg for vgName, vg in six.iteritems(self._reloadvgs(vgNames))
-                if vgName in vgNames]
+        try:
+            return [vg for vgName, vg in six.iteritems(
```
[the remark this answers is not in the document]
vjuranek: It was in the past, when we had to support both python 2 and 3. Now we support only python 3, so `six` can be (and shold be) removed. This is done incrementally, typically when touching some piece of code, it's updated accordingly.

lib/vdsm/storage/blockSD.py
aesteve-rh: No! Thanks for explaining!

And yes, I removed the commit because it came from my confusion between `sdUUID` and `vgUUID`. Good thing I isolated the change in a single commit and was easy to just get rid of it :)

lib/vdsm/storage/lvm.py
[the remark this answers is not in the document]
aesteve-rh: I have removed it from the whole module in a separate commit, there were not that many appearances. I will take it into account from now on.

lib/vdsm/storage/lvm.py
```diff
@@ -902,7 +902,7 @@ def getVgs(self, vgNames):
             return []
 
     def getAllVgs(self):
-        # Get everything we have
+        """Get everything we have."""
```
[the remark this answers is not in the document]
aesteve-rh: Done!

lib/vdsm/storage/blockSD.py
aesteve-rh: OK I will reopen this thread.

`sdUUID` and `vgUUID` should be the same? Because they do not seem to be the same in `blocksd_test.py`

```python
    sd_uuid = str(uuid.uuid4())
    lvm.createVG(sd_uuid, [dev], blockSD.STORAGE_UNREADY_DOMAIN_TAG, 128)
    vg = lvm.getVG(sd_uuid)

    dom = blockSD.BlockStorageDomain.create(
        sdUUID=sd_uuid,
        domainName="domain",
        domClass=sd.DATA_DOMAIN,
        vgUUID=vg.uuid,
        version=3,
        storageType=sd.ISCSI_DOMAIN)
```

All the `creates` in the tests are like this, and there both `sdUUID` and `vgUUID` are actually VG name and VG UUID, respectively.

And then both `checkVGBlockSizes` and `getVGBlockSizes`  in the `lvm` module receive `vgName` instead of `vgUUID`. 
It was very misleading because
1. The parameter in `checkVGBlockSizes` and `getVGBlockSizes` was called `vgUUID`, and it was used as vgName
2. `blockSD` was sending `sdUUID` that works as a synonym for `vgUUID`, but it is actually a vgName

That is why it worked, but it was fairly easy to misinterpret. I need to revert some changes and do some renaming. In general, all these names need some cleanage/homogenization, I might do it in a different branch.

vjuranek: > sdUUID and vgUUID should be the same? 

Sorry, I was wrong, `sdUUID` should be the same as `vg_name`.

lib/vdsm/storage/lvm.py
```diff
@@ -887,15 +887,19 @@ def getVg(self, vgName):
...
-        Only returns found VGs.
+        Fill the cache but do not use it.
+        Return only found VGs.
         """
         self.stats.miss()
-        return [vg for vgName, vg in six.iteritems(self._reloadvgs(vgNames))
-                if vgName in vgNames]
+        try:
+            return [vg for vgName, vg in six.iteritems(
+                self._reloadvgs(vgNames)) if vgName in vgNames]
```
nirs: indentation does not match how we use list comprehensions. It should be:

    [vg for vgName, vg
     in six.iteritems(self._reloadvgs(vgNames)) if vgName in vgNames]

Or:

    [vg for vgName, vg
     in six.iteritems(self._reloadvgs(vgNames))
     if vgName in vgNames]

Or just simplify by removing the useless siz.iteritems now:

    [vg for vgName, vg in self._reloadvgs(vgNames).items()
     if vgName in vgNames]

And the code will be more clear if we separte the call that can fail:

    try:
        vgs = self._reloadvgs(vgNames)
    except se.VolumeGroupDoesNotExist as e:
        handle the error...

    return [vg for vgName, vg in vgs.items() if vgName in vgNames]

tests/storage/lvm_test.py
```diff
@@ -1059,6 +1059,37 @@ def test_vg_get_error(fake_devices):
     assert "Fake lvm error" in e.value.value
 
 
+def test_vg_get_vgs_with_non_existing_vg(fake_devices):
```
nirs, line 1074: Better test this using lvm.getVGs() with real vgs. If we cannot covert all the interesting cases you can add a test using fakes.

We already test this in https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/tests/storage/lvm_test.py#L1255
so we may not need another test.

lib/vdsm/storage/lvm.py
```diff
@@ -913,8 +917,12 @@ def getAllVgs(self):
...
                 for name in stalevgs:
                     del vgs[name]
-                reloaded = self._reloadvgs(stalevgs)
-                vgs.update(reloaded)
+                try:
+                    reloaded = self._reloadvgs(stalevgs)
+                    vgs.update(reloaded)
+                except se.VolumeGroupDoesNotExist:
+                    # LVM error prevents reload and return stale VGs
+                    pass
```
nirs, line 924: This again must be done before or together with the change raising in _reloadvgs().

We cannot have a change that breaks some callers and fix this in later commits.

tests/storage/lvm_test.py
```diff
@@ -1090,6 +1090,95 @@ def test_vg_get_vgs_with_non_existing_vg(fake_devices):
     assert len(vgs) == 0
 
 
+def test_vg_get_all_miss_all_stale_but_one(fake_devices):
```
nirs, line 1105: This is already tested in
- https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/tests/storage/lvm_test.py#L1529
- https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/tests/storage/lvm_test.py#L1547
- https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/tests/storage/lvm_test.py#L1564
- https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/tests/storage/lvm_test.py#L1564

We may not need any new tests.

lib/vdsm/storage/lvm.py
```diff
@@ -1318,7 +1318,7 @@ def getVGbyUUID(vgUUID):
             log.debug("%s", e, exc_info=True)
             continue
     # If not cry loudly
-    raise se.VolumeGroupDoesNotExist("vg_uuid: %s" % vgUUID)
+    raise se.VolumeGroupDoesNotExist(UNKNOWN, f"Not found vg_uuid {vgUUID}")
```
nirs, line 1320: Looks like we are abusing se.VolumeGroupDoesNotExist here. Since engine is using this
error, we probably need to keep using it for both case for looking up VG by name and uuid, so need to make the error more flexible, so we can use it for both cases:

    se.VolumeGroupDoesNotExist(vg_uuid="vg-uuid")
    se.VolumeGroupDoesNotExist(vg_name="vg-name")
    se.VolumeGroupDoesNotExist(vg_name="vg-name", error=e)

lib/vdsm/storage/lvm.py
```diff
@@ -37,7 +37,6 @@
 import threading
 
 from itertools import chain
-import six
```
nirs, line 40: This is a good change but it does not belong to this PR. Please send it as a separate PR.

tests/storage/lvm_test.py
```diff
@@ -1059,6 +1059,18 @@ def test_vg_get_error(fake_devices):
...
 
 
+def test_vg_get_invalid_output(monkeypatch, fake_devices):
+    fake_runner = FakeRunner(out=b"Fake lvm output")
+    lc = lvm.LVMCache(fake_runner)
+
+    monkeypatch.setattr(lvm, "_lvminfo", lc)
+
+    with pytest.raises(lvm.InvalidOutputLine) as e:
+        lvm.getVG("vg-name")
```
nirs, line 1069: This will raise in this line:
https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/lib/vdsm/storage/lvm.py#L616

Are you sure this is correct?

tests/storage/lvm_test.py
```diff
@@ -1179,6 +1191,32 @@ def test_vg_get_all_hit_no_stale(fake_devices):
     assert len(vgs) == num_vgs
 
 
+def test_vg_remove_by_uuid(monkeypatch, fake_devices):
```
nirs, line 1194: We should test this using real vg, not fakes.

You can add a simple test like:
https://github.com/oVirt/vdsm/blob/013692737dc9dd89b977cb8a91e479c8757fd0db/tests/storage/lvm_test.py#L921

lib/vdsm/storage/lvm.py
```diff
@@ -1384,24 +1384,25 @@ def removeVGbyUUID(vgUUID):
         removeVG(vg.name)
```
nirs, line 1384: The commit message suggests that this change add missing tests, but the actual change looks like change in behavior, like acccepting vg uuid instead of vg name, and doing a look up of vg by UUID.

Please add the missing tests before changing the behavior to make sure we don't break callers.

lib/vdsm/storage/lvm.py
```diff
@@ -1384,24 +1384,25 @@ def removeVGbyUUID(vgUUID):
...
 
 
-def extendVG(vgName, devices, force):
+def extendVG(vgUUID, devices, force):
```
nirs, line 1387: This looks wrong. 

We have single caller:

    lib/vdsm/storage/blockSD.py:607: lvm.extendVG(self.sdUUID, devlist, force)

sdUUID is the storage domain UUID, which is the vg name (created by oVirt),
not the vg UUID (created by lvm).

Since this is called with a vg name, the entire change is incorrect.

tests/storage/lvm_test.py
```diff
@@ -1217,6 +1217,120 @@ def test_vg_remove_by_uuid(monkeypatch, fake_devices):
     assert len(lc._vgs) == len(fake_vgs) - 1
 
 
+def test_vg_check_block_sizes(monkeypatch, fake_devices):
```
nirs, line 1220: The test should be done using real pv and vg and no mocking.

We don't support 4k storage yet - if we will add such support, we can test it using
userstorage.

lib/vdsm/storage/blockSD.py
```diff
@@ -521,14 +521,14 @@ def _validateNotVgMetadataDevice(self, guid):
                 "This PV is used by LVM to store the VG metadata")
 
     def _validatePVsPartOfVG(self, pv, dstPVs=None):
-        vgPvs = {os.path.basename(pv) for pv in lvm.listPVNames(self.sdUUID)}
+        vgPvs = {os.path.basename(_pv) for _pv in lvm.listPVNames(pv.vg_name)}
```
nirs, line 524: Change is not needed since self.sdUUID is already a VG name.

Using _pv instead of pv to avoid shadowing the argument is good idea, in the current code it is not clear if the temporary variable pv in the set comprehension changes the argument, but we don't use private name for temporary arguments.

Using _pv also wrong like the previous code, since lvm.listPVNames() returns pv path, and not a pv named tuple. We should really use pv, vg, lv with lvm API only when we get back a named tuple.

We can use something like:

    {os.path.basename(path) for path in lvm.listPVNames(self.sdUUID)}  

to make the code more clear.

But this change doe snot belong to this PR.

nirs (changes requested):
Lets try to minimize the scope of this PR, include only the changes required to improve error handling, required for #159.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,42 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+
+
+def test_vg_get_error(fake_devices):
+    fake_runner = FakeRunner(rc=5, err=b"Fake lvm error")
+    lc = lvm.LVMCache(fake_runner)
+
+    with pytest.raises(se.VolumeGroupDoesNotExist) as e:
+        lc.getVg("non-existing-vg-name")
+    # Ensure that the error is captured and printed in the exception.
+    assert "Fake lvm error" in e.value.value
```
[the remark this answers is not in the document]
aesteve-rh: This is testing actually the usecase we have added in the branch for raising on a real lvm error. And for testing that, we need a fake runner. I'll rework the tests, removing some tests already covered with non-faked devices (as for your previous comment), and then see with which other test this fits best and merge them.

lib/vdsm/storage/lvm.py
aesteve-rh: `six` is removed in a different commit, but I will separate the call to reloadvgs.

The line split was done by `autopep8`. Actually, to me separating the for in two lines looks weird, but is good to know, I will take it into account from now on. Anyway, for things like this, it would be nice to have `coding guidelines` documented.

tests/storage/lvm_test.py
aesteve-rh: I will remove it then.

Last week I was having issues running sudo tests, so I pushed to increase coverage with non-sudo tests.

lib/vdsm/storage/lvm.py
aesteve-rh: Right, I realized when it was already late last Friday. All started with the methods below that were doing:
```
pvs = listPVNames(vgUUID)
```
But `listPVNames` was expecting a vg_name, not the UUID, so I changed it to translate it from uuid to name before passing it to `listPVNames`. And that triggered other changes like this one. I will revert these lines, since the problem was in the name of the parameter in `def checkVGBlockSizes(vgUUID, vgBlkSize=None):`. It should be either sdUUID or (preferably) vgName, not vgUUID.

tests/storage/lvm_test.py
aesteve-rh: What do you mean? That line was not covered, so I thought that it might be nice to slightly increase code coverage by testing that an invalid output of the lvm command will raise the expected exception. And by doing that, we get rid of the last red line in the test coverage summary for the reload method.

I'd say it is correct.

nirs: Right, I missed the check for lvm.InvalidOutputLine. We can keep this test, but it should check also how the failure affects the cache - I think the cache is not affected by this since the command did not fail. Not sure if this is good or bad, but this case is not really expected. I would add a TODO: to think if this is the best way to handle this unlikely error.

tests/storage/lvm_test.py
aesteve-rh: Right, I have removed them.

tests/storage/lvm_test.py
aesteve-rh: I have removed these for now, we can discuss if it is worth the effort or not to add them in a separate branch.

lib/vdsm/storage/lvm.py
```diff
@@ -604,6 +604,12 @@ def _reloadvgs(self, vgName=None):
                         "Marked vgs=%r as Unreadable due to reload failure",
                         logutils.Head(unreadable_vgs, max_items=20))
 
+                if len(vgNames) == 1 and not out:
+                    # Reload a specific VG name and failing without
+                    # an output might be indicative of a real error.
+                    raise se.VolumeGroupDoesNotExist.from_vg_name(
+                        vgName, str(error))
```
nirs, line 611: Better to pass the error as is - this allow the caller to do later:

    except se.VolumeGroupDoesNotExist as e:
        if e.rc == ...

And it will work in the same way for formatting the error, since formatting will convert the error to a string.

nirs (changes requested):
Partial review, will continue later.

lib/vdsm/storage/exception.py
```diff
@@ -1349,6 +1349,29 @@ class VolumeGroupDoesNotExist(StorageException):
...
+    def value(self):
+        return f"{self._log_vg()} error={self.error}"
+
+    @classmethod
+    def from_vg_name(cls, vg_name, error=''):
+        return cls(vg_name=vg_name, error=error)
+
+    @classmethod
+    def from_vg_uuid(cls, vg_uuid):
+        return cls(vg_uuid=vg_uuid)
```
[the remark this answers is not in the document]
aesteve-rh: The idea was to discourage the direct use of the constructor, because with all parameters optional it could be as well be misused and constructed without any parameter.

It still can, but at least the API hints how to use it. I was pondering different designs... One alternative would be to
```python
    assert vg_name or vg_uuid
```
... in the constructor of the exception. Or log an error if we don't want to raise from within an exception.

lib/vdsm/storage/lvm.py
aesteve-rh: Uhm, then error should be `Optional[LVMCommandError]`?

In that case, if we want to support custom error messages as strings (was my idea anyway), we would need a different `reason` parameter. Otherwise, in reusing `error` we cannot assume that `e.error.rc` is available (it could be either `LVMCommandError` or `str` or `None`).

In any case, even with `error` and `reason`, we would still need to do something like:
```python
except se.VolumeGroupDoesNotExist as e:
    if e.error and e.error.rc == ...
```

lib/vdsm/storage/exception.py
```diff
@@ -1349,6 +1349,29 @@ class VolumeGroupDoesNotExist(StorageException):
...
+
+    def _log_vg(self):
+        if self.vg_name:
+            return f"vg_name={self.vg_name}"
+
+        return f"vg_uuid={self.vg_uuid}"
+
+    @property
+    def value(self):
+        return f"{self._log_vg()} error={self.error}"
```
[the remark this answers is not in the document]
aesteve-rh: I did something similar, thanks for the suggestion :)

lib/vdsm/storage/lvm.py
aesteve-rh: I was also thinking of adding this to the constructor:
```python
    assert error is None or isinstance(error, LVMCommandError)
```

lib/vdsm/storage/lvm.py
nirs: We can add coding guidelines, not sure how easy it will be to get something that everyone agree with, but we can try.

lib/vdsm/storage/lvm.py
nirs: It makes sense and makes the code clear about the intent. In the worst case we will find that we need to pass another error in some case and we will make the type less strict.

There is one issue, assert statement may be removed in runtime if someone run this code with python -O option. So it is better to do:

    if error is not None and not isinstance(error, LVMCommandError):
        raise TypeError(
            f"Expecting instance of LVMCommandError, got {type(error)}")

lib/vdsm/storage/lvm.py
aesteve-rh: Lol, I know how difficult can it be to agree on something. Specially when some people have very strong opinions on some particular topics.

Personally, I do prefer to have coding guidelines, even if I don't like all of them. Helps to have the code more consistent, and not 'waste time' thinking on how to format the code. That is why I like having an auto formatter at hand.

But we can talk about it next week to see if it is worth or not. It is something more contributors have to agree on.

nirs (changes requested):

tests/storage/lvm_test.py
```diff
@@ -1093,6 +1093,28 @@ def test_vg_invalid_output(monkeypatch, fake_devices):
...
+    lc = lvm.LVMCache(fake_runner)
+    lc._stalevg = False
+
+    monkeypatch.setattr(lvm, "_lvminfo", lc)
+
+    # Create fake VG with no PVs.
+    fake_vg = make_vg(pvs=[], vg_name="vg")
+
+    # Assign fake VG to cache.
+    lc._vgs = {fake_vg.name: fake_vg}
```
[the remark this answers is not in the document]
aesteve-rh: I will remove the test after fixing the methods.

lib/vdsm/storage/lvm.py
```diff
@@ -867,29 +872,44 @@ def getPvs(self, vgName):
...
+        """
+        Get specific VG.
+        Raise a VolumeGroupDoesNotExist for LVM command errors or missing VG.
+        """
         vg = self._vgs.get(vgName)
         if not vg or vg.is_stale():
             self.stats.miss()
             vgs = self._reloadvgs(vgName)
+            if vgName not in vgs:
+                raise se.VolumeGroupDoesNotExist(vg_name=vgName)
```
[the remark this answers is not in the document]
aesteve-rh: I have done the refactor, trying to minimize code duplication. I have done it in a separate commit, but still not the easiest to read for review.

It really allows it to have clear behavior with different function for single- and multi- VG reloading. It will be nicer also when we extend this strategy to LVs and PVs.

lib/vdsm/storage/lvm.py
```diff
@@ -910,29 +916,39 @@ def getPvs(self, vgName):
...
     def getVgs(self, vgNames):
-        """Reloads all the VGs of the set.
+        """
+        Reload all the VGs of the set.
 
         Can block for suspended devices.
-        Fills the cache but not uses it.
-        Only returns found VGs.
+        Fill the cache but do not use it.
+        Return only found VGs.
```
nirs, line 940: Good changes, but not needed for this PR. Lets remove every change we don't need now to minimize this PR.

lib/vdsm/storage/lvm.py
```diff
@@ -910,29 +916,39 @@ def getPvs(self, vgName):
...
         return [vg for vgName, vg in self._reloadvgs(vgNames).items()
                 if vgName in vgNames]
 
     def getAllVgs(self):
-        # Get everything we have
+        """
+        Get everything we have.
+        Reload all VGs. Even on cache hits, reload stale VGs.
+        Return only found VGs.
+        """
```
nirs, line 951: Since we don't need to modify this code now, better not touch the docstring now. We can do this later in a cleanup PR.

nirs (changes requested):

lib/vdsm/storage/lvm.py
```diff
@@ -577,74 +577,117 @@ def _getVGDevs(self, vgNames):
                 devices = tuple(devices)
```
nirs, line 577: Note for the commit message - I think you mean:

    Extract helper for reloading vgs

lib/vdsm/storage/lvm.py
```diff
@@ -577,74 +577,117 @@ def _getVGDevs(self, vgNames):
...
+
+    def _updatevgs(self, lvm_output, vgs):
+        """
+        Update cached VGs based on the output of the LVM command.
+        Remove stale VGs (have not been updated) out of the 'vgs' list, or
+        the complete list of cached VGs if the command was run for all VGs.
+        Return list of all updated VGs.
+        """
+        assert self._lock.locked
+        upd_vgs = {}
```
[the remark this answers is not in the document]
aesteve-rh: ...to make it fit in one line when used in a generator below. Anyway, I will revert it to `updatedVGs`, was its original name, so that I restrain additional changes.

lib/vdsm/storage/exception.py
```diff
@@ -1349,6 +1349,22 @@ class VolumeGroupDoesNotExist(StorageException):
     code = 506
```
nirs, line 1349: Note for the commit message - title should be:

    storage.exception: Add error to VolumeGroupDoesNotExist

tests/storage/exception_test.py
```diff
@@ -77,6 +84,25 @@ def test_info():
...
+    # Empty constructor shall raise.
+    with pytest.raises(ValueError):
+        e = storage_exception.VolumeGroupDoesNotExist()
+
+    # Expected error type is LVMCommandError.
+    with pytest.raises(TypeError):
+        e = storage_exception.VolumeGroupDoesNotExist("vg", error="error")
+
+    # Correct initialization, test info.
+    lvm_error = storage_exception.LVMCommandError(0, 1, 2, 3)
```
nirs, line 98: I would make this more real:

    fake_error = storage_exception.LVMCommandError(
        rc=5, command=["fake"], out=["fake output"], err=["fake error"])

tests/storage/exception_test.py
```diff
@@ -77,6 +84,25 @@ def test_info():
...
+    e = storage_exception.VolumeGroupDoesNotExist("vg", error=lvm_error)
+    assert e.info() == {
+        "code": e.code,
+        "message": str(e)
+    }
```
nirs, line 103: I would test the formatted stringf instead of info(), so we can change the implementation later.

    e = storage_exception.VolumeGroupDoesNotExist("vg-name", error=lvm_error)
    assert e.error = fake_error

    formatted = str(e)
    assert "vg_name=vg-name" in formatted
    assert f"error={fake_error}" in formatted
    assert "vg_uuid=" not in formatted

lib/vdsm/storage/lvm.py
```diff
@@ -577,74 +577,117 @@ def _getVGDevs(self, vgNames):
...
+                # don't return here.
+
+            return self._updatevgs(out, vgNames)
+
+    def _reloadvgs(self, vgName=None):
+        """
+        Run LVM 'vgs' command and update VG names.
+        If no VG name is provided, reload all VGs.
+        """
         vgNames = normalize_args(vgName)
```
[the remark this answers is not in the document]
aesteve-rh: I 'strongly' agree in this. I will do it in a separate branch.

nirs (changes requested):

lib/vdsm/storage/lvm.py
```diff
@@ -1486,20 +1486,14 @@ def _checkpvsblksize(pvs, vgBlkSize=None):
             raise se.VolumeGroupBlockSizeError(vgBlkSize, pvBlkSize)
```
nirs, line 1545: Commit message need update

- we did not add reason argument.
- we simplified listPVNames() - need to explain why the previous implementation was wrong. This also worth its own commit to to keep changes in real code minimal.
- we removed the code raising if we get no pv, need to explain why this is not needed. I think it can be part of the commit simplifying listPVNames(), if I remember the reason corectly.

nirs (changes requested):

tests/storage/exception_test.py
```diff
@@ -77,6 +84,28 @@ def test_info():
...
+
+    # Correct initialization.
+    fake_error = storage_exception.LVMCommandError(
+        rc=5, cmd=["fake"], out=["fake output"], err=["fake error"])
+    e = storage_exception.VolumeGroupDoesNotExist("vg-name", error=fake_error)
+    assert e.error == fake_error
+    # Check error format
+    formatted = str(e)
+    assert "vg_name=vg-name" in formatted
+    assert f'error={fake_error}'.replace('\'', r'\'') in formatted
```
nirs, line 105: This is a little bit ugly and fragile in case python changes the way it format quoted string, but I don't have a better way.

lib/vdsm/storage/lvm.py
```diff
@@ -577,6 +577,73 @@ def _getVGDevs(self, vgNames):
                 devices = tuple(devices)
         return devices
 
+    def _updatevgs(self, vgs_output, vg_names):
+        """
+        Update cached VGs based on the output of the LVM command.
+        Remove stale VGs (have not been updated) out of the 'vg_names' list,
```
nirs, line 583: This is little bit confusing - we don't remove names from vg_names list.

Describing the side effects of this function can be something like this:

- Add new vgs to the cache
- Replace vgs in the cache with vgs reported by the vgs command, updating the vg attributes.
- If called without vg names, remove all vgs from the cache not reported by lvm
- If called with vg names, remove specifed vgs from the cache if they were not reported by lvm.

I think the best way to describe this is to have tests verifying the contents of the cache after the call with all cases. Using fake runner with fake lvm output is the easier way to test this. However since this patch just simple move, I don't think we need to spend time on the tests now, unless the code you are moving is not covered by the tests.

lib/vdsm/storage/lvm.py
```diff
@@ -589,59 +656,12 @@ def _reloadvgs(self, vgName=None):
...
-            # Remove stale VGs
-            staleVGs = [name for name in (vgNames or self._vgs)
-                        if name not in updatedVGs]
-            for name in staleVGs:
-                if name in self._vgs:
-                    log.warning("Removing stale VG %s", name)
-                    del self._vgs[name]
-                    # Remove fresh lvs indication of the vg removed from cache.
-                    self._freshlv.discard(name)
+            updatedVGs = self._updatevgs(out, vgNames)
```
nirs, line 697: Nice clean and easy to review refactoring!

lib/vdsm/storage/lvm.py
```diff
@@ -657,10 +660,20 @@ def _reload_single_vg(self, vg_name):
             if error:
                 self._update_stale_vgs([vg_name])
 
+                if not out:
+                    # Reload a specific VG name and failing without
+                    # an output might be indicative of a real error.
+                    raise se.VolumeGroupDoesNotExist(vg_name, error=error)
```
nirs, line 666: One small detail to make a more useful traceback. If we do:

    raise se.VolumeGroupDoesNotExist(vg_name, error=error).with_exception(error)

We will have single traceback for the entire call - from the caller frame to the frame calling the lvm command. In the current way we have 2 tracebacks - one from the call frame to this point, and another from this point to the call to lvm command.

Maybe we need a class method like LVMCommandError.from_error(), that will set the instance error attribute, and copy the error.__traceback__.

lib/vdsm/storage/lvm.py
```diff
@@ -657,10 +660,20 @@ def _reload_single_vg(self, vg_name):
...
+
                 # NOTE: vgs may return useful output even on failure, so we
                 # don't return here.
 
-            return self._updatevgs(out, [vg_name])
+            updated_vgs = self._updatevgs(out, [vg_name])
+
+            if vg_name not in updated_vgs:
+                raise se.VolumeGroupDoesNotExist(vg_name=vg_name)
```
nirs, line 674: I don't think this is possible - if the command succeeded, we got output for the vg. If the output is invalid, we will raise InvalidOutputLine. Otherwise we will get a vg object. However if my assumption is wrong, raising is the best thing we can do. Maybe a comment that this should never happen?

lib/vdsm/storage/lvm.py
```diff
@@ -657,10 +660,20 @@ def _reload_single_vg(self, vg_name):
             if error:
                 self._update_stale_vgs([vg_name])
 
+                if not out:
```
nirs, line 663: I'm not sure it is possible to get an error from vgs and some output, so we can simplify by raising on errors.

lib/vdsm/storage/lvm.py
```diff
@@ -657,10 +660,20 @@ def _reload_single_vg(self, vg_name):
...
+                    # Reload a specific VG name and failing without
+                    # an output might be indicative of a real error.
+                    raise se.VolumeGroupDoesNotExist(vg_name, error=error)
+
                 # NOTE: vgs may return useful output even on failure, so we
                 # don't return here.
```
nirs, line 669: Since error with output should not be possible, we can remove this comment.

nirs (commented):
Awasome!

lib/vdsm/storage/lvm.py
```diff
@@ -577,6 +577,73 @@ def _getVGDevs(self, vgNames):
                 devices = tuple(devices)
         return devices
 
+    def _updatevgs(self, vgs_output, vg_names):
```
nirs, line 580: Maybe to make it more clear that this must be called while holding the lock, call the helpers:

    _update_vgs_locked()
    _update_stale_vgs_locked()

lib/vdsm/storage/lvm.py
aesteve-rh: Yes, keeping the traceback is always a good idea. We will benefit from it the next time it fails, and some user reports with logs.

tests/storage/exception_test.py
aesteve-rh: A little, yes. If that gives us issues in the midterm, we can change the check to something else.

lib/vdsm/storage/lvm.py
```diff
@@ -577,6 +577,78 @@ def _getVGDevs(self, vgNames):
...
+        Update cached VGs based on the output of the LVM command:
+        - Add new VGs to the cache.
+        - Replace VGs in the cache with VGs reported by the 'vgs' command,
+          updating the VG attributes.
+        - If called without vg names, remove all VGs from the cache not
+          reported by LVM.
+        - If called with vg names, remove specifed VGs from the cache if they
+          were not reported by LVM.
+        Must be called while holding the lock.
+        Return list of all updated VGs.
```
nirs, line 591: Return dict of updated VGs.

nirs (dismissed):

nirs (approved):

Issue #43, opened 2022-01-15
Remove legacy automation


nirs:
- Rename automation to ci
- Remove legacy ovirt ci files

Issues:
- check-patch.packages is used in README.md for development setup. Create a script for setting up development environment?

PR #162, opened 2022-05-09
net: Add python3-libnmstate as explicit dependency

almusil:
With changes to nmstate-2.1 the python3-libnmstate
is not installed by default. This causes issues for
the unit tests as they need the library to work properly.

Add the python3-libnmstate as explicit dependency into
spec file as well, to be clear that this is required.

### DIFF
--- a/automation/check-patch.packages
+++ b/automation/check-patch.packages
@@ -28,6 +28,7 @@ python3-decorator
 python3-devel
 python3-dmidecode
 python3-ioprocess
+python3-libnmstate
 python3-libselinux
 python3-libvirt
 python3-magic
--- a/docker/network/unit/Dockerfile.centos-8
+++ b/docker/network/unit/Dockerfile.centos-8
@@ -14,6 +14,7 @@ RUN dnf -y install dnf-plugins-core \
         libnl3 \
         nmstate \
         python3-devel \
+        python3-libnmstate \
         python3-six \
         python3-pip \
         systemd \
--- a/docker/network/unit/Dockerfile.centos-9
+++ b/docker/network/unit/Dockerfile.centos-9
@@ -23,6 +23,7 @@ RUN dnf -y install dnf-plugins-core \
         libnl3 \
         nmstate \
         python3-devel \
+        python3-libnmstate \
         python3-six \
         python3-pip \
         systemd \
--- a/vdsm.spec.in
+++ b/vdsm.spec.in
@@ -379,6 +379,7 @@ Requires:       ovirt-openvswitch >= 2.15
 Requires:       ovirt-python-openvswitch >= 2.15
 Requires:       nmstate >= 1.0
 Requires:       nmstate-plugin-ovsdb
+Requires:       python3-libnmstate
 Requires:       libnl3
 Requires:       lldpad
 Requires:       python3-six


nirs (commented):
Looks good but seems that doc/development.md needs an update.

hbraha (dismissed):

almusil:
> Looks good but seems that doc/development.md needs an update.

Which part? Nothing mentions network there.

erav (dismissed):

nirs:
> > Looks good but seems that doc/development.md needs an update.
> 
> Which part? Nothing mentions network there.

I see that we did not update yet the example for installing dependencies
using the legacy automation/ directory (#43).

Ideally this will be replace with something like:
https://github.com/oVirt/ovirt-imageio/blob/master/docs/development.md#setting-up-development-environment

With this change libnmstate may be missing on developer machine since we
don't document that it is needed.

almusil:
> > > Looks good but seems that doc/development.md needs an update.
> > 
> > 
> > Which part? Nothing mentions network there.
> 
> I see that we did not update yet the example for installing dependencies using the legacy automation/ directory (#43).
> 
> Ideally this will be replace with something like: https://github.com/oVirt/ovirt-imageio/blob/master/docs/development.md#setting-up-development-environment
> 
> With this change libnmstate may be missing on developer machine since we don't document that it is needed.

Alright, but that's out of scope of this patch. Also one thing to note is that this package will always be installed on el8 with just `nmstate`, on el9 it will be installed with `nmstate-plugin-ovsdb`. Having it explicitly in the spec file basically does nothing except documenting that we need this dependency.

almusil:
> > > > > Looks good but seems that doc/development.md needs an update.
> > > > 
> > > > 
> > > > Which part? Nothing mentions network there.
> > > 
> > > 
> > > I see that we did not update yet the example for installing dependencies using the legacy automation/ directory (#43).
> > > Ideally this will be replace with something like: https://github.com/oVirt/ovirt-imageio/blob/master/docs/development.md#setting-up-development-environment
> > > With this change libnmstate may be missing on developer machine since we don't document that it is needed.
> > 
> > 
> > Alright, but that's out of scope of this patch. Also one thing to note is that this package will always be installed on el8 with just `nmstate`, on el9 it will be installed with `nmstate-plugin-ovsdb`. Having it explicitly in the spec file basically does nothing except documenting that we need this dependency.
> 
> Adding to the spec is not documenting anything, you don't expect developer to read the spec when preparing a development environment.
> 
> Can you add a little section for installing additional packages to the document?

I can add to the check-patch.packages, as the document is referring to that, which should be fixed, but it's still outside of scope of this patch.

nirs:
> > > > > > Looks good but seems that doc/development.md needs an update.
> > > > > 
> > > > > 
> > > > > Which part? Nothing mentions network there.
> > > > 
> > > > 
> > > > I see that we did not update yet the example for installing dependencies using the legacy automation/ directory (#43).
> > > > Ideally this will be replace with something like: https://github.com/oVirt/ovirt-imageio/blob/master/docs/development.md#setting-up-development-environment
> > > > With this change libnmstate may be missing on developer machine since we don't document that it is needed.
> > > 
> > > 
> > > Alright, but that's out of scope of this patch. Also one thing to note is that this package will always be installed on el8 with just `nmstate`, on el9 it will be installed with `nmstate-plugin-ovsdb`. Having it explicitly in the spec file basically does nothing except documenting that we need this dependency.
> > 
> > 
> > Adding to the spec is not documenting anything, you don't expect developer to read the spec when preparing a development environment.
> > Can you add a little section for installing additional packages to the document?
> 
> I can add to the check-patch.packages, as the document is referring to that, which should be fixed, but it's still outside of scope of this patch.

Adding to the packages list also works. I think this is in the scope of this change.
When adding new package, it must be added to the spec, docker files, and packages list.

nirs (approved):
Thanks!

PR #164, opened 2022-05-09
lvm_tests: use --devices instead of --config in commands

### LINKED ISSUE #163: lvm_tests: tests should not assume system configuration
### Description

LVM tests try to fake a config using lvm filter, but this cannot work for testing real lvm commands. The test must not assume anything about the system configuration.

This command is wrong:
```python
    # Simulate removal of the second PV on another host, leaving stale PV in
    # the cache.
    commands.run([
        "vgreduce",
        "--config", tmp_storage.lvm_config(),
        vg_name,
        stale_pv_name,
    ])

```

The test and the code can also not change the system configuration, it should adapt to it and use `--devices` on this system.

### Reproduce

Run a single test that employs the `stale_pv` fixture as root on a CentOS stream 8 or other supported system, e.g.:
- `sudo ~/.venv/vdsm/bin/tox -e storage tests/storage/lvm_test.py -- -k test_pv_stale_reload_one_stale`

Test will fail with:
```
E           vdsm.common.cmdutils.Error: Command ['vgreduce', '--config', 'global { use_lvmetad=0 } devices { write_cache_state=0 filter=["a|^/dev/loop10$|^/dev/loop11$|", "r|.*|"] }', '5f6c5d91-b734-4785-99a8-13aa137f61c0', '/dev/loop11'] failed with rc=5 out=b'' err=b'  Cannot use /dev/loop11: device is not in devices file\n  Volume group "5f6c5d91-b734-4785-99a8-13aa137f61c0" not found\n  Cannot process volume group 5f6c5d91-b734-4785-99a8-13aa137f61c0\n'
```

### Actions to fix it

We can fix this in two steps:
1. Check how we build the command in the real code, and replace the manual code
2. Ideally, we should really run these tests in a new VM started for this purpose, to allow quick feedback

aesteve-rh:
Some commands in lvm_tests try to force the configuration by passing the `--config` option, whereas they should adapt to it and use `--devices` instead.

Fixes #163 (covers solution 1)

### DIFF
--- a/tests/storage/lvm_test.py
+++ b/tests/storage/lvm_test.py
@@ -1937,7 +1937,7 @@ def stale_lv(tmp_storage):
     commands.run([
         "lvremove",
         "--devices", dev,
-        "{}/{}".format(vg_name, stale_lv_name),
+        f"{vg_name}/{stale_lv_name}",
     ])
 
     # The cache still keeps both lvs.
--- a/tests/storage/lvm_test.py
+++ b/tests/storage/lvm_test.py
@@ -1627,6 +1627,7 @@ def test_lv_activate_deactivate(tmp_storage):
 @requires_root
 @pytest.mark.root
 def test_lv_deactivate_in_use(tmp_storage):
+    # TODO: This test does not work in active hosts.
     dev_size = 1 * GiB
     dev = tmp_storage.create_device(dev_size)
     vg_name = str(uuid.uuid4())
--- a/tests/storage/lvm_test.py
+++ b/tests/storage/lvm_test.py
@@ -1038,17 +1038,18 @@ def stale_pv(tmp_storage):
     pvs = sorted(pv.name for pv in lvm.getAllPVs())
     assert pvs == sorted([good_pv_name, stale_pv_name])
 
+    devices = ",".join([good_pv_name, stale_pv_name])
     # Simulate removal of the second PV on another host, leaving stale PV in
     # the cache.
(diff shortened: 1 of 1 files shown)


tests/storage/lvm_test.py
```diff
@@ -1627,6 +1627,7 @@ def test_lv_activate_deactivate(tmp_storage):
 @requires_root
 @pytest.mark.root
 def test_lv_deactivate_in_use(tmp_storage):
+    # NOTE: This test does not work in active hosts.
```
nirs, line 1630: Lets use TODO: for these kind of comments, to make it easy to find it later.

tests/storage/lvm_test.py
```diff
@@ -1038,17 +1038,18 @@ def stale_pv(tmp_storage):
     pvs = sorted(pv.name for pv in lvm.getAllPVs())
```
nirs, line 1038: Note for the commit message:

Using --config is fine, and it does not overwrite system configuration,
it overwrites the configuration for the particular lvm command. The problem
is that we use lvm devicesfile by default, and when a system is configured
to use devicsfile you cannot use --config 'devices { filter = ...}'.

So this updates the tests so they can run on a host configured for vdsm, using
lvm devicesfile by default.

I'm not sure how this work in the container before and after the change, but we can dig deeper later.

nirs (approved):

nirs (approved):

PR #166, opened 2022-05-09
iscsi: Keep existing session on "session exists"

nirs:
If logging in to a node fails with "session exists" (error 15):

We use to remove the node, which disconnects the node and remove it.
This is not new behavior, but it seems that in 4.4 this cleanup was not
effective in the case of logging in to the same connection more than
once, and now it reliably disconnect the first node and leave the host
without any nodes, which makes it non operational.

Change iscsiadm to raise new IscsiSessionExists error, and keep the
existing session when handling this error.

With this change, if you try to connect to the same target more than
once, the host should end with one connected target.

Bug-Url: https://bugzilla.redhat.com/2083271

### DIFF
--- a/lib/vdsm/storage/iscsi.py
+++ b/lib/vdsm/storage/iscsi.py
@@ -231,6 +231,10 @@ def loginToIscsiNode(iface, target):
     log.info("Logging in to iscsi target %s via iface %s", target, iface.name)
     try:
         iscsiadm.node_login(iface.name, target.address, target.iqn)
+    except iscsiadm.IscsiSessionExists:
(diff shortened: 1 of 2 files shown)


aesteve-rh (approved):

vjuranek (approved):

nirs:
Tested in QE environment that had the same connection duplicated 3 times in connectStorageServer(). During the connection we see one "session exists" error, report
one connection in error status, but the host remain connected using one session.

Not sure why we get only one error, maybe an issue with iscsiadm when using concurrent logins.

PR #168, opened 2022-05-09
iscsi: Keep existing session on "session exists" (backport for 4.5.0)

nirs:
If logging in to a node fails with "session exists" (error 15):

We use to remove the node, which disconnects the node and remove it.
This is not new behavior, but it seems that in 4.4 this cleanup was not
effective in the case of logging in to the same connection more than
once, and now it reliably disconnect the first node and leave the host
without any nodes, which makes it non operational.

Change iscsiadm to raise new IscsiSessionExists error, and keep the
existing session when handling this error.

With this change, if you try to connect to the same target more than
once, the host should end with one connected target.

NOTE: This is a backport for testing the patch with QE automated tests.

Bug-Url: https://bugzilla.redhat.com/2083271


nirs:
Too late for 4.5.0 now, the fix will be available in 4.5.1 via #166.

PR #170, opened 2022-05-10
lvm_test: increase vg test coverage

aesteve-rh:
This branch adds a couple tests to slightly improve code coverage for VG-handling methods in the `lvm` module.

Includes a small fix for a typo in the `lvm` library.

### DIFF
--- a/tests/storage/lvm_test.py
+++ b/tests/storage/lvm_test.py
@@ -1023,6 +1023,63 @@ def test_vg_create_multiple_devices(tmp_storage):
         assert pv.vg_name == ""
 
 
+@requires_root
+@pytest.mark.root
+def test_vg_remove_by_uuid(tmp_storage):
+    dev_size = 10 * GiB
+
+    dev1 = tmp_storage.create_device(dev_size)
+    dev2 = tmp_storage.create_device(dev_size)
+    vg1_name = str(uuid.uuid4())
+    vg2_name = str(uuid.uuid4())
+
+    lvm.createVG(vg1_name, [dev1], "initial-tag", 128)
+    lvm.createVG(vg2_name, [dev2], "initial-tag", 128)
+
+    vg = lvm.getVG(vg1_name)
+
+    clear_stats()
+    lvm.removeVGbyUUID(vg.uuid)
+    check_stats(hits=0, misses=1)
+
+    # Ensure we have removed the matching VG.
+    with pytest.raises(se.VolumeGroupDoesNotExist):
+        lvm.getVGbyUUID(vg.uuid)
+
+    with pytest.raises(se.VolumeGroupDoesNotExist):
+        lvm.getVG(vg1_name)
+
+    assert len(lvm.getAllVGs()) == 1
+
+    # The other VG is still available.
+    vg2 = lvm.getVG(vg2_name)
+    assert vg2.name == vg2_name
+
+
+def test_vg_invalid_output(monkeypatch, fake_devices):
+    fake_runner = FakeRunner(out=b"Fake lvm output")
+    lc = lvm.LVMCache(fake_runner)
+    lc._stalevg = False
+
+    # Create fake devices.
+    fake_pv = make_pv(pv_name=fake_devices[0], vg_name="vg")
+    fake_vg = make_vg(pvs=[fake_pv.name], vg_name="vg")
+
+    # Assign fake PV, VG to cache.
+    lc._pvs = {fake_pv.name: fake_pv}
+    lc._vgs = {fake_vg.name: fake_vg}
+
+    monkeypatch.setattr(lvm, "_lvminfo", lc)
+
+    # TODO: is this the best way to handle this unlikely error
+    # An error with an LVM invalid output shall keep the local cache intact
+    with pytest.raises(lvm.InvalidOutputLine) as e:
+        lvm.getVG("vg-name")
+    assert "Fake lvm output" in str(e.value)
+    assert len(lc._vgs) == 1
+    assert len(lc._pvs) == 1
+
+
 @pytest.fixture
 def stale_pv(tmp_storage):
     dev_size = 1 * 1024**3
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -605,7 +605,7 @@ class LVMCache(object):
                         logutils.Head(unreadable_vgs, max_items=20))
 
                 # NOTE: vgs may return useful output even on failure, so we
-                # don't retrun here.
+                # don't return here.
 
             updatedVGs = {}
             vgsFields = {}


tests/storage/lvm_test.py
```diff
@@ -2292,7 +2342,7 @@ def make_pv(pv_name, vg_name):
 
 def make_vg(vg_name, pvs):
     return lvm.VG.fromlvm(
-        "uuid",
+        uuid.uuid4(),
```
nirs, line 2345: uuid.uuid4() is not a string. It may work by accident, but always use str(uuid.uuid4()) in the code and the tests.

Switching the code to use uuid.UUID object instead of strings can make sense but this is a huge change so it is unlikely to happen.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+def test_vg_remove_by_uuid(tmp_storage):
+    dev_size = 10 * GiB
+    num_vgs = 2
+    devs = [tmp_storage.create_device(dev_size) for _ in range(0, num_vgs)]
+    vg_names = [str(uuid.uuid4()) for _ in range(0, num_vgs)]
+
+    # TODO: should work also in read-only mode.
+    for idx, vg_name in enumerate(vg_names):
+        lvm.createVG(vg_name, [devs[idx]], f"initial-tag-{idx}", 128)
+
```
nirs, line 1037: Code above is too complicated. If we create 2 devices and 2 vgs, there to make the test more complicated than:

    dev1 = tmp_storage.create_device(dev_size)
    dev2 = tmp_storage.create_device(dev_size)

    lvm.createVG("vg-1", [dev1], f"initial-tag", 128)
    lvm.createVG("vg-2", [dev2], f"initial-tag", 128)
    
The initial tag is really the same for all vgs, no need to create unique tags.
Check how we use the intial tag in the real code.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+    clear_stats()
+    vg = lvm.getVG(vg_names[0])
+    check_stats(hits=0, misses=1)
+
+    assert vg.name == vg_names[0]
+    assert vg.pv_name == (devs[0],)
+    assert vg.tags == ("initial-tag-0",)
+    assert int(vg.extent_size) == 128 * MiB
```
nirs, line 1045: We already tested create vg elsewhere, why do we need all these asserts?

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+
+    lvm.removeVGbyUUID(vg.uuid)
+    # Ensure we have removed the matching VG and only it.
+    with pytest.raises(se.VolumeGroupDoesNotExist):
+        lvm.getVGbyUUID(vg.uuid)
```
nirs, line 1050: Lets also check that lvm.getVG(vg.name) raise the same error.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+
+    assert len(lvm.getAllVGs()) == num_vgs - 1
```
nirs, line 1052: Can simply:

    assert len(lvm.getAllVGs()) == 1

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+    for idx, vg_name in enumerate(vg_names[1:], 1):
+        clear_stats()
+        vg = lvm.getVG(vg_name)
+        check_stats(hits=1, misses=0)
+
+        assert vg.name == vg_name
+        assert vg.pv_name == (devs[idx],)
+        assert vg.tags == (f"initial-tag-{idx}",)
+        assert int(vg.extent_size) == 128 * MiB
```
nirs, line 1061: This just check that the other vg exists, no?

So:

    vg2 = lvm.getVG("vg-2")
    assert vg2.name == "vg-2"

tests/storage/lvm_test.py
aesteve-rh: Good catch. I might revert this actually, I made it because I was looking for more 'complex' scenarios with fake devices, with more fake VGs, etc. And current flow works for 1 VG with 1 PV only. But I ended up removing those tests, this change does not make sense anymore AND probably we will not ever need such complex scenarios anyway.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
 
+@requires_root
+@pytest.mark.root
+def test_vg_remove_by_uuid(tmp_storage):
+    dev_size = 10 * GiB
+    num_vgs = 2
+    devs = [tmp_storage.create_device(dev_size) for _ in range(0, num_vgs)]
+    vg_names = [str(uuid.uuid4()) for _ in range(0, num_vgs)]
+
+    # TODO: should work also in read-only mode.
```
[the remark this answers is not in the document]
aesteve-rh: The comment appears in other places, should I remove it elsewhere too or only if we revisit the code then we do the updates?

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,56 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+def test_vg_get_invalid_output(monkeypatch, fake_devices):
+    fake_runner = FakeRunner(out=b"Fake lvm output")
+    lc = lvm.LVMCache(fake_runner)
+
+    monkeypatch.setattr(lvm, "_lvminfo", lc)
+
+    with pytest.raises(lvm.InvalidOutputLine) as e:
+        lvm.getVG("vg-name")
+    assert "vgs" in e.value.command
+    assert "Fake lvm output" in e.value.line
```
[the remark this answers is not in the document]
aesteve-rh: Uhm, cache is empty, I will fill it with some data and think about what would happen to test it properly.

I'd say it shall remain unchanged, except if simulate the fake output with an error, then it might downgrade the stale VGs to `Unreadable`.

tests/storage/lvm_test.py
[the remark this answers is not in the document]
nirs: You can start by not copying it into new test. Removing it from other tests is unrelated cleanup that can be done later, or now in another PR.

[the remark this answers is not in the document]
aesteve-rh: Then should be covered already.

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,63 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+    vg = lvm.getVG(vg1_name)
+    check_stats(hits=0, misses=1)
+
+    lvm.removeVGbyUUID(vg.uuid)
+    check_stats(hits=0, misses=2)
+    # Ensure we have removed the matching VG.
+    with pytest.raises(se.VolumeGroupDoesNotExist):
+        lvm.getVGbyUUID(vg.uuid)
+        lvm.getVG(vg1_name)
+        check_stats(hits=0, misses=4)
```
nirs, line 1049: lines 1048,1049 not reachable since we raie in line 1047.
We need to have pytest.raises() block for each check.
We can remove the stats check, so we will have:

    with pytest.raises(se.VolumeGroupDoesNotExist):
        lvm.getVGbyUUID(vg.uuid)

    with pytest.raises(se.VolumeGroupDoesNotExist):
        lvm.getVG(vg.name)

tests/storage/lvm_test.py
```diff
@@ -1023,6 +1023,63 @@ def test_vg_create_multiple_devices(tmp_storage):
...
+    vg1_name = str(uuid.uuid4())
+    vg2_name = str(uuid.uuid4())
+
+    lvm.createVG(vg1_name, [dev1], "initial-tag", 128)
+    lvm.createVG(vg2_name, [dev2], "initial-tag", 128)
+
+    clear_stats()
+    vg = lvm.getVG(vg1_name)
+    check_stats(hits=0, misses=1)
+
```
nirs, line 1042: It would be useful to clear the starts here, before we call interesting call.

nirs (changes requested):

tests/storage/lvm_test.py
aesteve-rh: Yeah I wasn't sure, but it crossed my mind this may happen. Thanks! I'll fix this.

nirs (approved):

PR #171, opened 2022-05-10
net: Fix the tuple assign

almusil:
The tuple returned from split_switch_type
was not properly assigned to variables.
This caused the if statement to be always true,
it did not cause any issues, but the network capabilities
were also getting OvS info which wasn't necessary.

Assign it properly so only the first dictionary is used.

### DIFF
--- a/lib/vdsm/network/netswitch/configurator.py
+++ b/lib/vdsm/network/netswitch/configurator.py
@@ -205,7 +205,9 @@ def netinfo(vdsmnets=None, compatibility=None):
     ovs_nets, _ = util.split_switch_type(
         running_config.networks, running_config={}
     )
-    ovs_bonds = util.split_switch_type(running_config.bonds, running_config={})
+    ovs_bonds, _ = util.split_switch_type(
+        running_config.bonds, running_config={}
+    )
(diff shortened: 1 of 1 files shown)


erav (approved):

hbraha (dismissed):

lib/vdsm/network/netswitch/configurator.py
```diff
@@ -205,7 +205,9 @@ def netinfo(vdsmnets=None, compatibility=None):
     ovs_nets, _ = util.split_switch_type(
         running_config.networks, running_config={}
     )
-    ovs_bonds = util.split_switch_type(running_config.bonds, running_config={})
+    ovs_bonds, _ = util.split_switch_type(
+        running_config.bonds, running_config={}
+    )
```
nirs, line 210: Where is the typo? Is this a fix of another change? What is the issue this solves?

almusil: The tuple was not properly assigned.

nirs: Thanks, the commit message make the issue very clear now!

Issue #155, opened 2022-04-27
Host cannot connect to storage domains


ferradeira:
After upgrade from 4.4 to 4.5 host cannot be activated because cannot connect to data domain.
I have a data domain in NFS (master) and a GlusterFS. It complains about the Gluster domain:
The error message for connection node1-teste.acloud.pt:/data1 returned by VDSM was: XML error

```
# rpm -qa|grep glusterfs*
glusterfs-10.1-1.el8s.x86_64
glusterfs-selinux-2.0.1-1.el8s.noarch
glusterfs-client-xlators-10.1-1.el8s.x86_64
glusterfs-events-10.1-1.el8s.x86_64
libglusterfs0-10.1-1.el8s.x86_64
glusterfs-fuse-10.1-1.el8s.x86_64
glusterfs-server-10.1-1.el8s.x86_64
glusterfs-cli-10.1-1.el8s.x86_64
glusterfs-geo-replication-10.1-1.el8s.x86_64
```

engine log:
```
[2022-04-27 13](callto:2022-04-27 13):35:16,118+01 ERROR [org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-66) [e
be79c6] EVENT_ID: VDS_STORAGES_CONNECTION_FAILED(188), Failed to connect Host NODE1 to the Storage Domains DATA1.
[2022-04-27 13](callto:2022-04-27 13):35:16,169+01 ERROR [org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-66) [e
be79c6] EVENT_ID: STORAGE_DOMAIN_ERROR(996), The error message for connection node1-teste.acloud.pt:/data1 returned by VDSM was: XML error
[2022-04-27 13](callto:2022-04-27 13):35:16,170+01 ERROR [org.ovirt.engine.core.bll.storage.connection.FileStorageHelper] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-66) [ebe79c6
] The connection with details 'node1-teste.acloud.pt:/data1' failed because of error code '4106' and error message is: xml error
```

vdsm log:
```
[...]

PR #172, opened 2022-05-11
gluster: fix compatibility with Gluster v10 or greater,

rchikatw:
gluster: fix compatibility with Gluster v10 or greater

In gluster v10 or greater stripe count is removed and when
stripe count is not returned from the volume info command
setting the stripeCount to default value that is 1

Fixes: #155
Bug-Url: https://bugzilla.redhat.com/2078569
Signed-off-by: Ritesh Chikatwar <rchikatw@redhat.com>

### DIFF
--- a/lib/vdsm/gluster/cli.py
+++ b/lib/vdsm/gluster/cli.py
@@ -426,7 +426,7 @@ def _parseVolumeInfo(tree):
             value["volumeStatus"] = VolumeStatus.OFFLINE
         value['brickCount'] = el.find('brickCount').text
         value['distCount'] = el.find('distCount').text
-        value['stripeCount'] = el.find('stripeCount').text
+        value['stripeCount'] = el.findtext('stripeCount', '1')
         value['replicaCount'] = el.find('replicaCount').text
         value['disperseCount'] = el.find('disperseCount').text
         value['redundancyCount'] = el.find('redundancyCount').text
--- a/tests/gluster_cli_test.py
+++ b/tests/gluster_cli_test.py
@@ -164,9 +164,126 @@ class GlusterCliTests(TestCaseBase):
         volumeInfo = gcli._parseVolumeInfo(tree)
         self.assertEqual(volumeInfo, oVolumeInfo)
 
+    def _parseVolumeStripeCount_empty(self):
+        out = """<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+            <cliOutput>
+              <opRet>0</opRet>
+              <opErrno>0</opErrno>
+              <opErrstr/>
+              <volInfo>
+                <volumes>
+                  <volume>
+                    <name>music</name>
+                    <id>b3114c71-741b-4c6f-a39e-80384c4ea3cf</id>
+                    <status>1</status>
+                    <statusStr>Started</statusStr>
+                    <brickCount>2</brickCount>
+                    <distCount>2</distCount>
+                    <replicaCount>2</replicaCount>
+                    <disperseCount>0</disperseCount>
+                    <arbiterCount>1</arbiterCount>
+                    <redundancyCount>0</redundancyCount>
+                    <type>2</type>
+                    <typeStr>Replicate</typeStr>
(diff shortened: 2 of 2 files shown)


rchikatw:
@nirs Can you please review this

lib/vdsm/gluster/cli.py
```diff
@@ -426,7 +426,10 @@ def _parseVolumeInfo(tree):
             value["volumeStatus"] = VolumeStatus.OFFLINE
         value['brickCount'] = el.find('brickCount').text
         value['distCount'] = el.find('distCount').text
-        value['stripeCount'] = el.find('stripeCount').text
+        if el.find('stripeCount') is not None:
+            value['stripeCount'] = el.find('stripeCount').text
+        else:
+            value['stripeCount'] = 1
```
nirs, line 432: This adds an integer

I think a better way to do this is (did not test):

    value['stripeCount'] = el.findtext("stripeCount", "1")

https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.ElementTree.findtext

nirs (changes requested):
Thanks @rchikatw!

nirs:
Also, we are missing a test for parsing xml without stripCount. I think testing xml with stripCount is already tested.

lib/vdsm/gluster/cli.py
rchikatw: Yes make sense it works i tested.

rchikatw: Addressed

lib/vdsm/gluster/cli.py
```diff
@@ -426,7 +426,10 @@ def _parseVolumeInfo(tree):
             value["volumeStatus"] = VolumeStatus.OFFLINE
         value['brickCount'] = el.find('brickCount').text
         value['distCount'] = el.find('distCount').text
-        value['stripeCount'] = el.find('stripeCount').text
+        if el.find('stripeCount') is not None:
+            value['stripeCount'] = el.find('stripeCount').text
```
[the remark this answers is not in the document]
rchikatw: Addressed

lib/vdsm/gluster/cli.py
```diff
@@ -426,10 +426,7 @@ def _parseVolumeInfo(tree):
             value["volumeStatus"] = VolumeStatus.OFFLINE
```
aesteve-rh, line 426: This commit and the one before could be squashed, or even better, drop the other one and use this directly with the right solution.

tests/gluster_cli_test.py
```diff
@@ -164,9 +164,126 @@ def _parseVolumeInfo_test(self):
         volumeInfo = gcli._parseVolumeInfo(tree)
```
aesteve-rh, line 164: This commit is not signed-off.

lib/vdsm/gluster/cli.py
aesteve-rh: Just to explain the reasoning, we don't squash commit when merging the PR, so we try to keep the git history as clean as possible.

In this case I would squash your commits to just 2, one with the solution (your last commit as I'm commenting this), and another with the additional test. Both signed off and properly commented (in the case of the test, a bit of what is being covered by it would help a lot!).

nirs:
@rchikatw if I understand this issue correctly, this PR purpose is to fix compatibility with Gluster 10 (or 9.2?), removing "stripeCount" from the response.

Please update the commit message and PR message to reflect this. Adding None check is not a good description of the change.

lib/vdsm/gluster/cli.py
rchikatw: Actually, I feel it's better to keep test case and code changes in the same commit so I squashed my commit into 1.

aesteve-rh (approved):

gobindadas (commented):
LGTM

nirs (approved):
Thanks @rchikatw!

PR #176, opened 2022-05-12
merge: consider type when reducing a volume

aesteve-rh:
When finalizing a merge, only the format (cow) and the `optimal_size` are considered to discern when to reduce a volume. However, calculating the optimal size can be avoided in the case of a preallocated block volume, as it cannot be reduced.

Also, in the StorageDomain, `reduceVolume` can be called with no check at all, which is dangerous.

Use the volume `can_reduce` check before obtaining the optimal size, and before reducing the volume (both in the cold merge and the storage domain), so that we consider the type of volume before reducing it.

Fixes: https://github.com/oVirt/vdsm/issues/185
Bug-Url: https://bugzilla.redhat.com/2081493

### DIFF
--- a/lib/vdsm/storage/merge.py
+++ b/lib/vdsm/storage/merge.py
@@ -273,14 +273,17 @@ def finalize(subchain):
             else:
                 _finalize_internal_merge(dom, subchain)
 
-            if subchain.base_vol.chunked():
+            if subchain.base_vol.can_reduce():
                 # If the top volume is leaf, the base volume will become a leaf
                 # after the top volume is deleted.
                 optimal_size = subchain.base_vol.optimal_size(
                     as_leaf=subchain.top_vol.isLeaf())
                 actual_size = subchain.base_vol.getVolumeSize()
 
-        if subchain.base_vol.chunked() and optimal_size < actual_size:
+        # Optimal size must be computed while the image is prepared, but
+        # reducing with the volume still active will issue a warning from LVM.
+        # Thus, reduce after having teardown the volume.
+        if subchain.base_vol.can_reduce() and optimal_size < actual_size:
             _shrink_base_volume(subchain, optimal_size)
 
 
--- a/tests/storage/blocksd_test.py
+++ b/tests/storage/blocksd_test.py
@@ -1643,6 +1643,66 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
     assert vol.getLegality() == sc.ILLEGAL_VOL
 
 
+@requires_root
+def test_reduce_volume_called(domain_factory, fake_task, fake_sanlock):
+    sd_uuid = str(uuid.uuid4())
+    dom = domain_factory.create_domain(sd_uuid=sd_uuid, version=5)
+
+    img_uuid = str(uuid.uuid4())
+    vol_id = str(uuid.uuid4())
+    vol_capacity = 10 * GiB
+    initial_size = 5 * GiB
+
+    dom.createVolume(
+        imgUUID=img_uuid,
+        capacity=vol_capacity,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType=sc.DATA_DISKTYPE,
+        volUUID=vol_id,
+        desc="Over extended volume",
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID,
+        initial_size=initial_size)
+
+    # Prepare volume before reducing.
+    vol = dom.produceVolume(img_uuid, vol_id)
+    vol.prepare()
+    dom.reduceVolume(img_uuid, vol_id, allowActive=True)
+
+    # Check that (prepared) volume size has been reduced.
+    new_size = dom.getVolumeSize(img_uuid, vol_id).apparentsize
+    assert new_size == vol.optimal_size()
+
+
+@requires_root
+def test_reduce_volume_skipped(domain_factory, fake_task, fake_sanlock):
+    sd_uuid = str(uuid.uuid4())
+    dom = domain_factory.create_domain(sd_uuid=sd_uuid, version=5)
+
+    img_uuid = str(uuid.uuid4())
+    vol_id = str(uuid.uuid4())
+    vol_capacity = 10 * GiB
+
+    dom.createVolume(
+        imgUUID=img_uuid,
+        capacity=vol_capacity,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.PREALLOCATED_VOL,
+        diskType=sc.DATA_DISKTYPE,
+        volUUID=vol_id,
+        desc="Base volume",
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID)
+
+    # Volume is not prepared but it does fail since we skip the reduce
+    # for preallocated volume.
+    dom.reduceVolume(img_uuid, vol_id)
+
+    # Check that volume size has not changed.
+    assert dom.getVolumeSize(img_uuid, vol_id).apparentsize == vol_capacity
+
+
 LVM_TAG_CHARS = string.ascii_letters + "0123456789_+.-/=!:#"
 
 LVM_TAGS = [
--- a/tests/storage/merge_test.py
+++ b/tests/storage/merge_test.py
@@ -362,7 +362,12 @@ class TestFinalizeMerge:
 
     # TODO: use one make_env for all tests?
     @contextmanager
-    def make_env(self, sd_type='block', format='raw', chain_len=2):
+    def make_env(
+            self,
+            sd_type='block',
+            format='raw',
+            prealloc=sc.SPARSE_VOL,
+            chain_len=2):
         size = 2 * GiB
         base_fmt = sc.name2type(format)
         with fake_env(sd_type) as env:
@@ -372,7 +377,8 @@ class TestFinalizeMerge:
                 mp.setattr(blockVolume, 'rm', FakeResourceManager())
                 mp.setattr(image, 'Image', FakeImage)
 
-                env.chain = make_qemu_chain(env, size, base_fmt, chain_len)
+                env.chain = make_qemu_chain(
+                    env, size, base_fmt, chain_len, prealloc=prealloc)
 
                 volumes = {(vol.imgUUID, vol.volUUID): FakeVolume()
                            for vol in env.chain}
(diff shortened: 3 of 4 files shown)


lib/vdsm/storage/merge.py
```diff
@@ -266,7 +266,8 @@ def finalize(subchain):
                 # optimal_size must be called when the volume is prepared
```
nirs, line 266: We need to check if a volume can be reduced also in line 265. If a volume cannot be reduced, there is no need to get the optimal size.

tests/storage/merge_test.py
```diff
@@ -305,14 +305,19 @@ class TestFinalizeMerge:
     def make_env(self, sd_type='block', format='raw', chain_len=2):
         size = MiB
         base_fmt = sc.name2type(format)
+        if sd_type == 'block' and format == 'raw':
+            prealloc = sc.PREALLOCATED_VOL
+        else:
+            prealloc = sc.SPARSE_VOL
```
nirs, line 311: Don't add logic here, the test should configure this properly to test all cases.

tests/storage/merge_test.py
```diff
@@ -455,8 +460,12 @@ def test_reduce_chunked(self):
                 ('reduce', (base_vol.optimal_size(),), {}),
             ]
 
-    def test_reduce_not_chunked(self):
-        with self.make_env(sd_type='file', format='cow', chain_len=4) as env:
+    @pytest.mark.parametrize("sd_type, format", [
+        pytest.param('file', 'cow'),  # Not chunked, reduce not called
+        pytest.param('block', 'raw'),  # Preallocated, reduce not called
+    ])
```
nirs, line 466: This does not test the interesting cases - the issue is with preallocated qcow2 on block, and this tests preallocated raw.

nirs (changes requested):
This conflicts with #133. Lets wait until it is merged. I plan to work on this next week.

lib/vdsm/storage/merge.py
aesteve-rh: OK.

Why is the optimal size obtained within the context and the shrink called outside the context?

lib/vdsm/storage/merge.py
```diff
@@ -266,7 +266,8 @@ def finalize(subchain):
...
                 optimal_size = subchain.base_vol.optimal_size()
 
-        if subchain.base_vol.chunked():
+        if (subchain.base_vol.chunked()
+                and subchain.base_vol.getType() == sc.SPARSE_VOL):
```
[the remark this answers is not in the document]
aesteve-rh: First point is tackled here: https://github.com/oVirt/vdsm/pull/178

For the rest, I will wait until #133 :)

lib/vdsm/storage/merge.py
nirs: To compute optimal size we measure the image using "qemu-img measure" - so we must prepare the image (activate the logical volume in block storage). Reducing the image is possible when the logical volume is active, but LVM issues warnings about this, so we prefer to teardown the volume before reducing.

The flow is:
- activate the lv and its parent lv (recursively)
- measure the qocw2 image in the lv
- deactivate the lv and its parent (recursively)
- if the lv is bigger than the optimal size, reduce it.

I guess the code needs a comment explaining this.

aesteve-rh: Great, thanks for the explanation :)
I assumed something like that from the code, but a clear explanation is much better! I may add a comment to explain that in the code.

tests/storage/merge_test.py
```diff
@@ -538,8 +544,18 @@ def test_reduce_chunked_leaf(self):
...
 
-    def test_reduce_not_chunked(self):
-        with self.make_env(sd_type='file', format='cow', chain_len=4) as env:
+    @pytest.mark.parametrize("sd_type, format, prealloc", [
+        # Not chunked, reduce not called
+        pytest.param('file', 'cow', sc.SPARSE_VOL),
+        # Preallocated, reduce not called
+        pytest.param('block', 'cow', sc.PREALLOCATED_VOL),
+    ])
+    def test_reduce_not_called(self, sd_type, format, prealloc):
```
nirs, line 553: Maybe test_reduce_skipped?

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,63 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+        capacity=vol_capacity,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType=sc.DATA_DISKTYPE,
+        volUUID=vol_id,
+        desc="Base volume",
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID)
+
+    # Prepare volume before reducing
```
nirs, line 1666: Keeping . at the end of comments will make this even nicer.

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,63 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.PREALLOCATED_VOL,
+        diskType=sc.DATA_DISKTYPE,
+        volUUID=vol_id,
+        desc="Base volume",
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID)
+
+    # Volume is not prepared, but should not fail as long as reduce
+    # is not executed for preallocated volumes
```
nirs, line 1695: I would change this to something like:

    # Volume is not prepared but it does fail since we skip the reduce
    # for preallocated volume.

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,63 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+    dom.reduceVolume(img_uuid, vol_id)
+
+    # TODO: is there a better way to check that reduce has not been executed?
+    # Check that (prepared) volume size has not changed
```
nirs, line 1699: The TODO: is not needed - we don't want to check that reduce was not called, but that reducing a preallocated volume does not change the volume size, and this check is the
best way to do this.

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,63 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+    vol_capacity = 10 * GiB
+
+    dom.createVolume(
+        imgUUID=img_uuid,
+        capacity=vol_capacity,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType=sc.DATA_DISKTYPE,
+        volUUID=vol_id,
+        desc="Base volume",
```
nirs, line 1662: "over extended volume"

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,63 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID)
+
+    # Prepare volume before reducing
+    dom.produceVolume(img_uuid, vol_id).prepare()
+    dom.reduceVolume(img_uuid, vol_id)
+
+    # Check that (prepared) volume size has been reduced
+    assert dom.getVolumeSize(img_uuid, vol_id).truesize < vol_capacity
```
nirs, line 1671: This does not test that we reduced the volume to the optimal size.

To fix this test:
1. Create the volume with initial_size=5 * GiB
2. Check that the volume size after the reduce is vol.optimal_size()

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,63 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID)
+
+    # Volume is not prepared, but should not fail as long as reduce
+    # is not executed for preallocated volumes
+    dom.reduceVolume(img_uuid, vol_id)
+
+    # TODO: is there a better way to check that reduce has not been executed?
+    # Check that (prepared) volume size has not changed
+    assert dom.getVolumeSize(img_uuid, vol_id).truesize == vol_capacity
```
nirs, line 1700: using apparentsize is better. truesize is kind of not working for block storage, the value is always same as apparentsize. In file storage this is the size allocated
(like du /path).

nirs (commented):
Nice!

nirs:
This fixes 2 bugs:
- reduce after cold merge - https://bugzilla.redhat.com/2081493
- reduce when calling Volume.reduce - we don't have a bug

It will be useful to add upstream issue to the second fix, since users may hit this issue
and the issue will help them to understand the problem and find the fix.

You should be able to reproduce this issue using the reduce example in the sdk:
https://github.com/oVirt/python-ovirt-engine-sdk4/blob/main/examples/reduce_disk.py

Create preallocated volume enabled for incremental backup. The disk will be created fully allocated. Run the reduce script - without the fix it will be reduced to 2.62g. With the fix it will not be reduced.

tests/storage/blocksd_test.py
aesteve-rh: Agh, I try because I see all comments with it and want to keep it consistent, but I'm not used to have periods at the end of sentences in comments, so I miss them easily. I'll fix them.

tests/storage/blocksd_test.py
aesteve-rh: I was not sure since in merge_tests the call is actually checked using the `__calls__` attribute, using fake volumes, and hence the comment. I will remove it.

tests/storage/blocksd_test.py
```diff
@@ -1643,6 +1643,66 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
...
+        initial_size=initial_size)
+
+    # Prepare volume before reducing.
+    vol = dom.produceVolume(img_uuid, vol_id)
+    vol.prepare()
+    dom.reduceVolume(img_uuid, vol_id, allowActive=True)
+
+    # Check that (prepared) volume size has been reduced.
+    assert dom.getVolumeSize(
+        img_uuid, vol_id).apparentsize == vol.optimal_size()
```
nirs, line 1675: This is valid python, but harder to read. It will be nicer as:

    new_size = dom.getVolumeSize(img_uuid, vol_id).apparentsize
    assert new_size == vol.optimal_size()

nirs (dismissed):

nirs (approved):

Issue #56, opened 2022-01-30
Remove vdsm.storage.compat module


nirs:
This module added to deal with missing sanlock and ioprocess
in python 3, and with older versions of sanlock that use 
async= instead of wait=.

These issues are not relevant for a while, so the module can be removed.

Blocks #55

PR #177, opened 2022-05-13
storage.compat: remove module

aesteve-rh:
This module used to handle missing ioprocess
and sanlock libraries in Python 3, and a workaround
to support sanlock older versions, that used 'async='
instead of to 'wait='.

However, it is not relevant anymore.

Fixes: #56


lib/vdsm/storage/outOfProcess.py
```diff
@@ -30,6 +30,7 @@
 import weakref
 
 from functools import partial
```
nirs, line 32: We separate standard library moudules and 3rd party modules with blank line. You change remove the line separating "import six" and "from functools import partial".

lib/vdsm/tool/check_volume_leases.py
```diff
@@ -23,11 +23,11 @@
 
 from collections import defaultdict
 import argparse
```
nirs, line 25: Missing blank line between argparse (stdlib) and sanlock (3rd party).

tests/storage/clusterlock_test.py
```diff
@@ -24,13 +24,14 @@
 import json
 import time
```
nirs, line 25: While you touch this file, it will be useful to add the missing blank line before pytest.

tests/storage/fakesanlock.py
```diff
@@ -26,12 +26,13 @@
 import socket
 import threading
 
+import sanlock
+
 from operator import itemgetter
```
nirs, line 31: This is standard libray import, sanlock should come after it.

nirs (commented):
Looks good except import order and separation of standard library and 3rd party modules.

vjuranek (approved):

PR #178, opened 2022-05-13
volume: add 'can_reduce' method

aesteve-rh:
Currently there is not check performed from within
the volume class before a reduce is attempted. This may
result in reduce calls with wrong volume type (i.e., preallocated).

In preparation, provide a 'can_reduce' method in
the volume base class, so that it could be called from the
appropiate place before a volume is reduced.


vjuranek (approved):

tests/storage/blockvolume_test.py
```diff
@@ -196,3 +199,14 @@ def test_repair_capacity(self, capacity, virtual_size, expected_capacity):
 
             vol.updateInvalidatedSize()
             assert vol.getMetadata().capacity == expected_capacity
+
+    @permutations([
+        # format, type (prealloc), expected_result
```
nirs, line 204: Replacing expected_result with can_reduce can make this more clear.

tests/storage/blockvolume_test.py
```diff
@@ -196,3 +199,14 @@ def test_repair_capacity(self, capacity, virtual_size, expected_capacity):
...
+        (sc.RAW_FORMAT, sc.PREALLOCATED_VOL, False),  # invalid format and type
+        (sc.COW_FORMAT, sc.PREALLOCATED_VOL, False),  # invalid type
+        (sc.RAW_FORMAT, sc.SPARSE_VOL, False),  # invalid format
+        (sc.COW_FORMAT, sc.SPARSE_VOL, True),  # all OK
+    ])
+    def test_can_reduce(self, format, type, expected_result):
+        with self.make_volume(size=GiB, format=format, prealloc=type) as vol:
```
nirs, line 211: prealloc=type is very strange - if we call the argument prealloc=, we want to call the parameter prealloc.

tests/storage/blockvolume_test.py
```diff
@@ -112,17 +112,20 @@ def test_fail_invalid_block_volume_size(self, preallocate, vol_format):
 class TestBlockVolumeManifest(VdsmTestCase):
 
     @contextmanager
-    def make_volume(self, size, storage_type='block', format=sc.RAW_FORMAT):
+    def make_volume(self, size, format=sc.RAW_FORMAT, prealloc=sc.SPARSE_VOL):
+        storage_type = 'block'
```
nirs, line 116: Good change - but since we really used this once, we can simply inline "block" in the single line using this.

tests/storage/blockvolume_test.py
```diff
@@ -112,17 +112,20 @@ def test_fail_invalid_block_volume_size(self, preallocate, vol_format):
...
         img_id = make_uuid()
         vol_id = make_uuid()
         # TODO fix make_volume helper to create the qcow image when needed
         with fake_env(storage_type) as env:
             if format == sc.RAW_FORMAT:
-                env.make_volume(size, img_id, vol_id, vol_format=format)
+                env.make_volume(size, img_id, vol_id,
+                                vol_format=format, prealloc=prealloc)
```
nirs, line 123: We prefer this style when arguments do not fit in one line:

    env.make_volume(
        size, img_id, vol_id, vol_format=format, prealloc=prealloc)

If argument do not fit this way, use one line per argument:

    env.make_volume(
        size,
        img_id,
        vol_id,
        vol_format=format,
        prealloc=prealloc)

nirs (commented):
Nice!

lib/vdsm/storage/blockVolume.py
```diff
@@ -70,6 +70,9 @@ def is_block(cls):
     def chunked(self):
         return self.getFormat() == sc.COW_FORMAT
 
+    def can_reduce(self):
+        return self.chunked() and self.getType() == sc.SPARSE_VOL
```
nirs, line 74: This is fine for cold merge, where we use BlockVolumeManifest object. But in Volume.reduce, we use:
https://github.com/oVirt/vdsm/blob/bec5e7c94d25d02cadb3e9c9a481a6963151211c/lib/vdsm/storage/blockSD.py#L1600

Which uses a BlockVolume object (using a BlockVolumeManifest).

Why do we have both BlockVolume and BlockVolumeManifest? this is a result of cancelled project to remove the SPM in ovirt 4.0. The plan was to extract BlockVolumeManifest that will be used for new code path in ovirt 4.0, and remove BlockVolume in later release. Sadly this project was canceled in the middle, leaving us with worse code.

So what we have do now, is create wrappers for code that need to be accessed from both BlockVolume and BlockVolumeManifast. Most of the wrappers are in sd.py on the StorageDomain super class. See StorageDomain.supports_bitmap_operations() for example wrapper.

We have very bad tests for the wrapers in sdm_indirection_test.py, don't waste your time adding tests there, we need to remove these tests.

tests/storage/blockvolume_test.py
```diff
@@ -196,3 +198,15 @@ def test_repair_capacity(self, capacity, virtual_size, expected_capacity):
 
             vol.updateInvalidatedSize()
             assert vol.getMetadata().capacity == expected_capacity
+
+    @permutations([
+        # format, prealloc, can_reduce
+        (sc.RAW_FORMAT, sc.PREALLOCATED_VOL, False),  # invalid format and type
+        (sc.COW_FORMAT, sc.PREALLOCATED_VOL, False),  # invalid type
+        (sc.RAW_FORMAT, sc.SPARSE_VOL, False),  # invalid format
+        (sc.COW_FORMAT, sc.SPARSE_VOL, True),  # all OK
```
nirs, line 207: I forgot to submit this comment last time - we don't test anything invalid here, all
arguments are valid, so the comments are misleading.

We can describe the arguments better like this:

        # Raw or preallocated disks cannot be reduced.
        (sc.RAW_FORMAT, sc.PREALLOCATED_VOL, False),
        (sc.COW_FORMAT, sc.PREALLOCATED_VOL, False),
        # Cow sparse can be reduced.
        (sc.COW_FORMAT, sc.SPARSE_VOL, True),

We don't need to test:

        (sc.RAW_FORMAT, sc.SPARSE_VOL, False),

Because this is an invalid combination, we cannot have such disks. You can add a comment about this if you think this is not clear enough. It may help new contributors.

nirs (changes requested):

lib/vdsm/storage/blockVolume.py
aesteve-rh: I hope I got it right. Not sure if you were actually suggesting me to make the wrapper in the `StorageDomain`, or just take it as an example. But to me, it looked better fit to make the wrapper in the `Volume` class.

Made it in a separate commit to make it easier to spot and fix if needed :)

lib/vdsm/storage/volume.py
```diff
@@ -1634,6 +1637,7 @@ class VolumeLease(guarded.AbstractLock):
     """
     Extend AbstractLock so Volume Leases may be used with guarded utilities.
     """
+
```
nirs, line 1640: Unrelated change, best to avoid such changes.

If you want to unify spacing in the entire code base do this separately.

nirs (commented):
Wrapper looks good.

lib/vdsm/storage/volume.py
aesteve-rh: `autopep8` trolled me a bit on this one, I will be more careful with it, sorry.

PR #180, opened 2022-05-15
Fix reduce after active layer cold merge

### LINKED ISSUE #179: Merge reduces base volume too much when finalizing
When a cold merge is finalized, we reduce the base volume to optimal size. However
the optimal size calculation is wrong for active merge, since the top volume is
removed after finalizing merge. During finalize merge, the top volume still exist,
so the base volume is not leaf. But after the top volume is deleted, the base 
volume becomes the leaf, so it needs more space.

            if subchain.base_vol.chunked():
                # optimal_size must be called when the volume is prepared
                optimal_size = subchain.base_vol.optimal_size()
                actual_size = subchain.base_vol.getVolumeSize()

        if subchain.base_vol.chunked() and optimal_size < actual_size:
            _shrink_base_volume(subchain, optimal_size)

This happens only in active merge. For internal merge the base volume does not
become the leaf after the merge, so the current code is correct.

When starting the VM after cold merge, the disk does not have enough free space,
and the guest is more likely to pause when writing to the disk. Usually vdsm
detects that the disk needs extension when the vm is started and extend the disk
before the guest can write any data. Here is an example log:

```
2022-05-15 22:36:57,091+0300 INFO  (vm/6e3be0cf) [virt.vm] (vmId='6e3be0cf-2d12-4bb6-9409-f6b8e1bf49e3') CPU running: domain i
nitialization (vm:6019)
...
2022-05-15 22:36:57,526+0300 INFO  (periodic/1) [virt.vm] (vmId='6e3be0cf-2d12-4bb6-9409-f6b8e1bf49e3') drive 'sdb' needs exte
nsion (storage:979)
2022-05-15 22:36:57,526+0300 INFO  (periodic/1) [virt.vm] (vmId='6e3be0cf-2d12-4bb6-9409-f6b8e1bf49e3') Requesting an extensio
n for volume e2115e7b-0d88-47df-a8ee-1acb7db6f1d9 on domain bafd0f16-9aba-4f9f-ba90-46d3b8a29157 block_info BlockInfo(index=1,
 name='sdb', path='/rhev/data-center/mnt/blockSD/bafd0f16-9aba-4f9f-ba90-46d3b8a29157/images/e6d71bf6-80cf-407c-b7d0-2083a40e7
584/e2115e7b-0d88-47df-a8ee-1acb7db6f1d9', allocation=0, capacity=10737418240, physical=134217728, threshold=0) (thinp:388)
```

## How to reproduce

1. Create vm with 10g empty thin disk (creates 2.5g empty volume)
2. Create snapshot including only think disk (creates new 2.5g snapshot)
3. Delete the snapshot
4. When snapshot is deleted, check volume size using lvs:

       lvs --devicesfile= -o vg_name,lv_name,tags,size | grep disk-id

Actual result:
Volume size after the operation is 1.0 GiB
(Minimum disk size for cold merge in oVirt < 4.5.1)

Expected result:
Volume size after the operation is 2.62 GiB

## Calculating the expected size

The expected size is actual size + 2.5 GiB free space. Since the image is empty,
it actual size is 262144 bytes. You can check using:

    # qemu-img check /dev/vg-name/lv-name  
    Image end offset: 262144

2.5 GiB + 262144 bytes is 2684616704 bytes. Value need to be aligned
to VG extent size (128 MiB):

    >>> 2684616704 / (128 * 1024**2)
    20.001953125
    
So we need 21 extents:

    >>> 21 * 128 * 1024**2
    2818572288

Which is 2.62 GiB.

nirs:
When computing optimal size during finalize, if the top volume is a leaf
(active layer merge), the base volume will become leaf after the merge.
Compute the optimal size as if the base volume is a leaf.

Fixes #179

### DIFF
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -437,9 +437,14 @@ class BlockVolumeManifest(volume.VolumeManifest):
 
         return optimal_size
 
-    def optimal_size(self):
+    def optimal_size(self, as_leaf=False):
         """
         Return the optimal size of the volume, based on actual allocation.
+
+        If as_leaf is True, calculate the optimal size as if this volume is a
+        leaf. This is used when calculating the optimal size for the base
+        volume after a merge, before the top volume is deleted and the base
+        volume becomes the leaf.
 
         NOTE: The volume must be prepared so we can check the actual
         allocation.
@@ -451,7 +456,9 @@ class BlockVolumeManifest(volume.VolumeManifest):
         else:
             check = qemuimg.check(self.getVolumePath(), qemuimg.FORMAT.QCOW2)
             return self.optimal_cow_size(
-                check['offset'], self.getCapacity(), self.isLeaf())
+                check['offset'],
+                self.getCapacity(),
+                self.isLeaf() or as_leaf)
 
 
 class BlockVolume(volume.Volume):
--- a/lib/vdsm/storage/merge.py
+++ b/lib/vdsm/storage/merge.py
@@ -274,8 +274,10 @@ def finalize(subchain):
                 _finalize_internal_merge(dom, subchain)
 
             if subchain.base_vol.chunked():
-                # optimal_size must be called when the volume is prepared
-                optimal_size = subchain.base_vol.optimal_size()
+                # If the top volume is leaf, the base volume will become a leaf
+                # after the top volume is deleted.
+                optimal_size = subchain.base_vol.optimal_size(
+                    as_leaf=subchain.top_vol.isLeaf())
                 actual_size = subchain.base_vol.getVolumeSize()
 
         if subchain.base_vol.chunked() and optimal_size < actual_size:
--- a/tests/storage/blockvolume_test.py
+++ b/tests/storage/blockvolume_test.py
@@ -194,6 +194,17 @@ class TestBlockVolumeManifest(VdsmTestCase):
                         actual_size, virtual_size, vol.isLeaf()),
                     optimal_size)
 
+    @MonkeyPatch(blockVolume, 'config', CONFIG)
+    def test_optimal_size_cow_internal_as_leaf(self):
+        def fake_check(path, format):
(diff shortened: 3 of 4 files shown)


lib/vdsm/storage/blockVolume.py
```diff
@@ -434,10 +434,15 @@ def optimal_cow_size(cls, required_size, capacity, is_leaf):
 
         return optimal_size
 
-    def optimal_size(self):
+    def optimal_size(self, as_leaf=False):
         """
         Return the optimal size of the volume, based on actual allocation.
 
+        if as_leaf is True, calculate the optimal size as if this volume is a
```
vjuranek, line 441: Capital "I" in "If"

lib/vdsm/storage/blockVolume.py
```diff
@@ -434,10 +434,15 @@ def optimal_cow_size(cls, required_size, capacity, is_leaf):
...
+        leaf. This is used when calculating the optimal for the base volume
```
vjuranek, line 442: "optimal" -> "optimal size"

lib/vdsm/storage/blockVolume.py
```diff
@@ -428,6 +428,10 @@ def optimal_size(self):
             # minimal volume size.
             optimal_size = max(actual_size + MIN_PADDING, sc.MIN_CHUNK)
 
+        # Align to align size so callers can compare optimal size with the
```
aesteve-rh, line 423: This reads a bit weird, maybe add single quotes around 'align size' or add the underscore to mimic the actual attribute name would make it clearer.

nirs: You reviewed a patch which was part of #133, which is already merged now. I will post a cleanup for this comment separately since this change is not part of this PR.

aesteve-rh (approved):

aesteve-rh (approved):

vjuranek (approved):

nirs:
@vjuranek thanks for reviewing!

PR #183, opened 2022-05-17
vdsm-dev: helper script first drop

aesteve-rh:
Simple helper shell script to handle common vdsm operations
as a developer.
Some further **testing** and **feedback** is more than welcome (and probably **advised**).

The script handles:
- `venv` creation and activation before running tests (no more venv activate!)
- storage creation when needed
- configure, build and upgrade packages in a single command
- running tests
- running tests as root
- shows coverage summaries

### DIFF
--- a/contrib/vdsm-dev
+++ b/contrib/vdsm-dev
@@ -1,0 +1,204 @@
+#!/usr/bin/env bash
+# Author: Albert Esteve <aesteve@redhat.com>
+
+venv_path=$HOME/.venv/vdsm
+venv_activate_path=${venv_path}/bin/activate
+user_storage_path=/var/tmp/vdsm-storage/
+
+print_help()
+{
+    echo "oVirt/vdsm command line helper."
+    echo "Should be run from the repository root folder."
+    echo "Usage: $0 CMD [options]"
+    echo
+    echo "Available commands:"
+    echo "help:          Print this message and exits."
+    echo "build-podman:  Build RPMs in a container."
+    echo "                 Receives an container image as first argument [default=quay.io/ovirt/vdsm-test-centos-8]"
+    echo "build-upgrade: Build vdsm, make RPMs, and update local packages."
+    echo "test:          Run any test from specified environment using tox."
+    echo "root-test:     Run one test as root from specified environment using tox."
+    echo "                 Receives an environment with -e env [default=storage],"
+    echo "                 followed by the test path. [mandatory]"
+    echo "test-storage:  Run one/all storage tests using tox."
+    echo "coverage:      Show coverage results generated after a test run."
+    echo "                 Receives an environment with '-e env'. [default=storage]"
+    echo "Examples:"
+    echo "Run a storage test"
+    echo "  $ $0 test-storage storage/lvm_test.py"
+    echo "Run all virt and storage tests"
+    echo "  $ $0 test -e virt,storage"
+    echo "Run a test as root"
+    echo "  $ $0 root-test [-e storage] storage/storageserver_test.py"
+    echo "Show network coverage"
+    echo "  $ $0 coverage -e network"
+}
+
+
+check_storage_env()
+{
+    while test $# -gt 0
+    do
+        case "$1" in
+            -e)
+                shift
+                env=$1
+                if [ $env == *"storage"* ]; then
+                    make storage
+                fi
+                ;;
+            *)
+                # Let tox handle the rest of the arguments
+                ;;
+        esac
+        shift
+    done
+}
+
+
+if [ -z "$1" ]; then
+    echo "Command missing"
+    print_help
+    exit 1
+fi
+
+case $1 in
+
+    help )
+
+        print_help
+
+        ;;
+
+    build-podman )
+
+        shift
+        image=quay.io/ovirt/vdsm-test-centos-8
+        if [ ! -z "$1" ]; then
+            image="$1"
+        fi
+
+        podman run \
+            --env-host \
+            --privileged \
+            --rm \
+            -it \
+            --volume `pwd`:/vdsm:Z \
+            "$image" \
+            bash -c "cd /vdsm && ./autogen.sh --system --enable-timestamp && make rpm"
+
+        ;;
+
+    build-upgrade )
+
+        ./autogen.sh --system --enable-timestamp && make all rpm upgrade
+
+        ;;
+
+    test )
+
+        if [ ! -f $venv_activate_path ]; then
+            make venv
+        fi
+
+        # Loop all params passed to the script and check if storage is needed
+        check_storage_env "$@"
+
+        source ${venv_activate_path}
+        echo "Running: tox ${@:2}"
+        tox "${@:2}"
+
+        ;;
+
+    root-test )
+
+        if [ ! -f $venv_activate_path ]; then
+            make venv
+        fi
+
+        # Loop all params passed to the script and check if storage is needed
+        check_storage_env "$@"
+
+        shift
+        env="storage"
+        if [ "$1" == "-e" ]; then
+            shift
+            env="$1"
+            test="$2"
+        else
+            test="$1"
+        fi
+
+        if [ -z $test ]; then
+            echo "Missing mandatory test argument"
+            exit 1
+        fi
+
+        source ${venv_activate_path}
+        echo "Running: sudo ${venv_path}/bin/tox -e {$env} --workdir .root-tox  -- -vv $test"
+        echo "Coverage check skipped"
+        echo "Coverage report created at ./tests/htmlcov-root-$env"
+        sudo ${venv_path}/bin/tox -e {$env} --workdir .root-tox  -- -vv $test --cov-report=html:htmlcov-root-$env  --cov-fail-under 0
+
+        ;;
+
+    test-storage )
+
+        shift
+        test=$1
+       
+        if [ ! -f $venv_activate_path ]; then
+            make venv
+        fi
+
+        make storage
+
+        source ${venv_activate_path}
+        if [[ $test == "" ]]; then
+            echo "Running: tox -e storage"
+            tox -e storage
+        else
+            echo "Running: tox -e storage -- -vv $test ${@:2}"
+            echo "Coverage check skipped"
+            tox -e storage -- -vv $test "${@:2}" --cov-fail-under 0
+        fi
+
+        ;;
+
+    coverage )
+
+        env="storage"
+        while test $# -gt 0
+        do
+            case "$1" in
+                -e)
+                    shift
+                    env=$1
+                    ;;
+                *)
+                    echo "Unrecognized parameter: $1"
+                    ;;
+            esac
+            shift
+        done
+
+        cov_path="tests/htmlcov-$env/index.html"
+
+        if [ ! -f $cov_path ]; then
+            echo "No coverage summary found at $cov_path"
+            exit 1
+        fi
+
+        xdg-open $cov_path
+
+        ;;
+
+    * )
+
+        echo "Unknwon option: $1"
+        print_help
+        exit 1
+
+        ;;
+
+esac


contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env bash
```
nirs, line 1: vdsm-dev is good name for your ~/bin directory, but as part of vdsm we don't need
the "vdsm-" prefix.

Is this tool meant to be used from the contrib directory or copied
to ~/bin?

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+            exit 1
+        fi
+    fi
+}
+
+vdsm_build()
+{
+    git clean -dxf && ./autogen.sh --system --enable-timestamp && \
+         make && rm -rf ~/rpmbuild && make rpm
+}
```
nirs, line 59: This would be very useful as makefile target - just need to find a good name for this.

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+        done
+
+        cov_path="tests/htmlcov-$env/index.html"
+
+        if [ ! -f $cov_path ]; then
+            echo "No coverage summary found at $cov_path"
+            exit 1
+        fi
+
+        xdg-open $cov_path
```
nirs, line 175: This can also be nice as make target, maybe:

    make storage-coverage

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+
+    help )
+
+        print_help
+
+        ;;
+
+    build )
+
+        vdsm_build
```
mz-pdm, line 98: I would also suggest offering an option to build it in a podman container.  While the whole script can be run there, it's not obvious, according to my recent experience, for the general audience how to make it all properly.  A possibility to build Vdsm rpm's to a local ~/rpmbuild directory with a single command in any environment would make things much easier for those who need to run a particular version.  Also, running the build in a container would ensure the build is made in a clean and any environment.

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+        if [ ! -f "tests/$path" ]; then
+            echo "No test $path found"
+            exit 1
+        fi
+    fi
+}
+
+vdsm_build()
+{
+    git clean -dxf && ./autogen.sh --system --enable-timestamp && \
```
mz-pdm, line 57: `git clean -dxf` is dangerous to do without proper warnings.  Maybe better to run something like  `make clean` after running autogen.sh (I do it in my builds)?

mz-pdm (commented):
Thank you, nice to have such a thing.

I agree with Nir on moving all the parts that can be moved there to make targets.

contrib/vdsm-dev
aesteve-rh: I have it in my bin folder, but it will work the same here.

If we get to the point where we want to integrate this, we can think of a better name. Maybe `dev-run` or something like that? Anyway, let's first see if this script has interest or nah.

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+    deactivate
+}
+
+make_storage()
+{
+    if [ ! -d $user_storage_path ]; then
+        echo "Make storage at $user_storage_path"
+        make storage
+    fi
+}
```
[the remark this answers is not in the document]
aesteve-rh: Nah, I believe it is already correctly handled by the `userstorage` script. I'll remove the function and just call `make storage` every time we run a storage test.

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+    path=$1
+    storage_prefix=storage/
+    if [ ! -f "tests/$path" ]; then
+        path=${storage_prefix}${path}
+        if [ ! -f "tests/$path" ]; then
+            echo "No test $path found"
+            exit 1
+        fi
+    fi
+}
```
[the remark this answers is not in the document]
aesteve-rh: This is... something I thought it may be nice to have but then was too complex for what I was using it, so I will probably drop it.

The idea was to check if the test file exists, and if not, look for it and give the correct path to `tox`. And if it doesn't, just show an error before executing the test. Maybe with a `find` command, this could be achieved nicely.

This way one could do:
```
vdsm-dev test-storage lvm.py
```
And get the correct test without having to worry about the exact location of the file. But the gain is relatively small and we have to be careful in case the same test filename exists in two environments. So again, probably will drop it.

contrib/vdsm-dev
aesteve-rh: `make build-all`?

nirs: I'm always running this like this:

    git clean -dxf && ./autogen.sh --system --enable-timestamp && make && rm -rf ~/rpmbuild && make rpm && (cd ~/rpmbuild/RPMS && sudo dnf upgrade */*.rpm)

with this can use:

    make deploy

contrib/vdsm-dev
aesteve-rh: That sounds like a neat idea.

contrib/vdsm-dev
aesteve-rh: If it is only for storage, yes, it might be nice to have a make target for that.

In this script, it works nicely for any environment. Doing the same in a Makefile is not possible (well, it is by hacking the makefile and receiving envvars), or it would result in a bunch of targets for all test environments.

contrib/vdsm-dev
nirs: git clean -dxf is dangerous, but it helps to teach user good practice by deleting files kept in the repo without adding them to git. Unfortunately without deep cleaning, sometimes the build is not correct.

The most annoying issue is when you have missing tag, and --enable-timestamp does not crate a build with a timestamp. You pull the missing tag but without cleaning the next build is wrong again.

This probably means that make clean is not strong enough, we may need to remove more files.

aesteve-rh: I'd say is documented exactly like this in the READMEs. But I can change it if you know better options, or we can prompt the user for confirmation before executing.

mz-pdm: Having `git clean -dxf` in README is OK because the user must use it by hand and knows what's happening. It's different in a script. If any cleanup needed for a proper build is missing in `make clean` then it should be added there, rather than removing files created by the user (which can be anything -- notes, auxiliary scripts, testing data, ...).

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+    echo "help:          Print this message and exits."
+    echo "build:         Build vdsm and make RPMs."
+    echo "build-install: Build vdsm, make RPMs, and update local packages."
+    echo "test:          Run any test from specified environment using tox."
+    echo "test-storage:  Run one/all storage tests using tox."
+    echo "show-coverage: Show coverage results generated after a test run."
+    echo "               Receives an environment with '-e env'. [default=storage]"
+}
+
+venv_create()
```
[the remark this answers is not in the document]
aesteve-rh: https://github.com/oVirt/vdsm/pull/187

contrib/vdsm-dev
nirs: For coverage, it is useful to show coverage per tox env, or
combined coverage for the entire code. We already have code
to generate the combined coverage. I'm not sure how important
is the coverage per tox env, make just add

   make coverage

Creating the combined coverage and opening the index.html file in a browser.

We are also missing a way to publish coverage report from github builds.
I think the best way would be to integrate with services like [coveralls](https://coveralls.io/).

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+                    make_storage
+                fi
+                ;;
+            *)
+                # Let tox handle the rest of the arguments
+                ;;
+        esac
+        shift
+    done
+}
```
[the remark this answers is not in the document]
aesteve-rh: Since it does shift inside and all, I think it is actually easier to handle if it is kept separately.

But I have added some comment to clarify the purpose and renamed the function.

contrib/vdsm-dev
aesteve-rh: Removed the clean completely derived from the discussion in https://github.com/oVirt/vdsm/pull/242.

contrib/vdsm-dev
```diff
@@ -0,0 +1,187 @@
...
+        make_storage
+
+        source ${venv_activate_path}
+        if [[ $test == "" ]]; then
+            echo "Running: tox -e storage"
+            tox -e storage
+        else
+            echo "Running: tox -e storage -- -vv $test ${*:2}"
+            tox -e storage -- -vv $test "${*:2}"
+        fi
```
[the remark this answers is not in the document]
aesteve-rh: The main point is not avoiding to write this. The script handles the creation of the storage, creating (if required) and activating the venv, and then running the test. In that order. I still use it since I have it in my bin folder, and it is nice not having to worry about the venv activation/deactivation at all.

And if running from a new env (say, a new Vagrant VM), running a test through the script will generate the venv and the storage for you.
I think it can be improved to check also if it needs to run the `configure` and the first `make`.

contrib/vdsm-dev
aesteve-rh: Now that I came back to this, to refresh a bit this PR... is this point covered now that we have the `rpmbuild` directory inside the project's path?

The only thing needed now would be to actually run the build using podman, for which we can add a specific command (e.g., build-podman).

contrib/vdsm-dev
aesteve-rh: Where is the combined coverage generated? Is it the `lib` environment?

contrib/vdsm-dev
aesteve-rh: I did an attempt with `build-podman`. Please check how it looks, I have tried locally and seems to work fine, but maybe it can be improved or there is something I am missing.

I also tried to add commands to run tests in a container and things like that, but I think I would leave that for future improvements. For the time being, I guess one can just connect to a shell in the container and invoke the script to run the tests.

contrib/vdsm-dev
nirs: The code is in automation/check-patch*.sh. We used to run this after all the tests and publish the results in jenkins.

contrib/vdsm-dev
nirs: I think adding new target to the make will be more useful instead of a script in contrib.

Since we integrated most of the things in this script, maybe it is time to close this PR and open new one for every feature you want to add (e.g. build vdsm in a container).

aesteve-rh: Having it scripted allows to specify the container image in which you want to build as a script argument. The only way to achieve it in a Makefile is using an envvar (that I know, at least). If being able to select the image is not a desirable feature or using an envvar is ok, I can move it to the Makefile.

Anyway, I still see this script useful, it allows me not to worry about activating the `venv` or adding any other command line options that the script does transparently for you.

I added an option to run as root that deals with the same things as the usual run, but does not overwrite the workdir or the coverage folder. It can be further tuned to clean the new folders up after running if we want.

Also, we could integrate some other scripts in the `contrib` folder into this script, or the other way around. Examples: https://github.com/oVirt/vdsm/blob/master/contrib/shell_helper
https://github.com/oVirt/vdsm/blob/master/contrib/koji_build.sh
these could be updated and integrated together with this script.

Having all them centralized and documented is easier for new contributors.

contrib/vdsm-dev
```diff
@@ -0,0 +1,204 @@
...
+        if [ -z $test ]; then
+            echo "Missing mandatory test argument"
+            exit 1
+        fi
+
+        source ${venv_activate_path}
+        echo "Running: sudo ${venv_path}/bin/tox -e {$env} --workdir .root-tox  -- -vv $test"
+        echo "Coverage check skipped"
+        echo "Coverage report created at ./tests/htmlcov-root-$env"
+        sudo ${venv_path}/bin/tox -e {$env} --workdir .root-tox  -- -vv $test --cov-report=html:htmlcov-root-$env  --cov-fail-under 0
```
nirs, line 141: Running tests as root on a development machine is not a good idea. We need to make this easy to run a vm so we cannot harm the current host by a faulty test.

aesteve-rh:
I'll close this. The only big open left to address is the coverage target. But I have it noted for a future patch.

PR #184, opened 2022-05-18
net: Show IB interface in capabilities

almusil:
The IB interface was filtered as it was not considered
to be a NIC like interface. Add the ipoib type into
NIC like interfaces.

### DIFF
--- a/lib/vdsm/network/ipwrapper.py
+++ b/lib/vdsm/network/ipwrapper.py
@@ -90,6 +90,7 @@ class LinkType(object):
     TEAM = 'team'
     VETH = 'veth'
     VF = 'vf'
+    IPOIB = 'ipoib'
 
 
 @equals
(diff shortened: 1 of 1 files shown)


erav:
Is  a unit test required for this new type? 
Do we need to support it in engine?

almusil:
> Is a unit test required for this new type? 

Well we don't have unit test for this and for integration test we would require IB device to be present. 

> Do we need to support it in engine?

No engine is already able to show IB interfaces.

nirs:
So we try to support infininband? Do we have a feature page or more info on this?

This may requires changes in many components.

mwperina:
No feature page, just minimal change to make it working again. Let's just merge it and forget about infiniband

michalskrivanek:
> No feature page, just minimal change to make it working again. Let's just merge it and forget about infiniband

it used to work since https://bugzilla.redhat.com/show_bug.cgi?id=1856279 and broke somewhere along the way, so this is basically just a regression fix

PR #187, opened 2022-05-18
Makefile: add venv target

aesteve-rh:
Add `venv` target to the main Makefile to create a virtual environment and install the dependencies.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -35,6 +35,9 @@ SUBDIRS = \
 # prepared before running.
 SUBDIRS += tests
 
+VENV_DIR = $(HOME)/.venv/vdsm
+VENV = $(VENV_DIR)/bin
+
 include $(top_srcdir)/build-aux/Makefile.subs
 
 # This is an *exception*, we ship also vdsm.spec so it's possible to build the
@@ -134,6 +137,12 @@ check-unit:
 
 .PHONY: lint
 lint: gitignore execcmd black flake8 pylint
+
+.PHONY: venv
+venv:
+	python3 -m venv $(VENV_DIR) && \
+	$(VENV)/python3 -m pip install --upgrade pip && \
+	$(VENV)/python3 -m pip install -r docker/requirements.txt
 
 .PHONY: tests
 tests: tox
--- a/doc/development.md
+++ b/doc/development.md
@@ -30,11 +30,7 @@ Install additional packages for Fedora, CentOS, and RHEL:
 
 Create virtual environment for vdsm:
 
-    python3 -m venv ~/.venv/vdsm
-    source ~/.venv/vdsm/bin/activate
-    pip install --upgrade pip
-    pip install -r docker/requirements.txt
-    deactivate
+    make venv
 
 
 ## Building Vdsm


aesteve-rh:
This could also allow simplifying some scripts, as the `automation/tests-storage.sh` script. But I wanted to keep that for another patch maybe, if we see benefit in changing it.

Makefile.am
```diff
@@ -135,6 +138,18 @@ check-unit:
...
+venv:
+	python3 -m venv $(VENV_DIR) && \
+	$(VENV)/python3 -m pip install --upgrade pip && \
+	$(VENV)/python3 -m pip install -r docker/requirements.txt
+
+.PHONY: clean-venv
+clean-venv:
+	if [ -d "$(VENV_DIR)" ]; then \
+		rm -r $(VENV_DIR); \
+	fi
```
nirs, line 151: Do we really need this? I never cleaned my ~/.venv/vdsm directory. Why would you need to recreate it?

Makefile.am
```diff
@@ -135,6 +138,18 @@ check-unit:
 .PHONY: lint
 lint: gitignore execcmd black flake8 pylint
 
+.PHONY: venv
+venv:
+	python3 -m venv $(VENV_DIR) && \
+	$(VENV)/python3 -m pip install --upgrade pip && \
+	$(VENV)/python3 -m pip install -r docker/requirements.txt
```
nirs, line 145: Nice that we create and update the virtual environment without entering it.

Makefile.am
```diff
@@ -151,12 +151,12 @@ clean-venv:
 	fi
 
 .PHONY: tests
-tests: tox
-	tox -e "tests,lib,network,virt,gluster,hooks"
+tests: tox venv
+	. $(VENV)/activate && tox -e "tests,lib,network,virt,gluster,hooks"
```
nirs, line 155: Does it change the current shell?

I'm not sure it is a good idea to enter the venv temporarily for running the tests.

Having an easy way to create the venv is nice, simplifying life for new contributors.
But for developing, you really want to enter the venv and stay there for long time
(days?) so you edit and test quickly.

Also in the CI, we don't need the venv, since we install tox and the other
requirements in the container, so recreating the venv is just waste of time
and resources, and reason for failure (e.g. if pypi is not accessible).

I would not add a dependency or activation as part of running the tests.

Makefile.am
```diff
@@ -151,12 +151,12 @@ clean-venv:
...
 
 .PHONY: tests-storage
-tests-storage: tox
-	tox -e "storage"
+tests-storage: tox venv storage
+	. $(VENV)/activate && tox -e "storage"
```
nirs, line 159: same

Makefile.am
aesteve-rh: OK, for the CIs this is really not a good thing, true. I'll remove that dependency.

Makefile.am
aesteve-rh: What about the dependency with storage? That one is still valid I think.

mz-pdm (dismissed):

nirs (approved):
Nice! @aesteve-rh can you update doc/developmemnt.md to use the "make venv" instead of creating the venv manually?

nirs (approved):
Thanks, even nicer now!

mz-pdm (approved):

PR #192, opened 2022-05-19
lvm: make reloadlvs consistent with VGs

### LINKED ISSUE #211: lvm: rework getLV for Stale or Unreadable LVs
### Problem

LVM.getLV successfully returns an LV even if it is Stale or Unreadable (so no namedtuple's attributes),
and then the code that invoked it may raise `AttributeError` on access.

This will happen later and deep in unrelated code, and makes debugging harder.

Stale LV behaves similarly, but maybe even worse, you get `AttributeError`
without any context.

### Proposal

See who calls LVM.getLV and how these potential errors are handled.
It is probably better to fail immediately if the LV is not readable or stale. 
Explore this possibility and apply changes as needed.

aesteve-rh:
LVMCache._reloadvgs has been recently refactored for better error handling.

This branch extends the same strategy to `reloadlvs`:
- Refactor the method, according to its function:
  - Run LVM command
  - Try to reload stale LVs
  - Update all LVs
- Split reload into single and multi LVs
- Raise exceptions for missing volumes from the LVMCache, only when looking for specific volumes
  - Callers should get either a valid volume or propagate an exception
  - Fixes #211

### DIFF
--- a/lib/vdsm/storage/blockSD.py
+++ b/lib/vdsm/storage/blockSD.py
@@ -215,7 +215,7 @@ def _getVolsTree(sdUUID):
 
 
 def _iter_volumes(sdUUID):
-    for lv in lvm.getLV(sdUUID):
+    for lv in lvm.getAllLVs(sdUUID):
         if lv.name in SPECIAL_LVS_V4:
             # Exclude special volumes.
             continue
@@ -1072,7 +1072,7 @@ class BlockStorageDomain(sd.StorageDomain):
         if set((STORAGE_UNREADY_DOMAIN_TAG,)) != set(vg.tags):
             raise se.VolumeGroupHasDomainTag(vgUUID)
         try:
-            lvm.getLV(vgName)
+            lvm.getAllLVs(vgName)
             raise se.StorageDomainNotEmpty(vgUUID)
         except se.LogicalVolumeDoesNotExistError:
             pass
@@ -1263,7 +1263,7 @@ class BlockStorageDomain(sd.StorageDomain):
         # Remove all volumes LV if exists
         _removeVMSfs(lvm.lvPath(sdUUID, MASTERLV))
         try:
-            lvs = lvm.getLV(sdUUID)
+            lvs = lvm.getAllLVs(sdUUID)
         except se.LogicalVolumeDoesNotExistError:
             lvs = ()  # No LVs in this VG (domain)
 
--- a/lib/vdsm/storage/blockSD.py
+++ b/lib/vdsm/storage/blockSD.py
@@ -1071,11 +1071,10 @@ class BlockStorageDomain(sd.StorageDomain):
 
         if set((STORAGE_UNREADY_DOMAIN_TAG,)) != set(vg.tags):
             raise se.VolumeGroupHasDomainTag(vgUUID)
-        try:
-            lvm.getAllLVs(vgName)
+
+        if lvm.getAllLVs(vgName):
+            # There are already LVs in the VG
             raise se.StorageDomainNotEmpty(vgUUID)
-        except se.LogicalVolumeDoesNotExistError:
-            pass
 
         # Create metadata service volume. Metadata have to be stored always on
         # the VG metadata device, which is always the first PV.
@@ -1262,12 +1261,8 @@ class BlockStorageDomain(sd.StorageDomain):
         # Remove special metadata and service volumes
         # Remove all volumes LV if exists
         _removeVMSfs(lvm.lvPath(sdUUID, MASTERLV))
-        try:
-            lvs = lvm.getAllLVs(sdUUID)
-        except se.LogicalVolumeDoesNotExistError:
-            lvs = ()  # No LVs in this VG (domain)
 
-        for lv in lvs:
+        for lv in lvm.getAllLVs(sdUUID):
             # Fix me: Should raise and get resource lock.
             try:
                 lvm.removeLVs(sdUUID, (lv.name,))
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -767,7 +767,7 @@ class LVMCache(object):
         Must be called while holding the lock.
         Return dict of updated LVs.
         """
-        updatedLVs = {}
+        updated_lvs = {}
         for line in lvs_output:
             fields = [field.strip() for field in line.split(SEPARATOR)]
             if len(fields) != LV_FIELDS_LEN:
@@ -777,19 +777,19 @@ class LVMCache(object):
             # For LV we are only interested in its first extent
             if lv.seg_start_pe == "0":
                 self._lvs[(lv.vg_name, lv.name)] = lv
-                updatedLVs[(lv.vg_name, lv.name)] = lv
+                updated_lvs[(lv.vg_name, lv.name)] = lv
 
         # Determine if there are stale LVs
         items = self._lvs if lv_name is None else [(vg_name, lv_name)]
         stale_lvs = [lvn for vgn, lvn in items
-                     if vgn == vg_name and (vgn, lvn) not in updatedLVs]
+                     if vgn == vg_name and (vgn, lvn) not in updated_lvs]
 
         for lvName in stale_lvs:
             if (vg_name, lvName) in self._lvs:
                 log.warning("Removing stale lv: %s/%s", vg_name, lvName)
                 del self._lvs[(vg_name, lvName)]
 
-        return updatedLVs
+        return updated_lvs
 
     def _update_stale_lvs_locked(self, vg_name):
         """
@@ -797,7 +797,7 @@ class LVMCache(object):
         Log a warning if any LV has been made Unreadable.
         Must be called while holding the lock.
         """
-        updatedLVs = {}
+        updated_lvs = {}
         lv_names = (lvn for vgn, lvn in self._lvs if vgn == vg_name)
         for lvName in lv_names:
             key = (vg_name, lvName)
@@ -805,17 +805,17 @@ class LVMCache(object):
             if lv and lv.is_stale():
                 lv = Unreadable(lv.name)
                 self._lvs[key] = lv
-                updatedLVs[key] = lv
+                updated_lvs[key] = lv
 
-        if updatedLVs:
+        if updated_lvs:
             # This may be a real error (failure to reload existing LV)
             # or no error at all (failure to reload non-existing LV),
             # so we cannot make this an error.
             log.warning(
                 "Marked lvs=%r as Unreadable due to reload failure",
-                logutils.Head(updatedLVs, max_items=20))
+                logutils.Head(updated_lvs, max_items=20))
 
-        return updatedLVs
+        return updated_lvs
 
     def _reload_single_lv(self, vg_name, lv_name):
         """
@@ -846,24 +846,24 @@ class LVMCache(object):
 
         return updated_lvs[(vg_name, lv_name)]
 
-    def _reloadlvs(self, vgName):
+    def _reloadlvs(self, vg_name):
         cmd = list(LVS_CMD)
-        cmd.append(vgName)
+        cmd.append(vg_name)
 
         out, error = self.run_command_error(
-            cmd, devices=self._getVGDevs((vgName,)))
+            cmd, devices=self._getVGDevs((vg_name,)))
 
         with self._lock:
             if error:
-                return self._update_stale_lvs_locked(vgName)
+                return self._update_stale_lvs_locked(vg_name)
 
-            updatedLVs = self._updatelvs_locked(out, vgName)
+            updated_lvs = self._updatelvs_locked(out, vg_name)
 
-            self._freshlv.add(vgName)
+            self._freshlv.add(vg_name)
 
             log.debug("lvs reloaded")
 
-        return updatedLVs
+        return updated_lvs
 
     def _loadAllLvs(self):
         """
@@ -1057,7 +1057,7 @@ class LVMCache(object):
                 self.stats.hit()
         return list(vgs.values())
 
-    def getLv(self, vgName, lvName):
+    def getLv(self, vg_name, lv_name):
         """
         Get specific LV in specified VG.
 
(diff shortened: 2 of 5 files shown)


lib/vdsm/storage/lvm.py
```diff
@@ -736,10 +736,8 @@ def _updatelvs_locked(self, lvs_output, vg_name, lv_names):
         - Add new LVs to the cache.
         - Replace LVs in the cache with LVs reported by the 'lvs' command,
           updating the LV attributes.
-        - If called without lv names, remove all LVs in the VG from the cache
-          not reported by LVM.
-        - If called with lv names, remove specifed LVs in the VG from the cache
-          if they were not reported by LVM.
+        - Remove specifed LVs in the VG from the cache if they were not
+          reported by LVM.
```
nirs, line 767: This change is complicated and not really needed. We have 3 calls to _reloadlvs():

lib/vdsm/storage/lvm.py:993:                lvs = self._reloadlvs(vgName)
lib/vdsm/storage/lvm.py:1002:            lvs = self._reloadlvs(vgName)
lib/vdsm/storage/lvm.py:1782:        _lvminfo._reloadlvs(vg, newlv)

The only call with a lv name is in renameLV:

```
def renameLV(vg, oldlv, newlv):
    log.info("Renaming LV (vg=%s, oldlv=%s, newlv=%s)", vg, oldlv, newlv)
    cmd = ("lvrename",) + LVM_NOBACKUP + (vg, oldlv, newlv)
    try: 
        _lvminfo.run_command(cmd, devices=_lvminfo._getVGDevs((vg, )))
    except se.LVMCommandError as e:
        raise se.LogicalVolumeRenameError.from_lvmerror(e)
    else:
        _lvminfo._removelvs(vg, oldlv)
        _lvminfo._reloadlvs(vg, newlv)
```

Since you remove renameLV in this PR, we have only one case, lvNames=None.

Lets first remove renameLV - before any other change. Then remove the lvNames parameter and the code handling it.

When we finish to remove the dead code we can go back to unifying the behavior of _reloadlvs/pvs.

lib/vdsm/storage/blockVolume.py
```diff
@@ -35,7 +35,6 @@
 from vdsm.storage import lvm
```
nirs, line 35: Lets move this to first patch so we can simplify _reloadlvs.

lib/vdsm/storage/lvm.py
```diff
@@ -1417,12 +1420,11 @@ def getVGbyUUID(vgUUID):
...
-    # getLV() should not return None
-    if not lv:
-        raise se.LogicalVolumeDoesNotExistError("%s/%s" % (vgName, lvName))
-    else:
-        return lv
+    """
+    Return LV named tuple. Raise se.LogicalVolumeDoesNotExistError if the
+    LV does not exist.
+    """
+    return _lvminfo.getLv(vgName, lvName)
```
nirs, line 1441: This is not like getVG - when called with lvName=None, this returns all the lvs
in the vg, and raising se.LogicalVolumeDoesNotExistError does not make sense
since the caller did not ask to get a certain lv.

If we want to unify this we need to require lvName != None here and handle only the
case of returning single lv, and add another function getAllLVs(vgName) for
returning list of all lvs - which will never raise if the list is empty.

We also need to change LVMCache.getLv() to always require lv name, and add another methods like LVMCache.getAllLVs().

Lets first make the API sane before we try to unify error handling.

lib/vdsm/storage/lvm.py
```diff
@@ -577,9 +579,18 @@ def _reload_single_pv(self, pv_name):
 
         with self._lock:
             if error:
-                return self._update_stale_pvs_locked([pv_name])
+                updated_pvs = self._update_stale_pvs_locked([pv_name])
+                if pv_name not in updated_pvs:
+                    raise se.InaccessiblePhysDev((pv_name,))
```
nirs, line 584: Shouldn't we include the error in se.InaccessiblePhysDev like se.VolumeGruopDoesNotExists?

This will let the caller log the error if needed. What we do here will always hide the lvm error.

nirs (changes requested):

lib/vdsm/storage/blockVolume.py
nirs: Also, can you find the commit that remove the code using BlockVolume.rename() and
add

    Fixes: {commit-hash}

to the commit message?

lib/vdsm/storage/lvm.py
aesteve-rh: If we want to make the API consistent with VGs (which sounds good to me) we definitively need to add those extra methods. Also, it probably will make it easier for the reader to follow the code.

But the current API already does what you describe, only raise if `lvName != None`. This is handled by the `LVMCache.getLv()`. It will not raise if the list is empty, and actually that behavioral change triggered the changes in the `lvm_test` and `blockSD`. This is reflected in the commit message. However, it is not reflected correctly in the docstring of this function, unfortunately.

nirs: The current API may work but I don't want to invest time in improving
it.

lib/vdsm/storage/lvm.py
```diff
@@ -746,7 +746,7 @@ def _reloadvgs(self, vgName=None):
 
         return updatedVGs
 
-    def _updatelvs_locked(self, lvs_output, vg_name, lv_names):
+    def _updatelvs_locked(self, lvs_output, vg_name):
```
nirs: This should be the second patch after removing renameLV. Do you want to split this work to another PR to make it easier and quicker?

Adding more changes to this PR will only make it slower to review. Lets finish first with removal of dead code and options that we don't need. When we have simpler code it will be easier to refactor it safely.

nirs (changes requested):
Partial review - I want to finish first with removing renameLV and the lvNames argument.

lib/vdsm/storage/lvm.py
aesteve-rh: Yeah, good idea, I'll split this PR.

lib/vdsm/storage/lvm.py
```diff
@@ -782,14 +781,13 @@ def _updatelvs_locked(self, lvs_output, vg_name):
...
 
-    def _update_stale_lvs_locked(self, vg_name):
+    def _update_stale_lvs_locked(self, vg_name, lv_names):
         """
         Check if any LV is stale after an LVM command and make it Unreadable.
         Log a warning if any LV has been made Unreadable.
         Must be called while holding the lock.
         """
         updatedLVs = {}
-        lv_names = (lvn for vgn, lvn in self._lvs if vgn == vg_name)
```
aesteve-rh: With the rebase this commit makes little sense now. I might just change this line to a list to be consistent with the other updates that manage a list, or remove the commit entirely.

nirs: Lets remove the commit.

Yes, creating the list once is nice, but the code is nicer when we don't have to do
this, and we should really fix the cache structure so creating the list of lvs is
not complicated and expensive as it is now. Once we do this we don't need to be
clever about this.

For example if we had:

     cache = {
         "vgs": [
             "vg-name-1": {
                 "lvs": {
                     "lv-name-1": ...
                     ...
                 }
             ...             
             }
              ...
         ]
         "pvs": {
             "path": ...,
             ...
         }

Getting vg lvs would be:

    cache["vg-name-1"["lvs"]

Getting all lvs is chaining the lvs from all vgs.

The current cache try to make it easy to have flag list of lvs, we don't need to make it easy since we should never do operations on all lvs, only on certain vg.

nirs: The lv names are accessible via the vg, not the lv objects. We can change VG to be
a class with mutable lvs list and store there the LV named tuple, or have something like:

    class VGInfo:
       def __init__(self, vg, lvs):
           self.vg = vg
           self.lvs = lvs

And store the VGInfo object in the cache instead of the VG objects.

This makes it trivial to invalidate a vg and all the lvs from the cache,
or access one VG lvs.

lib/vdsm/storage/blockSD.py
```diff
@@ -1071,11 +1071,10 @@ def create(cls, sdUUID, domainName, domClass, vgUUID, storageType,
...
         if set((STORAGE_UNREADY_DOMAIN_TAG,)) != set(vg.tags):
             raise se.VolumeGroupHasDomainTag(vgUUID)
-        try:
-            lvm.getAllLVs(vgName)
+
+        if lvm.getAllLVs(vgName):
+            # There are already LVs in the VG
             raise se.StorageDomainNotEmpty(vgUUID)
-        except se.LogicalVolumeDoesNotExistError:
-            pass
```
nirs, line 1078: Much nicer like this - but do we have a case when lvm.getAllLVs(vgName) raises
and we need to handle here?

Previously we assumed that this error means the vg is empty. I wonder if we could
get this error when the vg was not empty, but we cannot access storage.

The new code seems better, in case of errors it will raise and fail the current flow.

lib/vdsm/storage/blockSD.py
```diff
@@ -1262,12 +1261,8 @@ def format(cls, sdUUID):
         # Remove special metadata and service volumes
         # Remove all volumes LV if exists
         _removeVMSfs(lvm.lvPath(sdUUID, MASTERLV))
-        try:
-            lvs = lvm.getAllLVs(sdUUID)
-        except se.LogicalVolumeDoesNotExistError:
-            lvs = ()  # No LVs in this VG (domain)
 
-        for lv in lvs:
+        for lv in lvm.getAllLVs(sdUUID):
```
nirs, line 1265: Much better!

lib/vdsm/storage/lvm.py
```diff
@@ -1425,12 +1427,11 @@ def getVGbyUUID(vgUUID):
...
-    # getLV() should not return None
-    if not lv:
-        raise se.LogicalVolumeDoesNotExistError("%s/%s" % (vgName, lv_name))
-    else:
-        return lv
+    """
+    Return LV named tuple. Raise se.LogicalVolumeDoesNotExistError if the
+    LV does not exist.
+    """
+    return _lvminfo.getLv(vgName, lv_name)
```
nirs, line 1448: Awsome!

lib/vdsm/storage/lvm.py
```diff
@@ -1017,49 +1017,58 @@ def getAllVgs(self):
...
+        Get specific LV in specified VG.
 
         If there are any stale LVs reload the whole VG, since it would
         cost us around same efforts anyhow and these stale LVs can
         be in the vg.
 
-        We never return Stale or Unreadable LVs when
-        getting all LVs for a VG, but may return a Stale or an Unreadable
-        LV when LV name is specified as argument.
+        May return a Stale or an Unreadable LV.
```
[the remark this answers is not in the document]
aesteve-rh: I'll write an issue with this point.

lib/vdsm/storage/lvm.py
```diff
@@ -1017,49 +1017,58 @@ def getAllVgs(self):
...
+        Arguments:
+            vgName (str): VG name to query.
+
+        Returns:
+            List of LV namedtuple for all lvs in VG vg_name.
+        """
+        if self._lvs_needs_reload(vg_name):
             self.stats.miss()
-            lvs = self._reloadlvs(vgName)
+            lvs = self._reloadlvs(vg_name)
```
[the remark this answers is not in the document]
aesteve-rh: Yes, I know. In this case, technically this is a new method (with existing code) and I thought it would be acceptable to change the parameter name.

I'll revert to the previous name and probably add a new commit only for renames (otherwise renaming never happens).

lib/vdsm/storage/blockSD.py
aesteve-rh: I don't think so. We reload LVs through an LVM with silent errors, and only print a warning in a reload failure for an Unreadable volume. Other than that, we do not raise anymore for the `AllLVs` getter.

I think the new code expresses better the fact that we expect an empty VG (without LVs). Maybe I'll add a `> 0` to the condition to make it more explicit.

lib/vdsm/storage/lvm.py
```diff
@@ -1017,49 +1017,58 @@ def getAllVgs(self):
...
-            namedtuple for all lvs in VG vgName.
+            LV nameduple.
         """
+        # vgName, lvName
+        lv = self._lvs.get((vgName, lvName))
+        if not lv or lv.is_stale():
+            self.stats.miss()
+            # while we here reload all the LVs in the VG
+            lvs = self._reloadlvs(vgName)
+            lv = lvs.get((vgName, lvName))
```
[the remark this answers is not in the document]
aesteve-rh: Here it can be `None`. In this patch I keep the code as is and is the top level `getLV` where we raise if None. In the next patch I move the raise here, closer to the getter.

Since we do not want to silence the `KeyError` but rather just raise when the key is not there, I'll follow your suggestion and change it for the brackets and catch the `KeyError` exception and raise from there.

lib/vdsm/storage/lvm.py
aesteve-rh: True, contains LV names not the objects. Still, it could be used to get the LVs in a VG without having to loop through them all. But I agree, it is better to change the design to some other option before workaround the current design.

I think I'll write an issue to keep it in mind.

aesteve-rh: Here it goes! https://github.com/oVirt/vdsm/issues/212

nirs (changes requested):
The PVs part looks ready - lets split it out so we can merge it.

The LVs part need to handle loading single lv and reporting the underlying error.

lib/vdsm/storage/lvm.py
```diff
@@ -725,57 +725,57 @@ def _updatelvs_locked(self, lvs_output, vg_name, lv_name=None):
...
             if lv.seg_start_pe == "0":
                 self._lvs[(lv.vg_name, lv.name)] = lv
-                updatedLVs[(lv.vg_name, lv.name)] = lv
+                updated_lvs[(lv.vg_name, lv.name)] = lv
 
         # Determine if there are stale LVs
         if lv_name is None:
             # All the LVs in the VG
-            staleLVs = [lvName for v, lvName in self._lvs
+            staleLVs = [lv_name for v, lv_name in self._lvs
```
nirs, line 733: This shadows the argument name - should be avoided in general.

nirs (changes requested):

lib/vdsm/storage/lvm.py
```diff
@@ -766,6 +777,28 @@ def _update_stale_lvs_locked(self, vg_name):
...
+        out, error = self.run_command_error(
+            cmd, devices=self._getVGDevs((vg_name,)))
+
+        with self._lock:
+            if error:
+                return self._update_stale_lvs_locked(vg_name)
+
+            updated_lvs = self._updatelvs_locked(out, vg_name, lv_name)
+
+            self._freshlv.add(vg_name)
```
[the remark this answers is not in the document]
aesteve-rh: It is wrong, mb, I'll fix it.

PR #198, opened 2022-05-23
Makefile: add upgrade target and custom rpm build folder

aesteve-rh:
Add the `upgrade` target to the makefile to be able to upgrade the system with the built rpm packages by just doing:

    make upgrade

Also, we move the built rpm packages from the default directory at `~/rpmbuilds` to a project-specific `build` folder, and add a `clean-build` target that is executed before building. This way we don't need to clean the folder manually, it gets cleaned when we execute `make clean` or `make rpm`.

With both new targets we can just build and deploy (upgrade) the system by doing:

    make rpm upgrade

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -166,6 +166,10 @@ all-local: \
 
 clean-local: clean-build
 
+.PHONY: upgrade
+upgrade:
+	sudo dnf upgrade $(RPM_TOPDIR)/RPMS/*/*.rpm
+
 .PHONY: srpm rpm clean-build
 
 srpm: clean-build dist
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,6 +37,7 @@ SUBDIRS += tests
 
 VENV_DIR = $(HOME)/.venv/vdsm
 VENV = $(VENV_DIR)/bin
+RPM_TOPDIR = $(PWD)/build
 
 include $(top_srcdir)/build-aux/Makefile.subs
 
@@ -163,14 +164,18 @@ clean-storage:
 all-local: \
 	vdsm.spec
 
-.PHONY: srpm rpm
+clean-local: clean-build
 
-srpm: dist
+.PHONY: srpm rpm clean-build
+
+srpm: clean-build dist
 	rpmbuild -ts $(if $(BUILDID),--define="extra_release .$(BUILDID)") \
+		--define="_topdir $(RPM_TOPDIR)" \
 		$(DIST_ARCHIVES)
 
-rpm: dist
+rpm: clean-build dist
 	rpmbuild -ta $(if $(BUILDID),--define="extra_release .$(BUILDID)") \
+		--define="_topdir $(RPM_TOPDIR)" \
 		--define="qemu_user $(QEMUUSER)" \
 		--define="qemu_group $(QEMUGROUP)" \
 		--define="with_ovirt_vmconsole $(OVIRT_VMCONSOLE)" \
@@ -180,6 +185,9 @@ rpm: dist
 		--define="vdsm_version $(PACKAGE_VERSION)" \
 		--define="vdsm_release $(PACKAGE_RELEASE)" \
 		$(DIST_ARCHIVES)
+
+clean-build:
+	rm -rf $(RPM_TOPDIR)
 
 dist-hook: gen-VERSION gen-ChangeLog
 .PHONY: gen-VERSION gen-ChangeLog
--- a/automation/rpm.sh
+++ b/automation/rpm.sh
@@ -20,7 +20,7 @@ mkdir -p ${EXPORT_DIR}
 
 cp $PWD/lib/vdsm/api/vdsm-api.html "${EXPORT_DIR}"
 
-find ~/rpmbuild \
+find $PWD/build \
     -iname '*.rpm' \
     -exec mv {} "${EXPORT_DIR}/" \;
 
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ Makefile
 Makefile.in
 aclocal.m4
 autom4te.cache
+build
 build-aux/compile
 build-aux/config.guess
 build-aux/config.sub
--- a/doc/development.md
+++ b/doc/development.md
@@ -47,12 +47,11 @@ To test Vdsm (refer to tests/README for further tests information):
 
 To create an RPM:
 
-    rm -rf ~/rpmbuild/RPMS/*/vdsm*.rpm
     make rpm
 
-To update your system with local build's RPM:
+To upgrade your system with local build's RPM:
 
-    (cd ~/rpmbuild/RPMS && sudo dnf upgrade */vdsm*.rpm)
+    make upgrade
 
 
 ## Making new releases


Makefile.am
```diff
@@ -163,7 +164,18 @@ clean-storage:
 all-local: \
 	vdsm.spec
 
-.PHONY: srpm rpm
+.PHONY: config
+config:
+	./autogen.sh --system --enable-timestamp
```
nirs, line 169: This does not work, since the Makefile is generated by running ./autogen.sh.

Makefile.am
```diff
@@ -163,7 +164,18 @@ clean-storage:
...
+
+.PHONY: dist-build
+dist-build: | clean config all clean-rpm rpm
```
nirs, line 172: We cannot use config here, but otherwise using make targets for building and installing looks good.

However if we need to run autogen.sh in some cases, for example for after adding missing tags, it is better to add external script for deploying, for example:

    # dev-install.sh
    #!/bin/sh
    The long one liner to clean, build, and upgrade the rpms...

Makefile.am
```diff
@@ -181,6 +193,9 @@ rpm: dist
 		--define="vdsm_release $(PACKAGE_RELEASE)" \
 		$(DIST_ARCHIVES)
 
+clean-rpm:
+	rm -rf $(RPM_DIR)
```
nirs, line 197: This delete the entire rpbbuild, which may be bad if you tried to build other packages
and you want to use the built packages. I do this when building vdsm since I never
build other stuff in rpmbuild, but I do the explicit "rm -rf ~/rpmbuild", so I'm not
surprised by the results.

I think we should start using our own rpmbuild directory - using a clean directory per
build, used only for vdsm. See imageio makefile for example:
https://github.com/oVirt/ovirt-imageio/blob/0568314462a47e8c3ddffc4c29520b67ced41c33/Makefile#L31

nirs (changes requested):

Makefile.am
aesteve-rh: True. It is a pity, it led to a nice solution where all the command were executed through dependencies. But if we cannot do it, we would need to skip the `clean config` targets.

I tested locally and seemed to work, but I didn't notice that for the next steps it was using the Makefile stored in memory and not the freshly generated.

Makefile.am
aesteve-rh: I wonder, actually reading this, if it would work if instead of having the targets as dependencies we do:

```
dist-build: | clean config
    make all && make clean-rpm && make rpm
```

In theory these are new instances of the newly generate Makefile, so it may work. Anyway, it is obscure and dangerous, so I will just remove the `config` target as the proper solution.

Makefile.am
```diff
@@ -37,6 +37,7 @@ SUBDIRS += tests
 
 VENV_DIR = $(HOME)/.venv/vdsm
 VENV = $(VENV_DIR)/bin
+RPM_TOPDIR = $(PWD)/build/rpm
```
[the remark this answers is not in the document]
nirs: In imageio we use a sub directory because we use the same build directory python used when running "python setup.py build", and after buidling rpms, we move them the output dir. In vdsm the build directory is used only for rpms, so we can simplify.

nirs (approved):
Very nice!

@mz-pdm @tinez @erav please review.

erav (commented):
I am not familiar with make files or rpmbuilds, so can't review this patch.

Makefile.am
```diff
@@ -37,6 +37,7 @@ SUBDIRS += tests
 
 VENV_DIR = $(HOME)/.venv/vdsm
 VENV = $(VENV_DIR)/bin
+RPM_TOPDIR = $(PWD)/build
```
mz-pdm, line 40: Do we really use the current working directory?  Or do we want to use the vdsm directory: $(lastword $(MAKEFILE_LIST))?  I'd prefer the latter, it's safer and IMO generally better.

Makefile.am
```diff
@@ -181,6 +189,9 @@
 		--define="vdsm_release $(PACKAGE_RELEASE)" \
 		$(DIST_ARCHIVES)
 
+clean-build:
```
mz-pdm, line 192: Why not to put it as a dependency also to `clean` target?

automation/rpm.sh
```diff
@@ -20,7 +20,7 @@ mkdir -p ${EXPORT_DIR}
 
 cp $PWD/lib/vdsm/api/vdsm-api.html "${EXPORT_DIR}"
 
-find ~/rpmbuild \
+find $PWD/build \
```
mz-pdm, line 23: $(RPM_TOPDIR)

mz-pdm (commented):
Generally good, some comments inside.

Makefile.am
aesteve-rh: Sounds better, yes. I took the same strategy as in `imageio`, but I agree with you that it would be safer. I'll try your proposal and change it.

Makefile.am
aesteve-rh: There is no clean target in the Makefile, if I understood correctly the flow, it is generated when you call the `autogen.sh` script. The way to influence which **files** will be removed in the `clean` target is by adding them to the `CLEANFILES` set.

automation/rpm.sh
aesteve-rh: It shouldn't be available from within this script?

mz-pdm: Ah, right, sorry for noise.

Makefile.am
mz-pdm: It seems `clean-local` target can be used for the purpose: https://www.gnu.org/software/automake/manual/automake.html#Clean

Makefile.am
nirs: @mz-pdm: What is the vdsm directory? Why using it is better?

nirs: Do you mean running "make -C /path/to/vdsm/dir"? Can you show an example creating build
directory in the wrong place with this pr?

mz-pdm: > Do you mean running "make -C /path/to/vdsm/dir"?

I was thinking of `make -f /path/to/vdsm/dir`. But this probably wouldn't work anyway and `-C` should be used instead.  OK.

Makefile.am
aesteve-rh: Great!

Atm I could just change `clean-build` ->  `clean-local` and it would work, but in case we want to extend it, I think it would be nice to keep them separated.

So I will keep `clean-build` and create a `clean-local` that depends on `clean-build`.

nirs: Changing clean-build to clean-local will make clean-local run when running `make clean`.
This is nice, but you already added the files in the build directory to CLEANFILES list,
so they are already cleaned when running `make clean`.

However when this is called `clean-local` people will add more files there and since
we depend on this target in `make rpm` these files will be remove, even if they are 
needed for building rpms.

Did you test

   make srpm rpm

works and create both rpms and srpm? Or does it create the first target and then remove
the build directory before creating the next?

Makefile.am
```diff
@@ -65,6 +66,7 @@ EXTRA_DIST = \
 CLEANFILES = \
 	vdsm.spec \
 	$(DIST_ARCHIVES) \
+	$(wildcard $(RPM_TOPDIR)/*/*/*.rpm) \
```
nirs, line 69: Does it works to remove the entire build directory or there is CLEANDIRS for this?

Makefile.am
aesteve-rh: On your first point, is more or less what I tried to say above. If we want to have a clean-local target for the future and keep adding other folders, having `clean-build` and `clean-local` separated is the right approach. Also, that allows to construct `clean-local` with dependencies if we want. And I can remove the wildcard from the `CLEANFILES` set.

About the
   
    make srpm rpm

That's a good point that I didn't think about, but I just tested and works fine, it only cleans it once at first.

Makefile.am
aesteve-rh: I tried for completeness. With -f fails due to the dependencies with other files. Fails to find them since they are linked with path relative to the repository root. If you invoke the script from somewhere else, they just break.

With -C works fine, it actually changes the directory before invoking, so it creates the `build` folder in the excepted place in the repository root.

I read the manual for other options and tried to think how they could break this. I could hack it by adding `-I` to include directories for dependencies on a run with `-f`. But nobody will realistically do that, ever. So I will keep it as it is for now, as it is kept consistent across projects and is secure enough. But it was a good point to bring up and test!

mz-pdm (approved):

PR #199, opened 2022-05-23
tox.ini: allow newer black version to match click version

aesteve-rh:
Latest click versions are not compatible with black
version used by tox (21.6b0). To fix it we either
have to move to a more recent version of black, which
currently breaks the linter in some tests, or we downgrade
the version of click to 8.0.2.

See https://github.com/psf/black/issues/2964 for context.

Error seen when running `make lint`:
```
  File "/home/aesteve/Work/ws1/vdsm/.tox/black/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "/home/aesteve/Work/ws1/vdsm/.tox/black/lib/python3.9/site-packages/black/__init__.py", line 1127, in patched_main
    patch_click()
  File "/home/aesteve/Work/ws1/vdsm/.tox/black/lib/python3.9/site-packages/black/__init__.py", line 1113, in patch_click
    from click import _unicodefun  # type: ignore
ImportError: cannot import name '_unicodefun' from 'click' (/home/aesteve/Work/ws1/vdsm/.tox/black/lib/python3.9/site-packages/click/__init__.py)
```

This patch allows newer black versions and fixes the linting problems found.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -169,7 +169,7 @@ commands=
 sitepackages = false
 skip_install = true
 deps =
-    black==21.6b0
+    black==22.3.0
 commands =
     black \
         -l 79 \
--- a/lib/vdsm/network/link/nic.py
+++ b/lib/vdsm/network/link/nic.py
@@ -47,7 +47,7 @@ def read_speed_using_sysfs(nic_name):
     # the device may have been disabled/downed after checking
     # so we validate the return value as sysfs may return
     # special values to indicate the device is down/disabled
-    if s in (2 ** 16 - 1, 2 ** 32 - 1) or s <= 0:
+    if s in (2**16 - 1, 2**32 - 1) or s <= 0:
         raise ReadSpeedValueError(s)
     return s
 
--- a/lib/vdsm/network/tc/_parser.py
+++ b/lib/vdsm/network/tc/_parser.py
@@ -47,15 +47,15 @@ def parse_rate(tokens):
     """Returns a numerical bit representation of the textual rate in tokens"""
     rate = next(tokens)
     if rate[-5:] == 'Gibit':
-        return int(float(rate[:-5]) * 1024 ** 3)
+        return int(float(rate[:-5]) * 1024**3)
     elif rate[-5:] == 'Mibit':
-        return int(float(rate[:-5]) * 1024 ** 2)
+        return int(float(rate[:-5]) * 1024**2)
     elif rate[-5:] == 'Kibit':
         return int(float(rate[:-5]) * 1024)
     elif rate[-4:] == 'Gbit':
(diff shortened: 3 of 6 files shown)


aesteve-rh:
[Click](https://click.palletsprojects.com/en/8.1.x/) is a pretty nice command line interface library that is used by black, you can see that it actually breaks in the __init__ file of the black package:
```
  File "/home/aesteve/Work/ws1/vdsm/.tox/black/lib/python3.9/site-packages/black/__init__.py", line 1113, in patch_click
    from click import _unicodefun  # type: ignore
```
In the project, the version of black is fixed to an older version, but click version is not fixed, causing this incompatibility. And it was consistently failing in my environment.

If I update black version instead, I see linter issues with `network` tests, but since that might be intended, I think the best option to make the environment consistent is to fix also click version to a compatible one.

almusil (dismissed):
Well not really intended, we would need to fix all the new errors that pop up during the update. Having fixed version of click is fine by me.

nirs:
Since we don't use click, this looks like an issue in black and should be fixed in black
and not in vdsm.

If new black version solved this, it is better to upgrade black and fix the code to match
new black version.

tox.ini
```diff
@@ -169,7 +169,7 @@ commands=
 sitepackages = false
 skip_install = true
 deps =
-    black==21.6b0
+    black>=21.6b0
```
nirs, line 172: I'm not sure this is useful, better to require the specific version fixing the issue
with click instead of a range.

nirs (commented):
The commit message says:

> Allow more recent versions of tox and fix linting issues.

But you upgraded `black` - typo?

Looks like the right way to fix this.

tox.ini
almusil: We should always use specific version as any new one might break the formatting out of nowhere.

almusil:
> Changed to allow newer black version instead of fixing click to an older version.
> 
> That broke the pipelines, but I'm not sure why. Any clue @almusil? Maybe it was actually fixed intentionally?

The black is invoked in lints and those are ok. Functional tests being broken is fault of nmstate not being back compatible on el9s. We are waiting for fix but it did not arrive yet.

aesteve-rh:
> The black is invoked in lints and those are ok. Functional tests being broken is fault of nmstate not being back compatible on el9s. We are waiting for fix but it did not arrive yet.

Thanks for the explanation. Since the only change of the branch was on the black version, I thought the error in the pipelines might be related and that some sort of compatibility was involved.

I'll set black to a specific version and wait for the fix to make the pipelines pass.

almusil (approved):
Thanks

nirs:
@aesteve-rh looks like you need to fix conflicts.

PR #209, opened 2022-05-25
managedvolume: Add lightos support

bennyz:
* Extract the supported drivers tuple to a variable to make it easier to add drivers in the future
* Add lightos to the supported drivers

### DIFF
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -58,9 +58,14 @@ DEV_MAPPER = "/dev/mapper"
 DEV_RBD = "/dev/rbd"
 VOLUME_LINK_DIR = "/run/vdsm/managedvolumes/"
 
+# Drivers that have been tested and are known to work.
 SUPPORTED_DRIVERS = (
+    # Drivers tested by the oVirt team
     "rbd",
     "iscsi",
+
+    # Tested by Muli Ben-Yehuda <info (at) lightbitslabs.com>
+    "lightos",
 )
 
 log = logging.getLogger("storage.managedvolume")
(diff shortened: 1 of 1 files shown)


nirs (changes requested):
Awsome

lib/vdsm/storage/managedvolume.py
```diff
@@ -108,10 +113,10 @@ def attach_volume(sd_id, vol_id, connection_info):
                     multipath_id=attachment.get("multipath_id"))
                 _invalidate_lvm_devices(attachment)
                 volume_type = connection_info["driver_volume_type"]
-                if volume_type not in ("rbd", "iscsi"):
+                if volume_type not in supported_drivers:
                     raise se.UnsupportedOperation(
-                        "Unsupported volume type, supported types are: "
-                        "rbd, iscsi")
+                        f"Unsupported volume type, supported types are: "
+                        "{supported_drivers}")
```
aesteve-rh, line 123: Does this work? I'd say the second line of the string needs the `f` to be an f-string. The first line does not need it, but I think looks better if we keep it for consistency.

bennyz: yeah, fixed it locally

aesteve-rh (dismissed):

nirs (dismissed):

lib/vdsm/storage/managedvolume.py
```diff
@@ -58,9 +58,14 @@
 DEV_RBD = "/dev/rbd"
 VOLUME_LINK_DIR = "/run/vdsm/managedvolumes/"
 
+# Drivers that have been tested and are known to work.
 SUPPORTED_DRIVERS = (
+    # Drivers tested by the oVirt team
     "rbd",
     "iscsi",
+
+    # Tested by Muli Ben-Yehuda
```
muliby-lb, line 67: feel free to also add a contact email address, say `<info (at) lightbitslabs.com>`

muliby-lb (approved):
Awesome!

PR #215, opened 2022-05-26
virt: use proper CPU pinning when resuming hibernated VM

nyoxi:
When hibernating a VM we store also the CPU pinning. This pinning needs
to be invalidated when resuming the VM because in most cases (except
when manual pinning is involved) the pinning is no longer usable either
because shared pool needs to be adapted or dedicated CPUs have changed.

New pinning passed from engine needs to be used instead.

Bug-Url: https://bugzilla.redhat.com/2083302
Signed-off-by: Tomáš Golembiovský <tgolembi@redhat.com>


lib/vdsm/virt/cpumanagement.py
```diff
@@ -160,6 +163,37 @@ def libvirt_cpuset_spec(cpus, cpu_list_length):
...
 
+def resume_cpu_pinning(from_engine, stored):
+    """
+    Update CPU pinning in domain XML of VM resumed from hibernation.
+
+    :param from_engine: domain XML passed from engine
+    :type from_engine: str
+    :param stored: domain XML stored in hibernation volume of VM
+    :param stored: str
+    """
```
mz-pdm, line 174: The arguments are not domain XMLs, they are parameter dictionaries.

lib/vdsm/virt/cpumanagement.py
```diff
@@ -160,6 +163,37 @@ def libvirt_cpuset_spec(cpus, cpu_list_length):
...
+        for vcpu in vmxml.find_all(cputune_stored, 'vcpupin'):
+            cputune_stored.remove(vcpu)
+    else:
+        cputune_stored = ET.Element('cputune')
+        dom_stored.append(cputune_stored)
+    # Move pinning we have from engine
+    cputune_engine = dom_engine.find('cputune')
+    if cputune_engine is not None:
+        for vcpu in vmxml.find_all(cputune_engine, 'vcpupin'):
+            cputune_engine.remove(vcpu)
```
mz-pdm, line 191: Why is this needed?

lib/vdsm/virt/cpumanagement.py
```diff
@@ -160,6 +163,37 @@ def libvirt_cpuset_spec(cpus, cpu_list_length):
...
+    :param stored: str
+    """
+    if '_srcDomXML' not in stored or 'xml' not in from_engine:
+        return stored
+    dom_stored = xmlutils.fromstring(stored['_srcDomXML'])
+    dom_engine = xmlutils.fromstring(from_engine['xml'])
+    # Remove any previous pinning
+    cputune_stored = dom_stored.find('cputune')
+    if cputune_stored is not None:
+        for vcpu in vmxml.find_all(cputune_stored, 'vcpupin'):
```
mz-pdm, line 182: ``for vcpupin ...``?

lib/vdsm/virt/cpumanagement.py
```diff
@@ -160,6 +163,37 @@ def libvirt_cpuset_spec(cpus, cpu_list_length):
...
+            cputune_stored.remove(vcpu)
+    else:
+        cputune_stored = ET.Element('cputune')
+        dom_stored.append(cputune_stored)
+    # Move pinning we have from engine
+    cputune_engine = dom_engine.find('cputune')
+    if cputune_engine is not None:
+        for vcpu in vmxml.find_all(cputune_engine, 'vcpupin'):
```
mz-pdm, line 190: ``for vcpupin ...``?

lib/vdsm/API.py
```diff
@@ -214,7 +215,15 @@ def create(self, vmParams):
                             self.log.info('loaded pickledMachineParams: %s',
                                           pickledMachineParams)
                             self.log.debug('former conf: %s', vmParams)
-                            vmParams.update(pickledMachineParams)
+                            modified_former_params = \
+                                cpumanagement.resume_cpu_pinning(
+                                    vmParams,
+                                    pickledMachineParams)
+                            if modified_former_params != pickledMachineParams:
```
mz-pdm, line 222: resume_cpu_pinning returns pickledMachineParams, which is modified by side-effect, so this condition is never satisfied.

lib/vdsm/virt/cpumanagement.py
```diff
@@ -160,6 +163,37 @@ def libvirt_cpuset_spec(cpus, cpu_list_length):
...
+            cputune_engine.remove(vcpu)
+            cputune_stored.append(vcpu)
+    stored['_srcDomXML'] = xmlutils.tostring(dom_stored)
+    return stored
```
mz-pdm, line 194: Since `stored` is modified by side-effect, it's just confusing to return it.  I'd suggest either not to return anything or to return the new _srcDomXML without modifying `stored`.

nyoxi:
I have resolved all the conversation because the associated code is no longer there. If some questions are still relevant please again.

lib/vdsm/virt/cpumanagement.py
```diff
@@ -136,6 +138,27 @@ def _assign_shared(cif, target_vm=None):
...
+
+    :param cpusets_dict: Dictionary with CPU sets
+    :type cpusets_dict: dict
+
+    :returns: list of strings
+    """
+    if len(cpusets_dict.keys()) == 0:
+        return []
+    result = [None] * (max(cpusets_dict.keys()) + 1)
+    for id, cpuset in cpusets_dict.items():
```
mz-pdm, line 157: Better to use `id_` or `cpu_id` to distinguish it from `id` built-in.

lib/vdsm/virt/vm.py
```diff
@@ -2813,32 +2814,44 @@ def _run(self):
...
+                    # (e.g. 'vmName') is present, we use the legacy path here.
+                    # Otherwise we leave srcDomXML untouched, since we don't
+                    # have anything from Engine (4.2.0) to update it with.
+                    if 'vmName' in self.conf:
+                        srcDomXML = self._correct_disk_volumes_from_conf(
+                            srcDomXML)
+                    srcDomXML = self._correctGraphicsConfiguration(srcDomXML)
                 else:
-                    self._connection.restore(fname)
+                    srcDomXML = self._connection.saveImageGetXMLDesc(fname)
```
mz-pdm, line 2832: This deserves commentary in the commit message -- we use a different srcDomXML now.  Does it matter?  Can there be changes performed in `Vm.__init__` on it that could be lost here?

lib/vdsm/virt/vm.py
```diff
@@ -2813,32 +2814,44 @@
...
+                    # Modify CPU pinning -- remove pinning saved during
+                    # hibernation and replace it with pinning provided by
+                    # engine in API call.
+                    self.log.info("loaded: %s", srcDomXML)
```
mz-pdm, line 2836: The logging message could be a bit more descriptive about what we log here.

lib/vdsm/virt/vm.py
```diff
@@ -2813,32 +2814,44 @@
...
+                    dom = xmlutils.fromstring(srcDomXML)
+                    dom = cpumanagement.replace_cpu_pinning(
+                        self, dom, self._initial_vcpupin)
+                    srcDomXML = xmlutils.tostring(dom)
+                    self.log.info("modified: %s", srcDomXML)
```
mz-pdm, line 2841: Is this needed when we log it again below?

lib/vdsm/virt/cpumanagement.py
nyoxi: done

lib/vdsm/virt/vm.py
nyoxi: You are right, this indeed deserves a comment. I have updated the commit message. For visibility here it is too:

    Domain XML that is used to restore the VM is loaded from libvirt's
    hibernation image which replaces XML passed from API call and that goes
    through init. This should not lead to any change in behavior. While we
    used the passed XML to create the domain it has been always immediately
    replaced by libvirt in the resume call. That means if we are doing any
    changes to the XML during early init such changes were always lost and
    will still remain lost.

lib/vdsm/virt/vm.py
nyoxi: It remained there from development. I have removed the log line.

lib/vdsm/virt/vm.py
nyoxi: Same as the other one... removed.

lib/vdsm/virt/cpumanagement.py
mz-pdm: Needs an update also on the following line.

lib/vdsm/virt/vm.py
mz-pdm: So this means:

- The hook obtains a more correct XML.
- The content of the domain XML passed to defineXML doesn't matter.
- libvirt takes and updates CPU pinning from the second argument of restoreFlags call.

Right?  Did you verify the last two facts above (libvirt documentation isn't very clear in those areas)?

nyoxi: > So this means:
> 
>     * The hook obtains a more correct XML.

Yes

> 
>     * The content of the domain XML passed to defineXML doesn't matter.

Yes, that is my suspicion. I planned to investigate it little bit more but right know it seems to me the `defineXML` is superfluous and has no effect. The complete domain XML is then replaced with the XML libvirt has (or with what we pass in `restoreFlags`).

> 
>     * libvirt takes and updates CPU pinning from the second argument of restoreFlags call.

Yes. 

> 
> Right? Did you verify the last two facts above (libvirt documentation isn't very clear in those areas)?

The second one yes, the first one I still want to investigate more. But this probably does not need to block the PR.

It occurred to me I have omitted another vital piece of information so I have updated the commit message one more time. Again for visibility:

    It may be tempting to pass domain XML from engine or from our pickled data
    to `resumeFlags()`. Neither is possible because our domain XML is not
    migration XML. Doing so would end in error because libvirt would fail to
    match the guest specific portion with the stored domain XML.

lib/vdsm/virt/cpumanagement.py
nyoxi: right, what was I thinking...

lib/vdsm/virt/vm.py
mz-pdm: OK.  Please check whether there is any relation between the defineXML domain XML and restoreFlags ones (probably the best way is to ask on the libvirt users list).  But yes, we needn't block this patch on this.

mz-pdm (approved):

lib/vdsm/virt/vm.py
```diff
@@ -2813,32 +2814,42 @@ def _run(self):
...
+                    srcDomXML = self._correctGraphicsConfiguration(srcDomXML)
                 else:
-                    self._connection.restore(fname)
+                    srcDomXML = self._connection.saveImageGetXMLDesc(fname)
+                    # Modify CPU pinning -- remove pinning saved during
+                    # hibernation and replace it with pinning provided by
+                    # engine in API call.
+                    dom = xmlutils.fromstring(srcDomXML)
+                    dom = cpumanagement.replace_cpu_pinning(
+                        self, dom, self._initial_vcpupin)
```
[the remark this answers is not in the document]
nyoxi: Resume from snapshot fixed in #251.

PR #216, opened 2022-05-27
lvm: error handling in reloadpvs consistent with VGs

aesteve-rh:
`LVMCache._reloadvgs` has been recently refactored for better error handling.

This branch extends the same strategy to `reloadpvs`:

- Refactor the method, according to its function:
  - Run LVM command
  - Try to reload stale PVs
  - Update all PVs
- Split reload into single and multi PVs
- Raise exceptions for missing volumes from the LVMCache, only when looking for specific volumes
  - Callers should get either a valid volume or propagate an exception

### DIFF
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -566,13 +566,29 @@ class LVMCache(object):
 
         return updatedPVs
 
+    def _reload_single_pv(self, pv_name):
+        """
+        Run LVM 'pvs' command and update PV pv_name.
+        """
+        cmd = list(PVS_CMD)
+        cmd.append(pv_name)
+
+        out, error = self.run_command_error(cmd)
+
+        with self._lock:
+            if error:
+                return self._update_stale_pvs_locked([pv_name])
+
+            return self._updatepvs_locked(out, [pv_name])
 
     def _reloadpvs(self, pvName=None):
-        cmd = list(PVS_CMD)
-
+        """
+        Run LVM 'pvs' command and update all PVs from pvNames.
+        If no PV name is provided, reload all PVs.
+        """
         pvNames = normalize_args(pvName)
-        if pvNames:
-            cmd.extend(pvNames)
+        cmd = list(PVS_CMD)
+        cmd.extend(pvNames)
 
         out, error = self.run_command_error(cmd)
 
@@ -889,7 +905,7 @@ class LVMCache(object):
         pv = self._pvs.get(pvName)
         if not pv or pv.is_stale():
             self.stats.miss()
-            pvs = self._reloadpvs(pvName)
+            pvs = self._reload_single_pv(pvName)
             pv = pvs.get(pvName)
         else:
             self.stats.hit()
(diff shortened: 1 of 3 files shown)


lib/vdsm/storage/lvm.py
```diff
@@ -529,42 +529,42 @@ def _updatepvs_locked(self, pvs_output, pv_names):
...
         """
         Check if any PV is stale after an LVM command and make it Unreadable.
         Log a warning if any PV has been made Unreadable.
         Must be called while holding the lock.
         """
-        updatedPVs = {}
-        pvNames = pv_names if pv_names else self._pvs
-        for p in pvNames:
+        updated_pvs = {}
+        pv_names = pv_names if pv_names else self._pvs
```
nirs, line 551: Before this change pVNames was a new variable, now it replaces pvNames. Since pvNames is either list of names or a dict, this may change pv_names to a dict. This works for iteration but may break the code in the rest of function.

It is better to avoid such manipulations, keep pv_names as list of strings, and use another name for the itetarion.

An easy way to avoid this would be:

   for p in (pv_names or self._pvs):

lib/vdsm/storage/lvm.py
```diff
@@ -592,28 +592,28 @@ def _reload_single_pv(self, pv_name):
 
             return updated_pvs[pv_name]
 
-    def _reloadpvs(self, pvName=None):
+    def _reloadpvs(self, pv_name=None):
```
nirs, line 595: Note for future change - using pv_name for None, single pv name or list of names is 
horrible!

We need to get rid of this by always calling with a list, and call this pv_names.

lib/vdsm/storage/lvm.py
```diff
@@ -950,16 +950,16 @@ def getPvs(self, vgName):
...
             if pv is None or pv.is_stale():
-                stalepvs.append(pvName)
+                stalepvs.append(pv_name)
             else:
                 pvs.append(pv)
 
         if stalepvs:
             self.stats.miss()
-            reloadedpvs = self._reloadpvs(pvName=stalepvs)
+            reloadedpvs = self._reloadpvs(pv_name=stalepvs)
```
nirs, line 962: Note for future patch - this is evil!

lib/vdsm/storage/lvm.py
```diff
@@ -1277,12 +1277,12 @@ def _lvs_proc_info(vg, lvs):
...
 
 
-def getPV(pvName):
+def getPV(pv_name):
     """
     Return PV named tuple. Raise se.InaccessiblePhysDev if the
     PV does not exist.
     """
-    return _lvminfo.getPv(_fqpvname(pvName))
+    return _lvminfo.getPv(_fqpvname(pv_name))
```
nirs, line 1285: The issue already exists, but we should really not use the same name for different things.

The function argument may be a wwid (e.g. xxxyyy). _fqpvname(pv_name) returns a path
(e.g. /dev/mapper/xxxyyy). We need to use consistent names for different things in the entire code base.

It is possible that getPV() works both with path and wwid - this is another issue.
I think the right way to handle this is to accept only path - same value used by lvm when creating pvs.

If the system uses wwids to access pvs in some place, the caller need to create path from the wwid. This layer should not know anything about wwids, only paths.

nirs (changes requested):
Looks good, expect one issue caused by renaming to snake_case.

nirs (approved):
Awesome, thanks!

PR #218, opened 2022-05-30
storage.exceptions: add generic class to except with an LVM error

aesteve-rh:
`VolumeGroupDoesNotExist` exception have been recently modified so that it can hold an optional `LVMCommandError` and print the attributes in an informative manner.

Take those changes to a generic class (i.e., `_HoldingLVMCommandError`) so that we can inherit from it and have the error attribute, parameter verification and consistent logging across all classes that may need to hold an LVMCommandError.

This PR also changes `LogicalVolumeDoesNotExists` to inherit from the new generic exception to hold the LVMCommandError.

Tests have been updated accordingly.

Depends on: https://github.com/oVirt/vdsm/pull/216

### DIFF
--- a/lib/vdsm/storage/exception.py
+++ b/lib/vdsm/storage/exception.py
@@ -1628,9 +1628,18 @@ class MissingTagOnLogicalVolume(StorageException):
     msg = "Missing logical volume tag."
 
 
-class LogicalVolumeDoesNotExistError(StorageException):
+class LogicalVolumeDoesNotExistError(_HoldingLVMCommandError):
     code = 610
     msg = "Logical volume does not exist"
+
+    def __init__(self, vg_name, lv_name, error=None):
+        super().__init__(error)
+        self.vg_name = vg_name
+        self.lv_name = lv_name
+
+    @classmethod
+    def from_error(cls, vg_name, lv_name, error):
+        return cls(vg_name, lv_name, error=error).with_exception(error)
 
 
 class LogicalVolumeCachingError(StorageException):
--- a/tests/storage/exception_test.py
+++ b/tests/storage/exception_test.py
@@ -88,6 +88,25 @@ def test_info():
         }
 
 
+def test_LogicalVolumeDoesNotExistError():
+    # Expected error type is LVMCommandError.
+    with pytest.raises(TypeError):
+        e = storage_exception.LogicalVolumeDoesNotExistError(
+            "vg-name", "lv-name", error="error")
+
+    # Correct initialization.
+    fake_error = storage_exception.LVMCommandError(
+        rc=5, cmd=["fake"], out=["fake output"], err=["fake error"])
+    e = storage_exception.LogicalVolumeDoesNotExistError(
+        "vg-name", "lv-name", error=fake_error)
+    assert e.error == fake_error
+    # Check error format
+    formatted = str(e)
+    assert "vg_name=vg-name" in formatted
+    assert "lv_name=lv-name" in formatted
+    assert "error=" in formatted
+
+
 def test_VolumeGroupDoesNotExist():
     # Require a VG name or UUID at initialization.
     # Empty constructor shall raise.
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -1409,7 +1409,7 @@ def getLV(vgName, lvName=None):
     lv = _lvminfo.getLv(vgName, lvName)
     # getLV() should not return None
     if not lv:
-        raise se.LogicalVolumeDoesNotExistError("%s/%s" % (vgName, lvName))
+        raise se.LogicalVolumeDoesNotExistError(vgName, lvName)
     else:
         return lv
 
--- a/tests/storage/storagefakelib.py
+++ b/tests/storage/storagefakelib.py
@@ -269,7 +269,7 @@ class FakeLVM(object):
         try:
             lv = self.lvmd[(vgName, lvName)]
         except KeyError:
-            raise se.LogicalVolumeDoesNotExistError("%s/%s" % (vgName, lvName))
+            raise se.LogicalVolumeDoesNotExistError(vgName, lvName)
         lv_md = deepcopy(lv)
         lv_attr = real_lvm.LV_ATTR(**lv_md['attr'])
         lv_md['attr'] = lv_attr


lib/vdsm/storage/exception.py
```diff
@@ -115,6 +115,22 @@ def __init__(self, volume, *args):
             self.value += list(args)
 
 
+class VolumeLVMCommandError(StorageException):
```
nirs, line 118: Not sure that Volume is relevant - this should be used for vgs, pvs, and lvs.
I think the common property for this base class is holding a LVMCommandError
and formatting the error using key=value format.

So maybe call it _HoldingLVMCommandError?

This makes it clear that this is not meant to be created or raise, but only
used a a base class.

lib/vdsm/storage/exception.py
```diff
@@ -115,6 +115,22 @@ def __init__(self, volume, *args):
...
+    code = 4001
+    msg = "Volume exception with LVM error"
```
nirs, line 120: This error is not meant to be used directly, only used as a base class, so it does
not need code and msg - the inheriting class should add them.

lib/vdsm/storage/exception.py
```diff
@@ -115,6 +115,22 @@ def __init__(self, volume, *args):
...
+
+    def __init__(self, error=None):
+        if error is not None and not isinstance(error, LVMCommandError):
+            raise TypeError(
+                f"Expecting instance of LVMCommandError, got {type(error)}")
+
+        self.error = error
```
nirs, line 127: Because we use the dict order, the order of the parameter depends now on the way
the sub class is implemented. For example:

    class ErrorIsFirst(VolumeLVMCommandError):
    
        def __init__(self, name, error=None):
            super().__init__(error)
            self.name = name

    class ErrorIsLast(VolumeLVMCommandError):
    
        def __init__(self, name, error=None):
            self.name = name
            super().__init__(error)

This is not that bad, but kind of surprising. I think it is better if we
control the formatting so it is not effected by the way the class is initialized.

lib/vdsm/storage/exception.py
```diff
@@ -115,6 +115,22 @@ def __init__(self, volume, *args):
...
+
+    @property
+    def value(self):
+        return ", ".join(f"{k}={v}" for k, v in self.__dict__.items() if v)
```
nirs, line 131: Nice that we have one place to format instead of repeating this in 3 classes.

It would be nice if we use this for all storage errors in the future.

lib/vdsm/storage/exception.py
```diff
@@ -1345,25 +1361,18 @@ class VolumeGroupAlreadyExistsError(StorageException):
...
     msg = "Volume Group does not exist"
 
     def __init__(self, vg_name=None, vg_uuid=None, error=None):
         if vg_name is None and vg_uuid is None:
             raise ValueError("Require a VG name or UUID")
 
-        if error is not None and not isinstance(error, LVMCommandError):
-            raise TypeError(
-                f"Expecting instance of LVMCommandError, got {type(error)}")
+        super().__init__(error)
```
nirs, line 1372: So here error will be logged first, but previously we logged it last.

lib/vdsm/storage/exception.py
```diff
@@ -1345,25 +1361,18 @@ class VolumeGroupAlreadyExistsError(StorageException):
...
 
         self.vg_name = vg_name
         self.vg_uuid = vg_uuid
-        self.error = error
```
nirs, line 1362: If we init the super class here, we keep the previous behavior.

tests/storage/exception_test.py
```diff
@@ -59,8 +59,12 @@ def test_collisions():
 
 
 def test_info():
+    skipped_exceptions = [
+        storage_exception.VolumeLVMCommandError,
+        storage_exception.VolumeGroupDoesNotExist,
+    ]
     for obj in find_module_exceptions(storage_exception, VdsmException):
-        if obj == storage_exception.VolumeGroupDoesNotExist:
+        if obj in skipped_exceptions:
```
nirs, line 68: We can change the filter to skip subclasses of _HoldingLVMCommandError

lib/vdsm/storage/exception.py
```diff
@@ -1622,10 +1622,21 @@ def __init__(self, lvname, tag):
...
+class LogicalVolumeDoesNotExistError(VolumeLVMCommandError):
     code = 610
     msg = "Logical volume does not exist"
 
+    def __init__(self, vg_name, lv_name, error=None):
+        super().__init__(error)
+
+        self.vg_name = vg_name
+        self.lv_name = lv_name
+        self.error = error
```
nirs, line 1634: If we init error in the superclass, why do we override it here?

aesteve-rh: agh, forgot to remove it (comes from another commit that had it).

lib/vdsm/storage/exception.py
aesteve-rh: I like the name, but the underscore at the beginning may be misinterpreted, I think. I would generally consider a class starting with an underscore to be internal (not to be imported), but not abstract.

Maybe we can let the code be more clear with it and overwrite the `__new__` method to check for the `cls` type, as in:
```python
    def __new__(cls, *args, **kwargs):
        # Base class cannot be directly instantiated.
        if cls is HoldingLVMCommandError:
            raise TypeError(
                f"{cls.__name__} exception is not meant to be directly instantiated.")
        return object.__new__(cls, *args, **kwargs)
```

nirs: I don't see the need for this complication. Using _Prefix makes it clear that you should not use this class outside of the module. We can assume that oVirt developers do not use private stuff from other modules.

lib/vdsm/storage/exception.py
nirs: We can fix find_exceptions() to skip private classes (another reason to use _prefix).

nirs: > Or maybe make the base class not inherit from anything and use multiple inheritance in the child classes.

Multiple inheritance when you have to `__init__` multiple super class does not sound promising. Lets try to keep it simple.

lib/vdsm/storage/exception.py
aesteve-rh: Sounds good. I'll create an issue.

I have a few notes taken that need to be written into issues...

lib/vdsm/storage/exception.py
aesteve-rh: I have fixed it with the value property, printing error always at the end.

It was a side effect I didn't see.

lib/vdsm/storage/exception.py
```diff
@@ -1589,23 +1597,14 @@ class CouldNotRetrieveLogicalVolumesList(StorageException):
...
-        self.error = error
+        super().__init__(error)
 
-    @property
-    def value(self):
-        ret_value = f"devices={self.devices}"
-        if self.error:
-            ret_value += f", error={self.error}"
-        return ret_value
+        self.devices = devices
```
nirs, line 1607: We can remove the empty line between super().__init__() and self.devices =

tests/storage/exception_test.py
```diff
@@ -43,6 +43,10 @@ def find_module_exceptions(module, base_class=None):
         if base_class and not issubclass(obj, base_class):
             continue
 
+        # Internal classes (starting with '_' by convention) are skipped.
```
nirs, line 46: "(starting ...)" is not needed, people know that names starting with _ are private.

nirs (changes requested):

tests/storage/exception_test.py
```diff
@@ -106,7 +106,8 @@ def test_VolumeGroupDoesNotExist():
     # Check error format
     formatted = str(e)
     assert "vg_name=vg-name" in formatted
-    assert f'error={fake_error}'.replace('\'', r'\'') in formatted
+    error_str = f'error={fake_error}'.replace('\'', r'\'') + '\''
```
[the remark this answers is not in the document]
aesteve-rh: OK, I'll change it. I agree that it was annoying to match and ugly. And this should suffice for the verification. It was nice mainly due to the `endswith`, but does not add that much verification value.

PR #222, opened 2022-05-31
Testing OST with master

nirs:
Empty change for testing OST with master.


michalskrivanek:
What's the point? master is being tested every 120 minutes in a loop

nirs:
> What's the point? master is being tested every 120 minutes in a loop

I did not this, but even if it does, I don't have a way to check the result
or know when it will run.

This way a contributor can run OST and prove that the failure is not relevant
to the pr that fails OST.

PR #226, opened 2022-06-01
virt: Fix long VNC passwords in vm_libvirt_hook.py

mz-pdm:
Newer versions of libvirt accept VNC passwords of the maximum length 8,
because QEMU uses only the first 8 characters anyway.  We have already
fixed the password length in Engine but if we migrate VMs created by
older Engines they may fail to start on the destination due to the
long VNC password.

Let’s make those VMs migratable to newer hosts by fixing the password
in the libvirt hook.  We can simply remove the extra unused characters
from the password to make libvirt happy.

Bug-Url: https://bugzilla.redhat.com/2090156

### DIFF
--- a/lib/vdsm/virt/libvirthook/vm_libvirt_hook.py
+++ b/lib/vdsm/virt/libvirthook/vm_libvirt_hook.py
@@ -1,6 +1,6 @@
 #!/usr/bin/python3
 
-# Copyright 2016-2019 Red Hat, Inc.
+# Copyright 2016-2022 Red Hat, Inc.
 #
 # This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


nirs (approved):
Horrible, but seems that we have no other way :-)

dupondje:
We can't lookup the password somewhere and reset it on upgrade?
As the passwords can be changed live

ahadas (approved):

mz-pdm:
> We can't lookup the password somewhere and reset it on upgrade? As the passwords can be changed live

Responded in BZ (briefly, the solution in this patch is much safer and easier).

nyoxi (approved):

liranr23 (approved):

PR #231, opened 2022-06-04
spec: Use COPR versioning in COPR builds

tinez:
Our COPR builds are not distinguishable on a per-commit basis - they all
end up having the same version, i.e.:

 4.50.1.1-1.fc35

Because of this the newer versions are never picked up by OST.
Let's align our versioning with the rest of the projects and use
the '%{?release_suffix}' variable when it's defined.


michalskrivanek:
#232 fixed it

PR #232, opened 2022-06-04
 Mark temp repo directory as safe for COPR

michalskrivanek:
When building from COPR the project is cloned into temporary directory,
which is not owned by current user. From git 2.35.2 such directory needs
to be marked as safe inn order git commands work correctly.


michalskrivanek:
in other projects buiklds fail, here it succeeded, but without actually running the desired copr makefile target.

PR #233, opened 2022-06-05
Vagrant based infrastructure

nirs:
Trying to create vagrant based infrastructure for running tests in VMs.

This is required for some storage tests that cannot work in containers, and 
for tests that need to run as root.

With the current image, one can run the storage tests in an almalinux-8 VM
using ready to use vagrant box.


.gitignore
```diff
@@ -83,6 +83,7 @@ static/usr/share/man/man8/vdsmd.8
 vdsm/rpc/vdsm-api.html
 vdsm_hooks/openstacknet/sudoers
 vdsm_hooks/checkips/vdsm-checkips.service
+.vagrant
```
aesteve-rh, line 86: Hidden files appear at the beginning of the list. I would expect this right after `.tox`.

michalskrivanek:
Why do we need yet another test solution. We already have OST. Creating another infra is counterproductive

nirs:
> Why do we need yet another test solution. We already have OST. Creating another infra is counterproductive

I don't think OST solves the same problem. Our problem is running the tests in containers
which cannot work for all tests, which means that tests do not run or worse run and give
wrong results.

We need to run tests in vms, and vagrant gives us easy way to share vm images and manage
them. As you can see this very little code is needed to use it.

michalskrivanek:
and very little code is needed for that in OST. you can see similar approach with go-ovirt-client, though that require and take advantage of oVirt deployment too.
Getting into Vagrant is just diverging efforts, we have OST plugged into CI, we won't do any such thing with vagrant or others, if you miss anything then improve OST instead

also, if this is not intrusive it would run well enough alongside the current full suite test

aesteve-rh:
I didn't know Vagrant, so I have jumped into the branch to test. And it is really nice! It really does everything for you, setting up the vm and installing everything with the shell script. It even allows to easily establish an ssh connection without any extra configuration. I mean, I basically only read the quickstart section of the docs, and it was enough to run a vm and connect using the Vagrant boxes in this PR.

The code is easy to read, and update, and thus, should be easy to support. For testing in the CIs we should probably stick with OST, that is true, it doesn't make sense to duplicate efforts. However, I agree with Nir in that this may solve a different issue, when it comes to managing and maintaining different vms for local testing. This makes sharing VMs easy, and gives a homogeneous playground for local testing to all developers.

I wonder how exactly we would do the testing, though. Would it use the shared folder between host and guest systems to obtain the project tests and run them?

nirs:
I pushed what I have locally as is to others can test the current code. For merging can can squash some changes together.

michalskrivanek:
It's extremely unlikely that we would ever use vagrant anywhere in oVirt. We evaluated twice in the past, we concluded it's not viable.
All CI testing is OST-based and this just creates yet another variant, yet another source of "little differences" that people spend hours to chase.
OST infrastructure has everything that you're trying to create here. No one will ever support this. You should have just run tests within OST either as part of a suite for merged stuff, or as a separate one that just runs vdsm unit tests for GH CI. And of course you can then run that yourself for your local testing, same as with vagrant.

nirs:
I don't have time to work on this.

PR #234, opened 2022-06-06
lvm: disable vg autoactivation after create

### LINKED ISSUE #229: lvm: properly disabling setautoactivation for VGs
In LVM configuration, `event_activation=1` is the rhel8 default, and it is generally discouraged disabling it in rhel8 because it's not reliable.

Vdsm had used event_activation=0 to disable autoactivation of VGs, but it was a wrong approach. It may only appear to work sometimes, based on timing of device attachment.

We should disable autoactivation of VGs with `vgchange --setautoactivation n`

aesteve-rh:
Disable setautoactivation attribute of a VG after
its creation in order to avoid unwanted VG activation.

Fixes #229


lib/vdsm/storage/lvm.py
```diff
@@ -1439,6 +1439,26 @@ def createVG(vgName, devices, initialTag, metadataSize, force=False):
         _lvminfo._invalidatevgs(vgName)
         log.debug("Cache after createvg %s", _lvminfo._vgs)
 
+    # Disable autoactivation after creating a VG
+    _disable_autoactivation_vg(vgName)
+
+
+def _disable_autoactivation_vg(vg_name):
+    """Avoid undesired VG activation by disabling autoactivation attribute
+    on a given VG.
```
nirs, line 1448: We always use this style:

    """
    Title ...

We need more info here on how disabling autoactivation works. How does it 
protect from activation? Which lvm command does autoactivation?

I think adding a test like this will make it more clear:
1. create and configure a new vg
2. create some lvs
3. run lvm commands that does auto activation
4. verify that no lv was activated

Other tests creating and activating lvs already verify lv activation flows.

lib/vdsm/storage/lvm.py
```diff
@@ -1439,6 +1439,26 @@ def createVG(vgName, devices, initialTag, metadataSize, force=False):
         _lvminfo._invalidatevgs(vgName)
         log.debug("Cache after createvg %s", _lvminfo._vgs)
 
+    # Disable autoactivation after creating a VG
+    _disable_autoactivation_vg(vgName)
```
nirs, line 1443: Doing it here makes sense, this is part of vg creation flow, and it should happen
before we create any lv on this vg.

The comment is not needed, it does not add any value. It can add value if
it will describe why this is done at this point (before any lv can be created).

There is an issue with error handling - if creating the vg succeeded, but disabling
autoactivation failed, we raise se.VolumeGroupCreateError. From caller point of view
creating the vg failed, so it may not try to clean up after the failure, and we are
left with lefover vg using the pv. This may break engine create vg flow and require
manual cleanup.

Please check if this can be integrated in a higher level, so if creating and 
configuring a vg fails no leftovers are left.

lib/vdsm/storage/lvm.py
```diff
@@ -1439,6 +1439,26 @@ def createVG(vgName, devices, initialTag, metadataSize, force=False):
...
+
+
+def _disable_autoactivation_vg(vg_name):
```
nirs, line 1446: _disable_vg_autoactivation?

lib/vdsm/storage/lvm.py
```diff
@@ -1439,6 +1439,26 @@ def createVG(vgName, devices, initialTag, metadataSize, force=False):
...
+
+    Args:
+        vg_name (str): Name of the VG
+
+    Raises:
+        se.VolumeGroupCreateError: LVM error occurred
+    """
+    cmd = ["vgchange", "--setautoactivation", "n", vg_name]
+    try:
+        _lvminfo.run_command(cmd)
```
nirs, line 1458: We need to use the vg devices in this command, so lvm does not try to look at other PVs. See how other commands use devices=.

nirs (changes requested):
This change try to handle creation of existing vg, but we need to handle also upgrade of existing vgs.

We may need a way to get the autoactivation property of a vg and store it in the cache, and decide when we check and update existing vgs so all vgs in a system have the --autoactivation "n" property.

lib/vdsm/storage/lvm.py
aesteve-rh: Ok, I'll mark the PR as draft for now and do some checks to see how can I test this.

I'll re-request the review when ready :)

aesteve-rh:
I have closed all comments in the PR because I realized that the option was also available in the `vgcreate` command, which naturally avoids the problems of having to clean leftover volumes, adding another top level function, etc.

The change is now a one-liner. Added a test to verify and exemplify the autoactivate issue.

vjuranek (approved):

lib/vdsm/storage/lvm.py
```diff
@@ -1479,7 +1479,8 @@ def createVG(vgName, devices, initialTag, metadataSize, force=False):
     _initpvs(pvs, metadataSize, force)
     _enable_metadata_area(pvs)
 
-    options = ["--physicalextentsize", "%dm" % (sc.VG_EXTENT_SIZE // MiB)]
+    options = ["--setautoactivation", "n",
+               "--physicalextentsize", "%dm" % (sc.VG_EXTENT_SIZE // MiB)]
```
nirs, line 1483: Nice! adding the option when creating the vg makes this atomic.

tests/storage/lvm_test.py
```diff
@@ -1669,6 +1669,35 @@ def test_vg_stale_reload_all_clear(stale_vg):
...
+    lvm.createLV(vg_name, lv2_name, 1024, activate=False)
+
+    # LVM command to autoactivate the LVs, but is disabled for the VG.
+    lvm.changelv(vg_name, [lv1_name, lv2_name], ("-a", "ay"))
+    assert not lvm.getLV(vg_name, lv1_name).active
+    assert not lvm.getLV(vg_name, lv2_name).active
+
+    # Enable VG autoactivation.
+    commands.run(
+        ["vgchange", "--setautoactivation", "y", "--devices", dev, vg_name])
```
nirs, line 1693: We don't need to test this.

tests/storage/lvm_test.py
```diff
@@ -1669,6 +1669,35 @@ def test_vg_stale_reload_all_clear(stale_vg):
...
+
+    # Autoactivate is disabled for all new VGs.
+    lvm.createVG(vg_name, [dev], "initial-tag", 128)
+    lvm.createLV(vg_name, lv1_name, 1024, activate=False)
+    lvm.createLV(vg_name, lv2_name, 1024, activate=False)
+
+    # LVM command to autoactivate the LVs, but is disabled for the VG.
+    lvm.changelv(vg_name, [lv1_name, lv2_name], ("-a", "ay"))
+    assert not lvm.getLV(vg_name, lv1_name).active
+    assert not lvm.getLV(vg_name, lv2_name).active
```
nirs, line 1689: Good test, but we want to test also:

    vgchange -a ay vg-name

nirs (changes requested):

nirs:
@aesteve-rh please update the commit message title `lvm: disable vg autoactivation after create`. We don't disable *after* creating the vg, and even if we did this is implementation detail that does not belong int the commit message title.

aesteve-rh:
Thanks for pointing that out, I forgot to update the commit after changing the code. Hope it is better now!

PR #235, opened 2022-06-06
tool.multipath: warn when user_friendly_names is enabled

aesteve-rh:
This PR updates the isconfigured check for the multipath in the vdsm-tool,
so that running:

    # vdsm-tool is-configured --module multipath

Will print a WARN message if there is any device or section in the multipath daemon configuration that enables 'user_friendly_names', as in:

```
WARNING: Invalid configuration: 'user_friendly_names' is enabled in multipath configuration:
  defaults {
    key value
    user_friendly_names yes
  }
  device {
    user_friendly_names yes
    key value
  }
This configuration is not supported and may lead to storage domain corruption.
```

### DIFF
--- a/lib/vdsm/storage/mpathconf.py
+++ b/lib/vdsm/storage/mpathconf.py
@@ -297,6 +297,8 @@ overrides {
 log = logging.getLogger("storage.mpathconf")
 
 Metadata = namedtuple("Metadata", "revision,private")
+Section = namedtuple("Section", "name,children")
+Pair = namedtuple("Pair", "key,value")
 
 
 def configure_blacklist(wwids):
@@ -428,3 +430,33 @@ def read_metadata():
         return Metadata(REVISION_OLD, private)
 
     return Metadata(REVISION_MISSING, private)
+
+
+def _parse_conf(mpath_out_it):
+    """
+    Recursively parse multipath daemon configuration output.
+
+    Args:
+        mpath_out_it (iterator): String iterator containing the output of the
+            multipath configuration.
+
+    Returns:
+        Parsed configuration.
+    """
+    res = []
+    for line in mpath_out_it:
+        line = line.strip()
+        if line.endswith("{"):
+            name, _ = line.rsplit(" ", 1)
+            res.append(Section(name, _parse_conf(mpath_out_it)))
+        elif line == "}":
+            break
+        else:
+            if not line:
+                # Discard empty lines
+                continue
+            key, value = line.split(" ", 1)
+            if value[0] == value[-1] == '"':
+                value = value[1:-1]
+            res.append(Pair(key, value))
+    return res
--- a/lib/vdsm/storage/mpathconf.py
+++ b/lib/vdsm/storage/mpathconf.py
@@ -33,6 +33,8 @@ import re
 from collections import namedtuple
 from string import Template
 
+from vdsm.common import cmdutils
+from vdsm.storage import multipath
 from . import fileUtils
 
 CONF_FILE = '/etc/multipath.conf'
@@ -293,12 +295,14 @@ overrides {
 """).substitute({"current_tag": CURRENT_TAG,
                  "no_path_retry": _NO_PATH_RETRY})
 
+_SKIPPED_SECTIONS = ("blacklist", "blacklist_exceptions")
 
 log = logging.getLogger("storage.mpathconf")
 
 Metadata = namedtuple("Metadata", "revision,private")
 Section = namedtuple("Section", "name,children")
 Pair = namedtuple("Pair", "key,value")
+Result = namedtuple("Result", "error,issues")
 
 
 def configure_blacklist(wwids):
@@ -430,6 +434,58 @@ def read_metadata():
         return Metadata(REVISION_OLD, private)
 
     return Metadata(REVISION_MISSING, private)
+
+
+def check_local_config():
+    """
+    Check if 'user_friendly_names' is enabled in the multipathd configuration.
+    The configuration is obtained by running the multipathd command to show
+    the local configuration and capturing the output.
+    Parse the output and check that the contents are valid.
+
+    Returns:
+        vdsm.storage.mpathconf.Result
+    """
+    error = None
+    try:
+        mpath_conf = _parse_conf(io.StringIO(multipath.show_config_local()))
+        issues = _check_conf(mpath_conf)
+    except cmdutils.Error as e:
+        error = f"Cannot check multipath configuration: {e}"
+        issues = []
+    else:
+        if issues:
+            error = ("Invalid configuration: 'user_friendly_names' is "
+                     "enabled in multipath configuration")
+    return Result(error, issues)
+
+
+def _check_conf(mpath_conf):
+    """
+    Check that 'user_friendly_names' is disabled in all the children
+    of the parsed multipathd configuration.
+    Search is recursive.
+
+    Args:
+        mpath_conf: Parsed structure of the multipathd configuration output.
+
+    Returns:
+        List[Section]: Sections that have invalid configuration.
+    """
+    issues = []
+    for item in mpath_conf:
+        if not item.children or item.name in _SKIPPED_SECTIONS:
+            continue
+
+        if all(isinstance(child, Section) for child in item.children):
+            issues.extend(_check_conf(item.children))
+        elif all(isinstance(child, Pair) for child in item.children):
+            if Pair("user_friendly_names", "yes") in item.children:
+                issues.append(item)
+        else:
+            raise ValueError(f"Invalid item: {item}")
+
+    return issues
 
 
 def _parse_conf(mpath_out_it):
--- a/lib/vdsm/storage/multipath.py
+++ b/lib/vdsm/storage/multipath.py
@@ -209,6 +209,18 @@ def is_ready():
     return "busy: false" in status
 
 
+def show_config_local():
+    """
+    Obtain the local configuration of the multipath daemon.
+    Must run as root.
+
+    Returns:
+        str: Output of the multipathd show command.
+    """
+    return commands.run(
+        [_MULTIPATHD.cmd, "show", "config", "local"]).decode('ascii')
+
+
 def reconfigure():
     """
     Invoke multipathd to reconfigure the multipaths.
--- a/lib/vdsm/tool/configurators/multipath.py
+++ b/lib/vdsm/tool/configurators/multipath.py
@@ -76,11 +76,7 @@ def configure():
     multipath.reconfigure()
 
 
-def isconfigured():
-    """
-    Check the multipath daemon configuration.
-    """
-
+def _check_mpath_metadata():
     if os.path.exists(mpathconf.CONF_FILE):
         revision, private = mpathconf.read_metadata()
         if private:
@@ -109,3 +105,25 @@ def isconfigured():
 
     sys.stdout.write("multipath requires configuration\n")
     return NO
+
+
+def isconfigured():
+    """
+    Check the multipath daemon configuration.
+    """
+    conf_metadata_ok = _check_mpath_metadata()
+    if conf_metadata_ok == YES:
+        check_result = mpathconf.check_local_config()
+        if check_result.error is not None:
+            # Just warn for now, do not fail the isconfigured check here.
+            sys.stdout.write(f"WARNING: {check_result.error}:\n")
+            for section in check_result.issues:
+                sys.stdout.write(f"  {section.name} {{\n")
+                for attr in section.children:
+                    sys.stdout.write(f"    {attr.key} {attr.value}\n")
+                sys.stdout.write("  }\n")
+            if check_result.issues:
+                sys.stdout.write("This configuration is not supported and "
+                                 "may lead to storage domain corruption.\n")
+
+    return conf_metadata_ok
--- a/tests/storage/mpathconf_test.py
+++ b/tests/storage/mpathconf_test.py
@@ -246,3 +246,143 @@ section5 {
             ])
         ]),
     ]
+
+
+def test_mpathd_ufn_enabled_single():
+    fake_conf = [
+        Section('defaults', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('devices', [
+            Section('device', [
+                Pair('user_friendly_names', 'no'),
+                Pair('key', 'value'),
+            ]),
+            Section('device', [
+                Pair('key', 'value')
+            ])
+        ]),
+    ]
+    issues = mpathconf._check_conf(fake_conf)
+    assert issues == [
+        Section('defaults', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'yes'),
+        ])
+    ]
+
+
+def test_mpathd_ufn_enabled_multiple():
+    fake_conf = [
+        Section('defaults', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('overrides', [
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('devices', [
+            Section('device', [
+                Pair('user_friendly_names', 'yes'),
+                Pair('key', 'value'),
+            ]),
+            Section('device', [
+                Pair('key1', 'value1'),
+                Pair('user_friendly_names', 'yes'),
+                Pair('key2', 'value2'),
+            ]),
+            Section('device', [
+                Pair('key', 'value')
+            ])
+        ]),
+    ]
+    issues = mpathconf._check_conf(fake_conf)
+    assert issues == [
+        Section('defaults', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('overrides', [
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('device', [
+            Pair('user_friendly_names', 'yes'),
+            Pair('key', 'value'),
+        ]),
+        Section('device', [
+            Pair('key1', 'value1'),
+            Pair('user_friendly_names', 'yes'),
+            Pair('key2', 'value2'),
+        ]),
+    ]
+
+
+def test_mpathd_ufn_blacklist():
+    fake_conf = [
+        Section('defaults', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'no'),
+        ]),
+        Section('blacklist', [
+            Pair('key', 'value'),
+            Section('name1', [
+                Pair('key1', 'value1'),
+                Pair('key2', 'value2'),
+            ]),
+            Section('name2', [
+                Pair('key', 'value')
+            ])
+        ]),
+        Section('blacklist_exceptions', [
+            Pair('key', 'value'),
+            Section('name3', [
+                Pair('key1', 'value1'),
+                Pair('key2', 'value2'),
+            ]),
+            Section('name4', [
+                Pair('key', 'value')
+            ])
+        ]),
+    ]
+    issues = mpathconf._check_conf(fake_conf)
+    assert not issues
+
+
+def test_mpathd_ufn_no_pairs():
+    fake_conf = [
+        Section('section1', [])
+    ]
+    issues = mpathconf._check_conf(fake_conf)
+    assert not issues
+
+
+def test_mpathd_ufn_empty():
+    fake_conf = []
+    issues = mpathconf._check_conf(fake_conf)
+    assert not issues
+
+
+def test_mpathd_ufn_all_disabled():
+    fake_conf = [
+        Section('defaults', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'no'),
+        ]),
+        Section('overrides', [
+            Pair('key1', 'value1'),
+            Pair('user_friendly_names', 'no'),
+            Pair('key2', 'value2'),
+        ]),
+        Section('devices', [
+            Section('device', [
+                Pair('user_friendly_names', 'no'),
+                Pair('key', 'value'),
+            ]),
+            Section('device', [
+                Pair('key', 'value')
+            ])
+        ]),
+    ]
+    issues = mpathconf._check_conf(fake_conf)
+    assert not issues
--- a/tests/storage/mpathconf_test.py
+++ b/tests/storage/mpathconf_test.py
@@ -18,9 +18,12 @@
 #
 # Refer to the README and COPYING files for full details of the license
 #
+import io
+
 import pytest
 
 from vdsm.storage import mpathconf
+from vdsm.storage.mpathconf import Section, Pair
 
 from . marks import requires_selinux
 
@@ -172,3 +175,74 @@ def test_mpath_empty_conf(fake_mpath_conf):
         revision=mpathconf.REVISION_MISSING,
         private=False,
     )
+
+
+def test_mpathd_parser():
+    mpath_out = """\
+section1 {
+    key1 "value 1"
+    key2 value2
+    key3 3
+}
+section2 {
+    key1 value1
+}
+section3 {
+}
+section4 {
+    entry {
+        key1 "value 1"
+        key2 "value 2"
+        key3 "value 3"
+    }
+    entry {
+        key1 "value1"
+    }
+    entry {
+    }
+}
+section5 {
+    key1 value1
+    key2 value2
+    entry {
+        key1 value1
+        key2 value2
+    }
+    entry {
+        key1 value1
+    }
+}"""
+    assert mpathconf._parse_conf(io.StringIO(mpath_out)) == [
+        Section('section1', [
+            Pair('key1', 'value 1'),
+            Pair('key2', 'value2'),
+            Pair('key3', '3')
+        ]),
+        Section('section2', [
+            Pair('key1', 'value1')
+        ]),
+        Section('section3', [
+        ]),
+        Section('section4', [
+            Section('entry', [
+                Pair('key1', 'value 1'),
+                Pair('key2', 'value 2'),
+                Pair('key3', 'value 3')
+            ]),
+            Section('entry', [
+                Pair('key1', 'value1')
+            ]),
+            Section('entry', [])
+        ]),
+        Section('section5', [
+            Pair('key1', 'value1'),
+            Pair('key2', 'value2'),
+            Section('entry', [
+                Pair('key1', 'value1'),
+                Pair('key2', 'value2')
+            ]),
+            Section('entry', [
+                Pair('key1', 'value1')
+            ])
+        ]),
+    ]


aesteve-rh:
This is probably worth a test. I have to check how the tool modules are tested, I took a brief look at it last week.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -327,44 +327,60 @@ def configure():
...
+
+    Args:
+        mpathconf (List[str]): Lines in the multipath daemon
+        configuration file.
+
+    Returns:
+        bool: check has passed.
+    """
+    first = mpathconf[0]
+    second = mpathconf[1]
```
nirs: Letting IndexError propagate here is not great. We don't want that the caller will have to know that we raise IndexError, and this can hide real error in other part of the function.

The error should be handled here, and we need to log a good error and return the right value to the caller.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -327,44 +327,60 @@ def configure():
     commands.run([_MULTIPATHD.cmd, "reconfigure"])
 
 
-def isconfigured():
+def _check_conf_rev(mpathconf):
```
nirs: We need more clear names like _check_conf_revision()

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -327,44 +327,60 @@ def configure():
...
     if os.path.exists(_CONF_FILE):
-        first = second = ''
         with open(_CONF_FILE) as f:
             mpathconf = [x.strip("\n") for x in f.readlines()]
         try:
-            first = mpathconf[0]
-            second = mpathconf[1]
+            conf_check = _check_conf_rev(mpathconf)
+            return YES if conf_check else NO
         except IndexError:
```
nirs: Do we want to fail on the first check or do all the checks and report all errors?
Reporting all errors at once may be more user friendly but harder to get right.

If we want to check all, and keep the results, lets use more clear names:

    revision_ok = _check_revision(mpathconf)
    user_friendly_names_ok = _check_user_friendly_names(mpathconf)

    report errors...

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -327,22 +327,20 @@ def configure():
...
     The multipath configuration file should contain a tag in form
     "RHEV REVISION X.Y" for this check to succeed.
     If the tag above is followed by tag "RHEV PRIVATE" the configuration
     should be preserved at all cost.
 
     Args:
-        mpathconf (List[str]): Lines in the multipath daemon
-        configuration file.
+        first (str): First line in the multipath daemon configuration file.
+        second (str): Second line in the multipath daemon configuration file.
```
nirs, line 339: Passing here the first lines of the file is not an improvement. Please keep the code as is, passing the contents of the file as lines.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -327,22 +327,20 @@ def configure():
...
 
     Returns:
         bool: check has passed.
     """
-    first = mpathconf[0]
-    second = mpathconf[1]
```
nirs: Please keep this, and handle the IndexError here.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -372,15 +370,13 @@ def isconfigured():
     """
     Check the multipath daemon configuration at /etc/multipath.conf.
     """
-
-    if os.path.exists(_CONF_FILE):
+    if os.path.exists(_CONF_FILE) and os.stat(_CONF_FILE).st_size > 0:
```
nirs, line 433: Empty /etc/multipath.conf is not a valid configuration and we should not keep special code to handle it.

The file cannot be empty, and must contain the "RHEV REVISION X.Y" and "RHEV PRIVATE"
lines. If not the configuration is invalid and vdsm should not start!

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -372,15 +370,13 @@ def isconfigured():
...
-        try:
-            conf_check = _check_conf_rev(mpathconf)
-            return YES if conf_check else NO
-        except IndexError:
-            pass
+            first, second = [x.strip("\n")
+                             for x in (f.readline() for _ in range(2))]
+
+        conf_check = _check_conf_rev(first, second)
+        return YES if conf_check else NO
```
nirs, line 441: This code should be kept as is, just remove the error handler since _check_conf_revision() should handle it.

    if os.path.exists(_CONF_FILE):
        with open(_CONF_FILE) as f:
            mpathconf = [x.strip("\n") for x in f.readlines()]
            revision_ok = _check_revision(first, second)
            ...

nirs (changes requested):

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: Uhm, afterwards I change it for `readline()` that avoids the exception. I'll reorder the patches so that it happens first, and we don't need to worry about the exception at all.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: OK. But it can be a list with two elements so that we can do:

     first, second = mpathconf

and not raise (with `readline()`).

nirs: But mpathconf should be the entire file. The caller should not do the work for the function.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: And it does. If the file exists **and has contents**, we verify contents, otherwise, we fail.

The only thing that changed is the "and" part.

nirs:
> This check may be run when vdsm starts (in a hook?).

All configurations run on every vdsm starts via "vdsm-tool is-configured".
If the configurator returns NO, vdsm will fail to start.

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +429,26 @@ def read_metadata():
...
+    """
+    # Assumptions in the regex:
+    # - If a section has nested sections, will not match key-value pairs in
+    #   the top level, only their sub-sections have key-value pairs checked.
+    # - Values can be either strings, numerical values, or quoted strings.
+    #   Therefore, match enabling strings both with and without double quotes.
+    # - Between the section string and the left bracket, a single space is
+    #   allowed as maximum.
+    # - Between keys and values at least one space is required.
+    pattern = r'\s+(\S+) +{[^{]+user_friendly_names +"?yes"?\s+}'
```
nirs, line 453: Few issues:
- The section name always starts with leading whitespace - how does it work for

      defaults {

      }

- This assumes that "user_friendly_names yes" is the last line before the "}"

This would be more clear if we use named regex and verbose option:

aesteve-rh: I think I tried key-values after the "user_friendly_names yes" and worked, but you are right that it should only accept "blank" characters, so maybe I didn't test it right.

I am currently trying to clear a few of the assumptions to rework the regex, but I'll take these comments into consideration next time I push. Possibly also add some simple tests to increase the coverage.

I'll draft the PR for now, until I figure out the small opens.

lib/vdsm/storage/multipath.py
```diff
@@ -209,6 +209,18 @@ def is_ready():
     return "busy: false" in status
 
 
+def get_local_config():
```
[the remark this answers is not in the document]
aesteve-rh: `show_config_local` sound good :+1:

nirs:
> Will print a WARN message if there is any device or section in the multipath daemon configuration that enables 'user_friendly_names', as in:
> 
> ```
>     WARNING: invalid configuration in /etc/multipath.conf:
>     device_name -> user_friendly_names:  yes
>     This configuration is not supported and may lead storage
>     domain corruption.
> ```

Issues:

- The configuration is likely not in /etc/multpath.conf buf from a drop-in file
  in /etc/multipath/conf.d/*.conf, and multipathd does not report the source of
  the configuration value.
- device_name cannot work because the configuration can come from defaults,
  overdides, devices, or multpaths. Devices and multipaths do not have a name
  but we can show the properties of the device or multipath entry.

We can use this:

    user_friendly_names is enabled for 'defaults' key=value ...
    user_friendly_names is enabled for 'overrides' key=value ...
    user_friendly_names is enabled for 'device' key=value ...
    user_friendly_names is enabled for 'multiapth' key=value ...
   
So the check function can collect the name and the attributes of any
unsupported block of confiugration.

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +432,95 @@ def read_metadata():
...
+            assert isinstance(attr, ConfSection)
+            # Nested attributes (recursive call)
+            _, check_ok = _check_conf_attrs(
+                f"{name}.{attr.name}", attr.attrs)
+            if not check_ok:
+                return attr, check_ok
+    return None, True
+
+
+def _parse_conf(mpath_out, parsed_conf, depth=0):
```
nirs, line 494: We can simplify this by accepting an iterator instead of list of lines.

For example the caller can use:

     import io
     it = io.StringIO(out)
     tree = _parse_conf(it)

The advantage of the iterator is that we don't need to manage the depth or pass the list to append value into. Instead we can just return the parsed value.

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +432,95 @@ def read_metadata():
...
+        parsed_conf (List): List to append the resulting parsed configuration.
+        depth (int): Recursion depth.
+
+    Returns:
+        Number of lines parsed in this recursion.
+    """
+    i = 0
+    while i < len(mpath_out):
+        line = mpath_out[i].strip()
+        i += 1
```
nirs, line 510: This is good way to do this in C, but we can make this much simpler using
iteration. Assuming the caller sent us an iterator over the lines of the config:

    for line in mpath_out:
        process line...

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +432,95 @@ def read_metadata():
...
+        if line.endswith("{"):
+            name, _ = line.rsplit(" ", 1)
+            nested_section = []
+            i += _parse_conf(mpath_out[i:], nested_section, depth + 1)
+            parsed_conf.append(ConfSection(name, nested_section))
```
nirs, line 515: If we return the parsed config, this becomes much simpler:

    res.append(name, _parse_conf(mpath_out))

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +432,95 @@ def read_metadata():
...
+        elif line == "}":
+            if depth > 0:
+                return i
```
nirs, line 518: It is better to break here and return at the bottom. What if the last } is missing?
The current code will return None.

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +432,95 @@ def read_metadata():
...
+        else:
+            if not line:
+                # Discard empty lines
+                continue
```
nirs, line 522: We don't have empty lines in multipathd output, but it will be more robust to ignore such lines.

lib/vdsm/storage/mpathconf.py
```diff
@@ -297,6 +299,8 @@
 log = logging.getLogger("storage.mpathconf")
 
 Metadata = namedtuple("Metadata", "revision,private")
+ConfSection = namedtuple("ConfSection", "name,attrs")
+ConfPair = namedtuple("ConfPair", "key,value")
```
nirs, line 303: Since this is the mpathconf module, why not use Section and Pair?

tests/storage/mpathconf_test.py
```diff
@@ -49,6 +52,32 @@
...
+FAKE_MPATHD_OUTPUT = Template("""\
+defaults {
+        verbosity 2
+        polling_interval 5
+        multipath_dir "/lib64/multipath"
+        path_selector "service-time 0"
+        prio_args ""
+        user_friendly_names $defaults_value
+        find_multipaths_timeout -10
+        enable_foreign "NONE"
```
nirs, line 64: I already commented about this - you test here only value in the middle, but we need to test:

- value is first
- value is last
- value is not first or last

This is true for all sections. So using one test config is not helpful.

For testing parsing, we don't care about actual keys - it is enough to use
only few keys, possibly always:

    section {
        key1 "value1"
        key2 "value2"
        key3 "value3"
    }

It would be nice to test sections with no key, 1 key, and multiple keys.

The parser does not care abut the actual names of the sections, it depends
only on the generic format (key, value). So we can test the parser with:

    section1 {
        key1 "value1"
        key2 "value2"
        key3 "value3"
    }
    section2 {
        key1 "value1"
    }
    section3 {
    }
    section4 {
        entry {
            key1 "value1"
            key2 "value2"
            key3 "value3"
        }
        entry {
            key1 "value1"
        }
        entry {
        }
    }

This will help to understand the parser and make the parser test very clear.
For example we can test:

    assert _parse_conf(it) == [
         Section(
             "section1", [
                 Pair("key1", "value1"),
                 ...
             ]
         ),
         ...
    ]

tests/storage/mpathconf_test.py
```diff
@@ -49,6 +52,32 @@
...
+        no_path_retry 16
+}
+multipaths {
+    multipath {
+        wwid "wwid1"
+        user_friendly_names $mpath_value
+    }
+    multipath {
+        wwid "wwid2"
+    }
```
nirs, line 77: You need to test also "devices" section.

tests/storage/mpathconf_test.py
```diff
@@ -172,3 +201,46 @@ def test_mpath_empty_conf(fake_mpath_conf):
...
+    return _factory
+
+
+@pytest.mark.parametrize("test_input,section", [
+    (["yes", "no", "no"], "defaults"),
+    (['"yes"', '"no"', '"no"'], "defaults"),
+    (["no", "yes", "no"], "overrides"),
+    (['"no"', '"yes"', '"no"'], "overrides"),
+    (["no", "no", "yes"], "multipath"),
+    (['"no"', '"no"', '"yes"'], "multipath"),
```
nirs, line 224: This is very hard to understand. what is the first "yes", "no", "yes"?

Don't try to create the shortest possible code. Try to create the easier to 
understand code. Test must the simple as possible to ensure they have no bugs.

tests/storage/mpathconf_test.py
```diff
@@ -172,3 +201,46 @@ def test_mpath_empty_conf(fake_mpath_conf):
...
+    defaults_value, overrides_value, mpath_value = test_input
+    fake_mpathd_config(FAKE_MPATHD_OUTPUT.substitute(
+        {"defaults_value": defaults_value,
+         "overrides_value": overrides_value,
+         "mpath_value": mpath_value}))
+    with pytest.raises(exception.InvalidConfiguration) as e:
+        mpathconf.check_local_config()
+    assert section == e.value.context["section"].name
+    invalid_pair = str(mpathconf.ConfPair('user_friendly_names', 'yes'))
+    assert invalid_pair in str(e.value)
```
nirs, line 236: I think we need simple test of the parser before we test the code validating the parsed tree.

tests/storage/mpathconf_test.py
```diff
@@ -172,3 +201,46 @@ def test_mpath_empty_conf(fake_mpath_conf):
...
+
+
+@pytest.mark.parametrize("value", ["no", '"no"'])
```
nirs, line 239: Using parameters for testing unquoting is nice and simple, but this can be tested in in the parser test. Once we tested that quoted values are unquoted, we can test only quoted values when checking detection of bad options.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +106,27 @@ def isconfigured():
...
+            f"Invalid configuration detected: {e.context['reason']}:\n")
+        sys.stdout.write(f"  {ufn_enabled_section.name} {{\n")
+        for attr in ufn_enabled_section.attrs:
+            line_indent = "    "
+            if attr.key == "user_friendly_names":
+                line_indent = "->" + line_indent
+            sys.stdout.write(f"{line_indent}{attr.key} {attr.value}\n")
+        sys.stdout.write("  }\n")
+        sys.stdout.write("This configuration is not supported "
+                         "and may lead storage domain corruption.\n")
```
nirs, line 130: Please share the output printed here in the commit message.

nirs (changes requested):

lib/vdsm/storage/mpathconf.py
aesteve-rh: Nice, I am not very familiar with the io functions, so I didn't know this could be done so easily.

This will make parsing a lot easier, thanks!

lib/vdsm/storage/mpathconf.py
aesteve-rh: This does not work if you want to jump iterations. That's why I used index-based iteration.
But with the iterator suggested above I can change this easily.

And for C, at least in C++ we have range-based loops available since C++11 :)

lib/vdsm/storage/mpathconf.py
aesteve-rh: Again, byproduct of not using iterator and wanting to jump iterations. Will get fixed with StringIO.

tests/storage/mpathconf_test.py
aesteve-rh: I didn't ignore the point. The cases of:
- value is first
- value is last
- value is not first or last
I addressed them by putting the user_friendly_names pair in different locations in different sections.

But nevermind that. I'll separate the config example in different blocks and tests, with generic names.

tests/storage/mpathconf_test.py
aesteve-rh: I'll use generic names instead.

tests/storage/mpathconf_test.py
aesteve-rh: Fair point. Although I don't think it is that difficult to understand, basically the first line of the function is telling you what each value means. And I've seen more complex uses of the parameters in tests. But ok, I'll separate the parser tests into simpler examples, and the checker itself into a separate set of tests and see how it looks.

aesteve-rh: So finally I kept a test with parameters, but simplified the way the test is parametrized so that it is more readable.

And complemented the test with specific tests for the position of the user_friendly_names attribute in a section.

IMO this is nicer than what I did in the previous version of my parametrized code, and nicer than copy&pasting code multiple times for different section. Let me know what you think.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +106,27 @@ def isconfigured():
...
+        mpathconf.check_local_config()
+    except exception.InvalidConfiguration as e:
+        ufn_enabled_section = e.context["section"]
+        sys.stdout.write(
+            f"Invalid configuration detected: {e.context['reason']}:\n")
+        sys.stdout.write(f"  {ufn_enabled_section.name} {{\n")
+        for attr in ufn_enabled_section.attrs:
+            line_indent = "    "
+            if attr.key == "user_friendly_names":
+                line_indent = "->" + line_indent
```
nirs, line 126: I don't think we need this since we start the warning with "user_friendly_names is enabled". Showing the config as is simpler and better.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +106,27 @@ def isconfigured():
...
+            sys.stdout.write(f"{line_indent}{attr.key} {attr.value}\n")
+        sys.stdout.write("  }\n")
+        sys.stdout.write("This configuration is not supported "
+                         "and may lead storage domain corruption.\n")
+        return NO
```
nirs, line 131: Returning NO means that vdsm will not start on a system with unsupported configuration. This may break existing setups so we may want to just warn about the issue without failing vdsm-tool.

We can start by warning, and later convert the warning to hard error if needed.

nirs (changes requested):

lib/vdsm/storage/mpathconf.py
```diff
@@ -428,3 +430,33 @@ def read_metadata():
         return Metadata(REVISION_OLD, private)
 
     return Metadata(REVISION_MISSING, private)
+
+
+def _parse_conf(mpath_out):
```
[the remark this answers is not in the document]
aesteve-rh: I'll append `_it` to the name.

lib/vdsm/storage/mpathconf.py
```diff
@@ -432,6 +434,62 @@ def read_metadata():
...
+            if attr.key == 'user_friendly_names' and attr.value == 'yes':
+                return None, False
+        else:
+            assert isinstance(attr, Section)
+            # Subsections (recursive call)
+            _, check_ok = _check_conf_attrs(
+                f"{name}.{attr.name}", attr.attrs)
+            if not check_ok:
+                return attr, check_ok
+    return None, True
```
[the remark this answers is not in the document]
aesteve-rh: As a possibility, if all the children must have the same type, then we can use `all` function to avoid checking only the first child. It is less performant but here is not important, and with the data size that the multipath configuration outputs should not matter anyway.

tests/storage/mpathconf_test.py
```diff
@@ -51,6 +53,46 @@
 }
 """
 
+FAKE_MPATHD_OUTPUT = Template("""\
```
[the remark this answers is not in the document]
aesteve-rh: There is no verification intention in the pairs here. The intent is just having a realistic output, as a hint to other maintainers. To me, it is nice having a real feel of the data that we will end up receiving.

But if you see not value in that I can standardize them with "key value" pairs.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: The whole point of the bug fix is making it fail to avoid wrong configurations, no? The issue may get unnoticed if the upgrade tool logs are not checked.

lib/vdsm/storage/mpathconf.py
```diff
@@ -432,6 +434,62 @@ def read_metadata():
...
+def check_local_config():
+    """
+    Check if 'user_friendly_names' is enabled in the multipathd configuration.
+    The configuration is obtained by running the multipathd command to show
+    the local configuration and capturing the output.
+    Parse the output and check that the contents are valid.
+
+    Raises:
+        exception.InvalidConfiguration: if any section have
+            'user_friendly_names' enabled. Contains the offending section name.
```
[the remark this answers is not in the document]
aesteve-rh: Updated also to contain multiple sections.

tests/storage/mpathconf_test.py
```diff
@@ -217,3 +259,79 @@ def test_mpathd_parser():
...
+@pytest.fixture
+def fake_mpathd_config(monkeypatch):
+    def _factory(mpath_output):
+        def mockreturn(*args):
+            return mpath_output
+
+        monkeypatch.setattr(
+            mpathconf.multipath, 'show_config_local', mockreturn)
+
+    return _factory
```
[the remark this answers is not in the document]
aesteve-rh: I have finally removed the `fixture`.

aesteve-rh:
I have addressed all comments. I left a couple threads open for you to read (@nirs), but still I have followed them too.

Hopefully this will be closer to a mergeable version.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: Ok. Let's go for it then, and start with a warning. My main concern is in that the warning can be easily missed, but I understand that breaking after an upgrade is not something desirable for users either.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
             Section('entry', [])
         ]),
     ]
+
+
+def test_mpathd_ufn_enabled_single():
+    fake_conf = [
+        Section('default', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'yes'),
```
nirs, line 233: I think this covers the last pair case...

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
...
+
+def test_mpathd_ufn_enabled_multiple():
+    fake_conf = [
+        Section('default', [
+            Pair('key', 'value'),
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('devices', [
+            Section('device', [
+                Pair('user_friendly_names', 'yes'),
```
nirs, line 262: and this cover the first pair case. If we add another device or multipath to the test data using middle pair, we cover the case.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
...
+            Pair('user_friendly_names', 'yes'),
+        ]),
+        Section('device', [
+            Pair('user_friendly_names', 'yes'),
+            Pair('key', 'value'),
+        ])
+    ]
+
+
+def test_mpathd_ufn_first_pair():
```
nirs, line 283: Already covered int the previous tests.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
...
+    assert issues == [
+        Section('section1', [
+            Pair('user_friendly_names', 'yes'),
+            Pair('key2', 'value2'),
+            Pair('key3', 'value3'),
+        ])
+    ]
+
+
+def test_mpathd_ufn_mid_pair():
```
nirs, line 301: Not covered in the previous tests but can be easily added to the multiple test.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
...
+    assert issues == [
+        Section('section1', [
+            Pair('key1', 'value1'),
+            Pair('user_friendly_names', 'yes'),
+            Pair('key3', 'value3'),
+        ])
+    ]
+
+
+def test_mpathd_ufn_last_pair():
```
nirs, line 319: Already covered by previous tests.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
...
+            Pair('user_friendly_names', 'yes'),
+        ])
+    ]
+
+
+def test_mpathd_ufn_only_pair():
+    fake_conf = [
+        Section('section1', [
+            Pair('user_friendly_names', 'yes'),
+        ])
```
nirs, line 341: This can also be added to the previous tests, maybe the multiple test?

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
             Section('entry', [])
         ]),
     ]
+
+
+def test_mpathd_ufn_enabled_single():
+    fake_conf = [
+        Section('default', [
```
nirs, line 231: Just for consistency, this is called "defaults". If you send another version it would be nice to rename in all tests.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,157 @@ def test_mpathd_parser():
...
+        ])
+    ]
+
+
+def test_mpathd_ufn_no_pairs():
+    fake_conf = [
+        Section('section1', [])
+    ]
+    issues = mpathconf._check_conf(fake_conf)
+    assert not issues
```
nirs, line 356: Testing empty list will be nice for completeness.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +106,25 @@ def isconfigured():
...
+        ufn_enabled_sections = e.context["sections"]
+        for section in ufn_enabled_sections:
+            sys.stdout.write(
+                f"Invalid configuration detected: {e.context['reason']}:\n")
+            sys.stdout.write(f"  {section.name} {{\n")
+            for attr in section.children:
+                sys.stdout.write(f"    {attr.key} {attr.value}\n")
+            sys.stdout.write("  }\n")
+            sys.stdout.write("This configuration is not supported "
+                             "and may lead to storage domain corruption.\n")
```
nirs, line 128: It would be better to use a specific error and format the error message in the exception class __str__(). The we will simplify do there:

    try:
        check config...
    except InvalidMultipathConfiguration as e:
        sys.stdout.write("WARNING: %s", e)

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +106,25 @@ def isconfigured():
...
+
+def isconfigured():
+    """
+    Check the multipath daemon configuration.
+    """
+    conf_metadata_ok = _check_mpath_metadata()
+    try:
+        mpathconf.check_local_config()
+    except exception.InvalidConfiguration as e:
+        ufn_enabled_sections = e.context["sections"]
```
nirs, line 119: Using the context works, but it will be better to handle error formatting in the exception class.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: I think is good as it is. The exception is a generic container with a bit of context.

`sections` contain the raw contents of the configuration structure, so it will get printed as:
```
Invalid configuration value: 'user_friendly_names' is enabled in multipath configuration {'sections': [Section(name='defaults', children=[Pair(key='key', value='value'), Pair(key='user_friendly_names', value='yes')]), Section(name='device', children=[Pair(key='user_friendly_names', value='yes'), Pair(key='key', value='value')])]}
```

Which is horrible for users. This structure needs to get transformed into a readable format, but the exception class is agnostic to whatever the context dictionary may contain. Then, we either transform it before in the mpathconf module or after, in the configurator. IMO, the configurator is a good place to do it, is where we have full control over what we want to show.

In here I just need to get the first and last lines out of the loop, and it looks good. I indented it wrongly in some update.

lib/vdsm/storage/mpathconf.py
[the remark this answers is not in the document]
aesteve-rh: Then I can use `all` here, that will fix checking only the first child and ensure consistency in the sections.

lib/vdsm/tool/configurators/multipath.py
nirs: The exception __str__ can return a string with multiple lines - the same thing you output here. But it is true that it will be harder to control how the output look in the tool.

If you want to format the message here, use more specific exception that has a public
"issues" or "sections" attribute instead of abusing the context dict.

aesteve-rh: Something like this?

```python
class InvalidMPathConfiguration(InvalidConfiguration):
    code = 103
    message = "Invalid configuration value"

    def __init__(self, reason, sections):
        super().__init__(reason)
        self.sections = sections
    
    def __str__(self):
        error_str = f"{self.msg}: {self.context['reason']}:\n"
        for section in self.sections:
            error_str += f"  {section.name} {{\n"
            for attr in section.children:
                error_str += f"    {attr.key} {attr.value}\n"
            error_str += "  }\n"
        error_str += "This configuration is not supported " \
                     "and may lead to storage domain corruption.\n"
        return error_str
```

nirs: I would use io.StringIO to format, but this is the idea. If you do the formatting
in `__str__` you don't need to provide a public `sections` attribute. If you want to format the error in the tool, using a public attribute is better.

aesteve-rh: StringIO sounds good. I will probably format the error in the _ _str_ _. As we are introducing a new exception class anyway, my argument above about keeping it agnostic to context is not important.

nirs: At the end what matters is what works best for our needs. The general approach in vdsm to to format errors in a useful way in the exception class. However we sue this only for logging when we care less about the formatting and more about the content. If formatting the error in the tool is easier do it in the tool.

Another option is to avoid the exception and instead return the result of the check to the tool, and format the warning in the tool. Even if we want to fail, we can do this
in the tool instead of the mpathconf module.

lib/vdsm/common/exception.py
```diff
@@ -505,6 +505,18 @@ class MigrationOperationError(VdsmException):
...
+class InvalidMPathConfiguration(InvalidConfiguration):
+    code = 103
+    message = "Invalid configuration value"
+
+    def __init__(self, reason, sections):
+        super().__init__(reason)
+        self.sections = sections
+
+    def __str__(self):
+        return f"{self.msg}: {self.context['reason']}"
```
nirs, line 517: Why do we need to override `__str__`? The only reason I can see is include the `sections` in the log, but you don't include it.

The easiest way to handle this is to inherit the super class `__str__` and make sure the sections are included in the context, which is already part of the supervlass `__str__`.

A possible solution can be:

    class InvalidMPathConfiguration(ContextException):
        ...
        @property
        def sections(self):
            return self.context["sections"]

Now we have good `__str__` without writing any code, and we have a public property to access the sections that can be used by the tool to format a warning.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +106,24 @@ def isconfigured():
...
+def isconfigured():
+    """
+    Check the multipath daemon configuration.
+    """
+    conf_metadata_ok = _check_mpath_metadata()
+    try:
+        mpathconf.check_local_config()
+    except exception.InvalidMPathConfiguration as e:
+        # Just warn for now, do not fail the isconfigured check here.
+        sys.stdout.write(f"WARNING: {e}:\n")
```
nirs, line 120: Using the error here means the error `__str__` cannot include all the info about the error, which is bad if we want to log it.

In the context of the tool, you should use your own message instead of using the error string representation.

Looks like you make the work harder by raising an error from mpathconf module, instead of reporting the status and using it here.

if we use:

    result = mpathconf.check_local_config()

And result is a named tuple with:
- error
- issues

We should format here the error and the issues, and we don't need to introduce a new exception. This is how lvmfilter and lvmdevices module work.

lib/vdsm/common/exception.py
aesteve-rh: The `__str__` method is overriden because the parent `__str__` returns a raw dictionary as in `{"reason": "this is the reason", "another_key": "another_value"}`, which is fine for the text of an exception but a bit ugly for the output of a command line tool.

Since I couldn't use the context dictionary from the tool where the exception is catched, I did override the string method.
I can expose the `reason` (and `context`) keys through properties as you suggest, but it is mostly a cosmetic discussion.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: That is also how this was coded before, until I changed it following a comment. I don't mind removing the exception and bringing that back again.

nirs (approved):

nirs:
OST failed when trying to start vdsmd service:

```
2022-07-25T13:31:46.609500+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40889]: vdsm: Running check_is_configured
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]: Traceback (most recent call last):
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/bin/vdsm-tool", line 209, in main
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     return tool_command[cmd]["command"](*args)
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/tool/__init__.py", line 40, in wrapper
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     func(*args, **kwargs)
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/tool/configurator.py", line 165, in isconfigured
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     m = [c.name for c in pargs.modules if _isconfigured(c) == configurators.NO]
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/tool/configurator.py", line 165, in <listcomp>
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     m = [c.name for c in pargs.modules if _isconfigured(c) == configurators.NO]
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/tool/configurator.py", line 103, in _isconfigured
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     return getattr(module, 'isconfigured', lambda: configurators.NO)()
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/tool/configurators/multipath.py", line 115, in isconfigured
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     check_result = mpathconf.check_local_config()
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/storage/mpathconf.py", line 447, in check_local_config
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     mpath_conf = _parse_conf(io.StringIO(multipath.show_config_local()))
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/storage/multipath.py", line 221, in show_config_local
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     [_MULTIPATHD.cmd, "show", "config", "local"]).decode('ascii')
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:   File "/usr/lib/python3.6/site-packages/vdsm/common/commands.py", line 102, in run
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]:     raise cmdutils.Error(args, p.returncode, out, err)
2022-07-25T13:31:48.055712+0000 ost-basic-suite-master-host-0 vdsmd_init_common.sh[40924]: vdsm.common.cmdutils.Error: Command ['/usr/sbin/multipathd', 'show', 'config', 'local'] failed with rc=1 out=b'' err=b'Jul 25 13:31:48 | /etc/multipath.conf does not exist, blacklisting all devices.\nJul 25 13:31:48 | You can run "/sbin/mpathconf --enable" to create\nJul 25 13:31:48 | /etc/multipath.conf. See man mpathconf(8) for more details\n'
```

Looks like an issue with OST - trying to start vdsm *before* vdsm is configured (vdsm-tool configure --force), but it needs more investigation.

@aesteve-rh did you test fresh install on a new vm?

We may need to skip the multipath check if vdsm was not configured yet (e.g. /etc/multipath.conf not installed yet).

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +105,24 @@ def isconfigured():
 
     sys.stdout.write("multipath requires configuration\n")
     return NO
+
+
+def isconfigured():
+    """
+    Check the multipath daemon configuration.
+    """
+    conf_metadata_ok = _check_mpath_metadata()
```
nirs, line 114: Checking multipath is possible only if conf_metadata_ok is "YES".

nirs (changes requested):

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +105,26 @@ def isconfigured():
...
+    if conf_metadata_ok == YES:
+        check_result = mpathconf.check_local_config()
+        if check_result.error is not None:
+            # Just warn for now, do not fail the isconfigured check here.
+            sys.stdout.write(
+                f"WARNING: Invalid configuration: {check_result.error}:\n")
```
nirs, line 120: If the check failed because running "multipathd show config local" failed, we log now:

    WARNING: Invalid configuration: error from multpathd...

Instead of:

    WARNING: Cannot check multipath configuration : error from multipathd...

nirs (changes requested):

aesteve-rh:
Now it fails with:

```
 2022-07-25T14:51:18.793734+0000 ost-basic-suite-master-host-1 vdsmd_init_common.sh[23057]:     raise ValueError(f"Invalid item: {item}")                                                                                                                                                                                     
 2022-07-25T14:51:18.794331+0000 ost-basic-suite-master-host-1 vdsmd_init_common.sh[23057]: ValueError: Invalid item: Section(name='blacklist', children=[Pair(key='devnode', value='^(rbd)[0-9]*'), Pair(key='devnode', value='!^(sd[a-z]|dasd[a-z]|nvme[0-9])'), Pair(key='protocol', value='(scsi:adt|scsi:sbp)'),          Section(name='device', children=[Pair(key='vendor', value='SGI'), Pair(key='product', value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='^DGC'), Pair(key='product', value='LUNZ')]), Section(name='device', children=[Pair(key='vendor', value='EMC'), Pair(key='product',               value='LUNZ')]), Section(name='device', children=[Pair(key='vendor', value='DELL'), Pair(key='product', value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='FUJITSU'), Pair(key='product', value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='IBM'),   Pair(key='product', value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='IBM'), Pair(key='product', value='S/390')]), Section(name='device', children=[Pair(key='vendor', value='LENOVO'), Pair(key='product', value='Universal Xport')]), Section(name='device',                           children=[Pair(key='vendor', value='(NETAPP|LSI|ENGENIO)'), Pair(key='product', value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='STK'), Pair(key='product', value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='SUN'), Pair(key='product',           value='Universal Xport')]), Section(name='device', children=[Pair(key='vendor', value='(Intel|INTEL)'), Pair(key='product', value='VTrak V-LUN')]), Section(name='device', children=[Pair(key='vendor', value='Promise'), Pair(key='product', value='VTrak V-LUN')]), Section(name='device', children=[Pair(key='vendor',     value='Promise'), Pair(key='product', value='Vess V-LUN')])]) 
```

Looks like the `multipath show conf` returned a weird output. I will check where this comes from. But it is mixing Pairs with Sections in the same level, and that is why it fails.

nirs:
Correct, the blacklist section support both pairs and sections. This is normal
and not related to OST.

We need to update tests to reflect real structure and change the parser to support
this.

We can skip the blackist section when checking since it cannot enable user friendly
names, but for parsing we need consume it.

lib/vdsm/storage/mpathconf.py
```diff
@@ -432,6 +435,58 @@ def read_metadata():
...
+
+    Args:
+        mpath_conf: Parsed structure of the multipathd configuration output.
+
+    Returns:
+        List[Section]: Sections that have invalid configuration.
+    """
+    issues = []
+    for item in mpath_conf:
+        if not item.children or item.name == "blacklist":
```
nirs, line 476: Three is also blacklist_exceptions which can contain the same content as blacklist.

tests/storage/mpathconf_test.py
```diff
@@ -224,3 +224,133 @@ def test_mpathd_parser():
...
+        ]),
+        Section('blacklist', [
+            Pair('key', 'value'),
+            Section('name1', [
+                Pair('key1', 'value1'),
+                Pair('key2', 'value2'),
+            ]),
+            Section('name2', [
+                Pair('key', 'value')
+            ])
```
nirs, line 313: There is also blacklist_exceptions. We need to test both of them with the parser (currently it does not have mixed sections in the config), and in the checker.

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -109,3 +105,26 @@ def isconfigured():
...
+    """
+    Check the multipath daemon configuration.
+    """
+    conf_metadata_ok = _check_mpath_metadata()
+    if conf_metadata_ok == YES:
+        check_result = mpathconf.check_local_config()
+        if check_result.error is not None:
+            # Just warn for now, do not fail the isconfigured check here.
+            sys.stdout.write(
+                f"WARNING: {check_result.error}:\n")
```
nirs, line 120: It does not fit in one line now?

tests/storage/mpathconf_test.py
```diff
@@ -172,3 +175,52 @@ def test_mpath_empty_conf(fake_mpath_conf):
...
+    entry {
+        key1 "value 1"
+        key2 "value 2"
+        key3 "value 3"
+    }
+    entry {
+        key1 "value1"
+    }
+    entry {
+    }
```
nirs, line 202: We need to test here mixed section with both pairs and other sections (e.g. blacklist).

nirs (changes requested):

lib/vdsm/storage/mpathconf.py
aesteve-rh: True, I just checked the documentation. I have added both. The other sections are "defaults", "multipaths", and "devices", so we should be fine excluding those two.

aesteve-rh:
The new error looks unrelated:

```
2022-07-26T10:18:43.150594+0000 ost-basic-suite-master-host-0 vdsm[41758]: ERROR Exception raised
File "/usr/lib/python3.6/site-packages/vdsm/vdsmd.py", line 159, in run serve_clients(log)
File "/usr/lib/python3.6/site-packages/vdsm/vdsmd.py", line 120, in serve_clients  atexit._run_exitfuncs()                                                                                                                                                                                                                       
File "/usr/lib/python3.6/site-packages/vdsm/clientIF.py", line 392, in prepareForShutdown secret.clear()                                                                                                                                                                                                                                
File "/usr/lib/python3.6/site-packages/vdsm/virt/secret.py", line 96, in clear  for virsecret in con.listAllSecrets():                                                                                                                                                                                                        
File "/usr/lib/python3.6/site-packages/vdsm/common/libvirtconnection.py", line 131, in wrapper ret = f(*args, **kwargs)
File "/usr/lib/python3.6/site-packages/vdsm/common/function.py", line 94, in wrapper return func(inst, *args, **kwargs)                                                                                                                                                                                                                                                                                          File "/usr/lib64/python3.6/site-packages/libvirt.py", line 6164, in listAllSecrets raise libvirtError("virConnectListAllSecrets() failed")  
libvirt.libvirtError: invalid connection pointer in virConnectListAllSecrets 
```

nirs (approved):

PR #239, opened 2022-06-07
mpathconf: move logic from tools.multipath to mpathconf

aesteve-rh:
Move all the logic to handle, write, and check multipathd configuration from the tools.multipath module to mpathconf.

The multipath module will act as a wrapper of the mpathconf module for the vdsm-tool, and it may add some logging and
extra functionality on top of it, but the logic shall be kept in the storage.mpathconf module.

Add some tests to verify the functions that were moved.

Update/homogenize some minor parts of the code.

### DIFF
--- a/lib/vdsm/storage/mpathconf.py
+++ b/lib/vdsm/storage/mpathconf.py
@@ -19,17 +19,21 @@
 #
 
 """
-This module provides multipath configuration functionality over the host.
-Currently it only provides blacklist configuration for setting WWIDs of
-disk devices which should be excluded from multipath management.
+This module provides multipath configuration functionality over the host:
+- Blacklist configuration for setting WWIDs of disk devices
+  which should be excluded from multipath management.
+- Configuration metadata parser.
 """
 import io
 import logging
 import os
 import re
 
+from collections import namedtuple
+
 from . import fileUtils
 
+CONF_FILE = '/etc/multipath.conf'
 _VDSM_MULTIPATH_BLACKLIST = '/etc/multipath/conf.d/vdsm_blacklist.conf'
 
 _HEADER = """\
@@ -39,8 +43,49 @@ _HEADER = """\
 
 """
 
+# The first line of multipath.conf configured by vdsm must contain a
+# "VDSM REVISION X.Y" tag.  Note that older version used "RHEV REVISION X.Y"
+# format.
+
+CURRENT_TAG = "# VDSM REVISION 2.2"
+
+_OLD_TAGS = (
+    "# VDSM REVISION 2.1",
+    "# VDSM REVISION 2.0",
+    "# VDSM REVISION 1.9",
+    "# VDSM REVISION 1.8",
+    "# VDSM REVISION 1.7",
+    "# VDSM REVISION 1.6",
+    "# VDSM REVISION 1.5",
+    "# VDSM REVISION 1.4",
+    "# VDSM REVISION 1.3",
+    "# VDSM REVISION 1.2",
+    "# RHEV REVISION 1.1",
+    "# RHEV REVISION 1.0",
+    "# RHEV REVISION 0.9",
+    "# RHEV REVISION 0.8",
+    "# RHEV REVISION 0.7",
+    "# RHEV REVISION 0.6",
+    "# RHEV REVISION 0.5",
+    "# RHEV REVISION 0.4",
+    "# RHEV REVISION 0.3",
+    "# RHAT REVISION 0.2",
+)
+
+# The second line of multipath.conf may contain PRIVATE_TAG. This means
+# vdsm-tool should never change the conf file even when using the --force flag.
+
+_PRIVATE_TAG = "# VDSM PRIVATE"
+_OLD_PRIVATE_TAG = "# RHEV PRIVATE"
+_PRIVATE_TAGS = (_PRIVATE_TAG, _OLD_PRIVATE_TAG)
+
+REVISION_OK = "OK"
+REVISION_OLD = "OLD"
+REVISION_MISSING = "MISSING"
 
 log = logging.getLogger("storage.mpathconf")
+
+Metadata = namedtuple("Metadata", "revision,private")
 
 
 def configure_blacklist(wwids):
@@ -130,3 +175,33 @@ def read_blacklist():
         wwids.add(wwid)
 
     return wwids
+
+
+def read_metadata():
+    """
+    The multipath configuration file at /etc/multipath.conf should contain
+    a tag in form "RHEV REVISION X.Y" for this check to succeed.
+    If the tag above is followed by tag "RHEV PRIVATE" (old format) or
+    "VDSM PRIVATE" (new format), the configuration should be preserved
+    at all cost.
+
+    Returns:
+        vdsm.storage.mpathconf.Metadata
+    """
+    first = second = ''
+    with open(CONF_FILE) as f:
+        mpathconf = [x.strip("\n") for x in f.readlines()]
+    try:
+        first = mpathconf[0]
+        second = mpathconf[1]
+    except IndexError:
+        pass
+    private = second.startswith(_PRIVATE_TAGS)
+
+    if first.startswith(CURRENT_TAG):
+        return Metadata(REVISION_OK, private)
+
+    if first.startswith(_OLD_TAGS):
+        return Metadata(REVISION_OLD, private)
+
+    return Metadata(REVISION_MISSING, private)
--- a/lib/vdsm/storage/mpathconf.py
+++ b/lib/vdsm/storage/mpathconf.py
@@ -22,6 +22,7 @@
 This module provides multipath configuration functionality over the host:
 - Blacklist configuration for setting WWIDs of disk devices
   which should be excluded from multipath management.
+- Vdsm-supported multipathd configuration.
 - Configuration metadata parser.
 """
 import io
@@ -30,6 +31,7 @@ import os
 import re
 
 from collections import namedtuple
+from string import Template
 
 from . import fileUtils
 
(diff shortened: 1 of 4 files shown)


lib/vdsm/storage/mpathconf.py
```diff
@@ -130,3 +167,44 @@ def read_blacklist():
...
+    with open(CONF_FILE) as f:
+        mpathconf = [x.strip("\n") for x in f.readlines()]
+    try:
+        first = mpathconf[0]
+        second = mpathconf[1]
+    except IndexError:
+        pass
+    if _PRIVATE_TAG in second or _OLD_PRIVATE_TAG in second:
+        log.info("Manual override for multipath.conf detected"
+                 " - preserving current configuration\n")
```
nirs, line 409: log.info does not work in the context of running vdsm tool. The tool logs to stdout now,
and we don't want to change this now. To move the code to storage, we need to change
the interface so this function does not return a boolean but a enum describing the state of the system.

I think in this case the function should return this info:
- revision: OK, OLD, MISSING
- private: bool

The caller will handle warn about the various states in the context of the configurator.

The lvmfilter module is a good example for this kind of code.

tests/storage/mpathconf_test.py
```diff
@@ -49,6 +49,21 @@
...
+"""
+
+MPATH_CONF_PRIVATE = f"""\
+{mpathconf.CURRENT_TAG}
+{mpathconf._PRIVATE_TAG}
+"""
+
+MPATH_CONF_OLD_TAG = f"""\
+{mpathconf._OLD_TAGS[-1]}
+# Invalid configuration
```
nirs, line 64: Are you sure this is invalid? This is just old revision that needs an upgrade.

tests/storage/mpathconf_test.py
```diff
@@ -49,6 +49,21 @@
...
 """
 
+MPATH_CONF = f"""\
+{mpathconf.CURRENT_TAG}
+# Valid configuration
+"""
+
+MPATH_CONF_PRIVATE = f"""\
+{mpathconf.CURRENT_TAG}
+{mpathconf._PRIVATE_TAG}
```
nirs, line 59: Missing comment, or maybe we don't need the comment in other examples. The constant name can be enough to explain the example.

tests/storage/mpathconf_test.py
```diff
@@ -102,3 +117,33 @@ def test_read_bad_blacklist(fake_conf):
...
+@pytest.fixture
+def fake_mpath_conf(tmpdir, monkeypatch):
+    fake_conf = tmpdir.join("multipath.conf")
+    monkeypatch.setattr(mpathconf, "CONF_FILE", str(fake_conf))
+
+    def _write_data(data):
+        fake_conf.write(data)
+        return fake_conf
+
+    return _write_data
```
nirs, line 131: Returning a function works but not consistent with other fixtures.

When we have a fixture that needs configuration, the best way to  implement it is
returning an object supporting attributes and methods that helps to create the fixture
content. This is the way pytest fixture are implemented and we follow this design.

The fixture for multiapth conf need to do 2 things:
- monkeypatch the path to the module will find the fake file
- write content - this must be done in the test context, since each test want to use different content.

The simplest way to implement this is to do:

    @pytest.fixture
    def fake_conf(tmpdir, monkeypatch):
        path = tmpdir.join("multipath.conf")
        monkeypatch.setattr(mpathconf, "CONF_FILE", str(path))
        return path

Now the test can write the needed content:

    def test_mpath_conf(fake_conf):
        fake_conf.write("test data")
        test it...

tests/storage/mpathconf_test.py
```diff
@@ -102,3 +117,33 @@ def test_read_bad_blacklist(fake_conf):
...
+    assert mpathconf.check_conf_revision()
+
+
+@pytest.mark.parametrize("data", [
+    pytest.param(MPATH_CONF_OLD_TAG, id="old_tag"),
+    pytest.param("", id="empty"),
+])
+def test_mpath_conf_invalid(fake_mpath_conf, data):
+    fake_mpath_conf(data)
+    assert not mpathconf.check_conf_revision()
```
nirs, line 155: I don't think using pytest parameters is helpful in this case. It will be more clear to have one test per case with a good name explaining the use case and keep the test content in the test function.

For example:

    def test_missing_conf(fake_conf):
        assert mpathconf.check_conf_revision() == MultipathConf(
            revision=MISSING,
            private=False,
        )

    def test_empty_conf(fake_conf):
        fake_conf.write("")
        assert mpathconf.check_conf_revision() == MultipathConf(
            revision=MISSING,
            private=False,
        )

    def test_empty_conf(fake_conf):
        fake_conf.write("# VDSM REVISION 2.0\n# VDSM PRIVATE\ndata\n")
        assert mpathconf.check_conf_revision() == MultipathConf(
            revision=OLD,
            private=True,
        )

Using the actual values in the test is better than using the constants - these values can never change since they were already released, so we don't get anything by using the constants. Using constants helps when we can change the constant value and get all the relevant code use the new value transparently. This is actually the worst thing we can do here - if someone will change the constant in the module, the test will pass while the real code will fail in runtime trying to parse old values.

Using the actual values is also more clear - we can see the actual data in the test and we don't need to search for the value to understand the test.

lib/vdsm/storage/mpathconf.py
```diff
@@ -169,6 +377,19 @@ def read_blacklist():
...
 
 
+def configure_multipathd():
+    """
+    Set up multipath daemon configuration to the known and supported state.
+    The original configuration, if any, is saved.
+    """
+    backup = fileUtils.backup_file(CONF_FILE)
+    if backup:
+        log.info("Previous multipath.conf copied to %s\n" % backup)
```
nirs, line 387: Again this log will be hidden in vdsm-tool it should go to stdout, but we don't want
to do this here.

We can return the backup path and let the caller do the logging in the configurator.

lib/vdsm/storage/mpathconf.py
```diff
@@ -169,6 +377,19 @@ def read_blacklist():
...
+
+    data = _CONF_DATA.encode('utf-8')
+    fileUtils.atomic_write(CONF_FILE, data, relabel=True)
+
```
nirs, line 391: Otherwise this change is nice and will make the configurator really simple!

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -248,15 +40,9 @@
 
 def configure():
     """
-    Set up the multipath daemon configuration to the known and
-    supported state. The original configuration, if any, is saved
+    Set up the multipath daemon configuration and start the service.
     """
-    backup = fileUtils.backup_file(mpathconf.CONF_FILE)
-    if backup:
-        sys.stdout.write("Previous multipath.conf copied to %s\n" % backup)
```
nirs, line 292: Another option to keep logging as is, is to pass a logger object to the mpathconf.configure_multiapthd(). We can use a standard logger object backed up
by sys.stdout, using simple log format, maybe only adding the logging level.

I think this should be fixed in vdsm-tool - use standard logger for all configurators, and stop using sys.stdout. But we cannot fix the entire world at the same time. We can start by adding a logger in this module, and later use the logger in all the configurators.

lib/vdsm/storage/mpathconf.py
```diff
@@ -400,14 +400,10 @@ def check_conf_revision():
...
     with open(CONF_FILE) as f:
-        mpathconf = [x.strip("\n") for x in f.readlines()]
-    try:
-        first = mpathconf[0]
-        second = mpathconf[1]
-    except IndexError:
-        pass
+        # Read only the lines that are going to be processed
+        first, second = [x.strip("\n")
+                         for x in (f.readline() for _ in range(2))]
```
nirs, line 406: This is much more complicated than needed. If you want to read the first 2 lines, just do:

    first = f.readline().rstrip("\n")
    second = f.readline().rstrip("\n")

Also we want to strip only the right side, so we should use rstrip().

lib/vdsm/tool/configurators/multipath.py
```diff
@@ -22,15 +22,13 @@
 import os
 import sys
 
-from vdsm.common import cmdutils
 from vdsm.common import commands
 from vdsm.storage import mpathconf
+from vdsm.storage.multipath import _MULTIPATHD
```
nirs, line 27: We should not import private names. Make it public if needed. But why do we need the path to multiapthd in the configurator? It should really use a storage module for controlling the daemon.

We can move the code to introspect and control the multipathd daemon to vdsm.storage.multipathd. Or move the code using it to mpathconf?

nirs (changes requested):

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: I like this solution mainly because it entails make the tool use logging which is per se a nice and relatively easy improvement, that solves the issue naturally.

But I don't like the idea of having it mixed up (logging + stdout), it can be confusing. I'll create an issue with the proposal, and we can work on it in a dedicated PR.

aesteve-rh:
This first forced push is for the rebase, to separate it from the actual changes after the review.

tests/storage/mpathconf_test.py
aesteve-rh: Right, is invalid in the sense that it would return a NO in the validity check in the `vdsm-tool`. I'll use more specific wording here.

tests/storage/mpathconf_test.py
aesteve-rh: I'll change and use the actual values instead of the constants. Got it.

Regarding the pytest parameters, I saw that the lvmconf_test was doing something similar and I liked it:
```
@pytest.mark.parametrize("data", [
    pytest.param(CONF, id="example"),
    pytest.param("", id="empty"),
])
def test_modify(tmpdir, data):
```
But with the new approach for what the function returns is true that it is a bit too complicated to keep it like that, so I'll change that too.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: Moved it to storage.multipath, better solution indeed.

lib/vdsm/tool/configurators/multipath.py
aesteve-rh: Issue: https://github.com/oVirt/vdsm/issues/258

lib/vdsm/storage/mpathconf.py
```diff
@@ -79,6 +80,213 @@
...
+
+# Options defined here override device specific options embedded into
+# multipathd.
+
+overrides {{
+    # NOTE: see comments in default section how to compute this value.
+    no_path_retry   {no_path_retry:d}
+}}
+
+""".format(current_tag=CURRENT_TAG, no_path_retry=_NO_PATH_RETRY)
```
nirs, line 288: str.format() is not the write tool for this file format, we have to use confusing {{ and }} everywhere. The previous code sing %(key)s and dict was better.

Template languages usually use {{key}} or {{{key}}} to avoid this issue, since actual text is unlikely to contain this.

lib/vdsm/storage/multipath.py
```diff
@@ -209,6 +209,14 @@ def is_ready():
     return "busy: false" in status
 
 
+def reconfigure():
+    """
+    Invoke multipathd to reconfigure the multipaths.
+    Must run as root.
+    """
+    commands.run([_MULTIPATHD.cmd, "reconfigure"])
```
nirs, line 217: We can implement this like resize_map(), so it works also when called as unprivileged user, but we don't have a use case for this so lets keep it simple as is.

nirs (changes requested):
Nice!

lib/vdsm/storage/mpathconf.py
```diff
@@ -39,9 +44,55 @@
...
+MultipathConf = namedtuple("MultipathConf", "revision,private")
+
+
+class Revision(Enum):
+    # Revision number appears in the configuration and is up to date.
+    ok = auto()
+    # Revision number appears in the configuration but is outdated.
+    old = auto()
+    # Revision number does not appear in the configuration.
+    missing = auto()
```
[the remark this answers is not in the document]
aesteve-rh: These are three variables that are related, so to me, it makes sense to put them in a collection such as an `Enum`. Also, the value of the variables doesn't really matter. With `Enum` we can use the more explicit canonical name instead of arbitrary data, such as strings, and not worry about their value.

I don't see why using magic strings makes the code more clear. Quite the opposite imo, makes people guess if they relate and have to infer it from the code and the variable name.

tests/storage/mpathconf_test.py
```diff
@@ -49,6 +49,24 @@
 }
 """
 
+MPATH_CONF_CURRENT_TAG = """\
+# VDSM REVISION 2.2
+"""
```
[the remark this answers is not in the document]
aesteve-rh: The rest of the constants in this file are used exactly once in the test (except `CONF`). But is fine, I agree that they fit better in their own test now that it is not parametrized. I'll change it.

lib/vdsm/storage/mpathconf.py
aesteve-rh: That's the disadvantage of `format`, yes. Still, it is already used in the same file with the escaped curly braces and all. But here it happens a lot more, so if you don't like it, I'll change it back.

aesteve-rh: I'll try with `string.Template`, seems the best fit for this usecase. Thanks for the suggestion!

lib/vdsm/storage/mpathconf.py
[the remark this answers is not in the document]
michalskrivanek: code consistency is a stronger preference than nuances in usage. There are so many of these constants in areas that no one knows by heart and that makes it difficult to replace in one go. Still, that would be the only way how to introduce such a change (if it would be worth it, which it is likely not). Having two different styles throughout the code is the worst option.

lib/vdsm/storage/mpathconf.py
```diff
@@ -130,3 +385,46 @@ def read_blacklist():
...
+        second = f.readline().rstrip("\n")
+
+    private = _PRIVATE_TAG in second or _OLD_PRIVATE_TAG in second
+
+    if CURRENT_TAG in first:
+        return Metadata(REVISION_OK, private)
+
+    for tag in _OLD_TAGS:
+        if tag in first:
+            return Metadata(REVISION_OLD, private)
```
aesteve-rh, line 428: Just because I like simple one-liners, this could also be written as:
```python
    if any(tag in first for tag in _OLD_TAGS):
        return Metadata(REVISION_OLD, private)
```
Both should have the same cost, and for this simple loop it is not complicated to read.

Honestly, I don't think I will change it, not worth it. But I thought it might be interesting to know your opinion on this style @nirs :)

nirs: Using any is nicer, and the cost does not matter sonce we run this once per install/upgrade.

But can we use this?

    first.startswith(_OLD_TAGS)

aesteve-rh: Oh! I didn't know `startswith` could take an iterable too!

Tried it locally, and it passes the test, so it seems to work correctly :+1: 

I can push it, then, and the patch would be ready to review. No need to run OST again, the code in this patch is not covered atm.

tests/storage/mpathconf_test.py
```diff
@@ -102,3 +102,64 @@ def test_read_bad_blacklist(fake_conf):
...
+def fake_mpath_conf(tmpdir, monkeypatch):
+    fake_conf = tmpdir.join("multipath.conf")
+    monkeypatch.setattr(mpathconf, "CONF_FILE", str(fake_conf))
+    return fake_conf
+
+
+def test_mpath_current_tag_conf(fake_mpath_conf):
+    mpath_conf_current_tag = """\
+# VDSM REVISION 2.2
+"""
```
nirs, line 126: Current target is must use the value from mpathconf.CURRENT_TAG since once we change it, the test will break, or we will have to change  the test each time we change the revision.

tests/storage/mpathconf_test.py
```diff
@@ -102,3 +102,64 @@ def test_read_bad_blacklist(fake_conf):
...
+    fake_mpath_conf.write(mpath_conf_current_tag)
+    assert mpathconf.read_metadata() == mpathconf.Metadata(
+        revision=mpathconf.REVISION_OK,
+        private=False,
+    )
+
+
+def test_mpath_private_current_tag_conf(fake_mpath_conf):
+    mpath_conf_current_private = """\
+# VDSM REVISION 2.2
```
nirs, line 136: Same here, use current "constant".

tests/storage/mpathconf_test.py
```diff
@@ -102,3 +102,64 @@ def test_read_bad_blacklist(fake_conf):
...
+"""
+    fake_mpath_conf.write(mpath_conf_current_tag)
+    assert mpathconf.read_metadata() == mpathconf.Metadata(
+        revision=mpathconf.REVISION_OK,
+        private=False,
+    )
+
+
+def test_mpath_private_current_tag_conf(fake_mpath_conf):
+    mpath_conf_current_private = """\
```
nirs, line 135: When the variable is the context of a single test we don't need the long name, "data" or "text" is better.

The name of the tests explains the data or text used here.

lib/vdsm/storage/mpathconf.py
```diff
@@ -82,6 +84,215 @@
...
+
+overrides {
+    # NOTE: see comments in default section how to compute this value.
+    no_path_retry   $no_path_retry
+}
+
+""").substitute({"current_tag": CURRENT_TAG,
+                 "no_path_retry": _NO_PATH_RETRY})
+
+
```
nirs, line 295: Much nicer!

lib/vdsm/storage/mpathconf.py
```diff
@@ -413,14 +413,11 @@ def read_metadata():
...
     with open(CONF_FILE) as f:
-        mpathconf = [x.strip("\n") for x in f.readlines()]
-    try:
-        first = mpathconf[0]
-        second = mpathconf[1]
-    except IndexError:
-        pass
+        # Read only the lines that are going to be processed
+        first = f.readline().rstrip("\n")
+        second = f.readline().rstrip("\n")
```
nirs, line 419: Since we check using startswith() we don't really need to strip the newline.

lib/vdsm/storage/mpathconf.py
```diff
@@ -130,3 +174,33 @@ def read_blacklist():
...
+    """
+    first = second = ''
+    with open(CONF_FILE) as f:
+        mpathconf = [x.strip("\n") for x in f.readlines()]
+    try:
+        first = mpathconf[0]
+        second = mpathconf[1]
+    except IndexError:
+        pass
+    private = _PRIVATE_TAG in second or _OLD_PRIVATE_TAG in second
```
nirs, line 421: If we add:

    _PRIVATE_TAGS = ("# VDSM PRIVATE", "# RHEV PRIVATE")

We can use:

    private = second.startswith(_PRIVATE_TAGS)

Which is more correct and consistent.

nirs (changes requested):
Awesome!

tests/storage/mpathconf_test.py
aesteve-rh: I had it like that, and changed it following your comment here: https://github.com/oVirt/vdsm/pull/239#discussion_r901147725

Or did I misunderstand the comment?

nirs: Using current tag does not mean we have to use complicated parameters, just include it in the tests that must use the current value.

lib/vdsm/storage/mpathconf.py
aesteve-rh: It looks really nice now that we use the same strategy in all cases, and all conditions have been simplified to use `startswith` :)

tests/storage/mpathconf_test.py
aesteve-rh: Got it.

aesteve-rh:
It should be all covered now. Hopefully this is the one!

I think the last changes with `startswith` was a little addition that has a very nice outcome in the code :)

nirs (approved):

PR #242, opened 2022-06-08
development.md: add coding guidelines

aesteve-rh:
First draft for including coding guidelines in the project.

### DIFF
--- a/doc/development.md
+++ b/doc/development.md
@@ -157,3 +157,179 @@ To add SPDX header to a file with a non-default license:
 Please check that all files are reuse-compliant before pushing your branch:
 
     make reuse
+
+
+## Coding guidelines
+
+The coding conventions in Vdsm are generally aligned with the
+[PEP 8](https://peps.python.org/pep-0008/) style recommendations.
+Code is linted with `flake8` and `pylint` based in the `/.pylintrc` file.
+You can trigger the linters check by pushing changes into your fork, which invokes:
+
+    make lint
+
+However, some code conventions are not explicitly enforced by the linters due
+to their limitations. Nonetheless, new code shall adhere to those
+rules, and old code should be updated as possible when patched.
+
+In general, it is very important for new code to be consistent with the existing
+code in the same module, or in the same package for new modules. In case of
+doubt, check the already existing code before!
+
+In this section, we will cover some of the most important conventions and
+preferences for contributing to Vdsm, as extension to PEP 8 general rules.
+
+### Imports
+
+First import standard library modules, then 3rd party modules, and finally
+project modules.
+Separate standard library modules and 3rd party modules with a blank line.
+Keep alphabetical order within each block.
+
+Less specific imports:
+
+    from vdsm.foo import baz
+
+Should come before more specific imports like:
+
+    from vdsm.foo.bar import baz
+
+Relative name imports are accepted, but better to avoid.
+
+### Indentation
+
+Use 4 spaces per indentation level.
+
+Continuation lines should align wrapped elements either vertically using
+Python's implicit line joining inside parentheses, brackets and braces, or
+using a hanging indent. When using a hanging indent, the following should
+be considered:
+- There should be no arguments on the first line.
+- Further indentation should be used to clearly distinguish itself as a continuation line.
+- Hanging indentation should stick to the 4-space rule.
+```python
+# Add 4 spaces (an extra level of indentation) to distinguish arguments from the rest.
+# Consider that having too many arguments in a function is an anti-pattern.
+foo = long_function_name(
+    var_one, var_two, var_three, var_four)
+
+# Recommended extra indentation with multiline if-statements.
+if (condition_1 and condition_2 [...]
+        and condition_n):
+    do_something()
+
+# But this is also possible.
+def long_function_name(
+        var_one, var_two, var_three,
+        var_four):
+    print(var_one)
+```
+
+To break comprehension Lists, Dicts, Sets as well as generator expressions,
+keep expressions that would be written in a single line together.
+Continuation lines stick to the same rules as above. Closing
+brace/bracket/parenthesis, if placed on a separate line, shall line up
+under the first character of the line start.
+
+```python
+# Vertically aligned and extra 4 spaces for condition continuation
+foo = [var for var in bar_list
+       if condition_1
+           and condition_2
+           and condition_3]
+```
+
+### Docstrings
+
+Always use the three double-quote `"""` format for docstrings
+(per [PEP 257](https://peps.python.org/pep-0257/)).
+A docstring shall be organized as multiple summary lines (one physical line
+not exceeding 79 characters) terminated by a period. First and last line
+of the docstring shall be blank.
+
+Module functions and classes should have a docstring, unless:
+- Not public.
+- Pertain to a test module.
+
+Class methods do not require docstring, but is recommended when it is a public
+entity or have a complex interface.
+
+The docstring should be imperative-style (e.g., `Verify configuration file.`)
+rather than descriptive-style (e.g., `Verifies configuration file.`).
+The docstring should describe the function's main purpose, its calling syntax,
+and/or semantics, but avoid implementation details.
+
+Certain aspects of a function shall be documented in special sections,
+listed below.
+
+#### Args: (or Arguments:)
+
+List each parameter by name. It is strongly encouraged that the type of the
+parameter follows the name, either enclosed in parentheses, with clear wording,
+or in [type annotation](https://docs.python.org/3/library/typing.html) format.
+Then, the description of the parameter separated by a colon. If the type is
+optional or can take multiple types, that can be specified with
+comma-separated lists or use type annotation syntax.
+If the description of the parameter is too long to fit in 79 characters,
+following hanging lines can be indented with 2 or 4 extra spaces
+(as long as it is consistent with surrounding docstrings).
+
+#### Returns: (or Yields: for generators)
+
+Start with the type of the return value, followed by a colon and its semantics.
+If the function only returns None, this section is not required. It may also
+be omitted if the docstring starts with Return or Yield
+(e.g., `Return a boolean indicating if the folder has the right permissions.`),
+and the opening sentence is sufficient to describe the return value.
+
+#### Raises:
+
+List all exceptions that are relevant to the interface, followed by a
+description. Use a similar exception name + (type) + colon + space and hanging
+indent style as described in `Args:`.
+
+```python
+def example_function(arg_1, arg_2=None):
+    """
+    This is an example docstring. Start with summary explanation of the
+    function main purpose. Can expand multiple lines.
+
+    Args:
+        arg_1 (_type_): Long description can be split into
+            multiple lines with extra indentation.
+        arg_2 (_type_): _description_
+
+    Returns:
+        _type_: _description_
+
+    Raises:
+        OSError: _description_
+    """
+```
+
+### Comment blocks and inline comments
+
+Comments are accepted in non-obvious parts of the code that need extra explanation.
+A comment can expand through multiple lines before the operation starts.
+Comments should have normal punctuation and grammar. Start sentences with
+uppercase letters, end with periods, and use commas where necessary.
+Comments should never describe the code.
+
+Inline comments are allowed, but discouraged. Inline comments are acceptable
+for very short comments, but shall never expand through multiple lines.
+To improve legibility, multiple inline comments in successive lines should be
+vertically aligned.
+
+```python
+    _MULTIPATHD = cmdutils.CommandPath(
+        "multipathd",
+        "/usr/sbin/multipathd",  # Fedora, EL7
+        "/sbin/multipathd")      # Ubuntu
+
+    # List of multipath devices that should never be handled by Vdsm. The
+    # main use case is to filter out the multipath devices the host is
+    # booting from when configuring hypervisor to boot from SAN. This device
+    # must have a special rule to queue I/O when all paths have failed, and
+    # accessing it in Vdsm commands may hang Vdsm.
+
+```


doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+The coding conventions in Vdsm are generally aligned with the
+[PEP 8](https://peps.python.org/pep-0008/) standard guidelines.
+Code is linted with `pylint` based in the `/.pylintrc` file.
+You can lint the code before submitting a patch by doing:
+
+    make lint
+
+However, some code conventions are not explicitly enforced by the linter due
+to legacy code and inconsistent styles. But new code shall adhere to those
+rules, and old code, should be updated as possible when patched.
```
mz-pdm, line 105: The punctuation here doesn't seem to match the text.

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+- Hanging indentation should stick to the 4-space rule.
+```python
+# Correct
+# Add 4 spaces (an extra level of indentation) to distinguish arguments from the rest.
+foo = long_function_name(
+        var_one, var_two, var_three, var_four)
+
+# Also with if-statements.
+if (condition_1 and
+        condition_2):
```
mz-pdm, line 149: This style of condition misalignment by definition is horrible but it seems to be quite popular and I've experienced insistence on it in Vdsm reviews.  I have better things to do than arguing about broken indentation so I can live with this rule.

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+def long_function_name(
+        var_one,
+        var_two,
+        var_three,
+        var_four):
+    print(var_one)
+
+# Hanging indents should add a level.
+foo = long_function_name(
+    var_one, var_two, var_three, var_four)
```
mz-pdm, line 162: What's the difference here against the first example?

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+Continuation lines stick to the same rules as above. Closing
+brace/bracket/parenthesis shall line up under the first character of the
+line start.
+
+```python
+# Correct
+# Vertically aligned and extra 4 spaces for condition continuation
+foo = [var for var in bar_list
+       if condition_1 and
+           condition_2]
```
mz-pdm, line 188: 8(

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+
+# Closing bracket starts under the first character
+foo = [
+    var for var in bar_list if condition_1
+]
```
mz-pdm, line 193: I wonder why the difference against function arguments.

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+### Docstrings
+
+Always use the three double-quote `"""` format for docstrings 
+(per [PEP 257](https://peps.python.org/pep-0257/)).
+A docstring should be organized as multiple summary lines (one physical line
+not exceeding 79 characters) terminated by a period. First and last line
+of the docstring shall be blank.
+
+Functions and classes must have a docstring, unless:
+- Not externally visible and very short.
```
mz-pdm, line 213: This should be relaxed to avoid having docstrings for the sake of having docstrings.  Generally, only public (non-test) entities should be required to have docstrings.  Docstrings in some non-public entities may be desirable but it depends on the qualities rather than quantities of the content and comments may be better if it is about explanation what's happening rather than how to call it.

Sometimes docstrings may not be needed even in public entities, for example, it's often impossible to provide a meaningful docstring for `def __init__(self)`.

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+The docstring should describe the function's main purpose, its calling syntax,
+and/or semantics, but avoid implementation details.
+
+Certain aspects of a function should be documented in special sections,
+listed below.
+
+#### Args: (or Arguments:)
+
+List each parameter by name. The type of the parameter may optionally follow
+the name, enclosed in parentheses, and then the description of the parameter
```
mz-pdm, line 227: Providing type information for all the parameters and return values should be strongly encouraged.  It can take several forms: the type name in parentheses, clear wording in the argument description, or type annotation.

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+```python
+def example_function(arg_1, arg_2=None):
+    """
+    This is an example docstring. Start with summary explanation of the
+    function main purpose. Can expand multiple lines.
+
+    Args:
+        arg_1 (_type_): Long description can be split into
+            multiple lines with extra indentation.
+        arg_2 (_type_, optional): _description_. Defaults to None.
```
mz-pdm, line 258: That the argument is optional with None as the default value is already clear from the function declaration.

doc/development.md
```diff
@@ -89,4 +89,205 @@
...
+```python
+    _MULTIPATHD = cmdutils.CommandPath("multipathd",
+                                       "/usr/sbin/multipathd",  # Fedora, EL7
+                                       "/sbin/multipathd")      # Ubuntu
+
+    # List of multipath devices that should never be handled by vdsm. The
+    # main use case is to filter out the multipath devices the host is
+    # booting from when configuring hypervisor to boot from SAN. This device
+    # must have a special rule to queue I/O when all paths have failed, and
+    # accessing it in vdsm commands may hang vdsm.
```
mz-pdm, line 291: Shouldn't "vdsm" start with a capital letter?

mz-pdm (commented):
Thank you for putting it together.  Some comments inside.

doc/development.md
aesteve-rh: Nothing here is written in stone, if there is a place for arguing is now :) The target is that we all reach an agreement and are comfortable with the result. I know, ambitious goal :P

That said, the reasoning for the extra 4 spaces here is to distinguish it from the code inside the if-statement. Text from PEP:
```
(...) the combination of a two character keyword (i.e. if), plus a single space, plus an opening parenthesis
creates a natural 4-space indent for the subsequent lines of the multiline conditional.
This can produce a visual conflict with the indented suite of code nested inside the if-statement,
which would also naturally be indented to 4 spaces.
```
However, PEP does not enforce it, and thus we can also not enforce it if we don't like it.

doc/development.md
aesteve-rh: Nothing really, is just depicting the two possible options for breaking arguments into multiple lines. The first example is preferred if the arguments will break into more than one line. This second example is preferred when arguments do not fit in the same line as the function definition, but they fit in a single line with the hanging indent.

doc/development.md
aesteve-rh: These extra 4 spaces could easily be omitted, since there are no `inside the if-statement` lines. I wrote the 4 spaces here as a visual clue that the line is a continuation of the conditional from the previous line. Let's see what the rest of the people thinks, but IMO this can be removed if we don't like them.

doc/development.md
aesteve-rh: I mean, more or less you listed the same exceptions as I wrote:
```
- Not externally visible and very short.
- Pertains to a test module.
```

Also, here I purposely wrote about functions, not class methods. Class methods require docstring only when they are not short/obvious and externally visible. Therefore, the constructor of a class does not need a docstring, I agree. Nor do, in general, the getters/setters for class attributes, for example. What I meant for classes is that the class itself deserves a docstring:
```
class RandomClass():
    """
   This class needs a docstring.
    """
```
I will relax a bit the text and add some example.

doc/development.md
aesteve-rh: Yeah... I copied the comment from the project because it looked like a nice example of a proper multi-line comment, but didn't notice that. I'll fix it here in the example or remove the sentence entirely.

doc/development.md
mz-pdm: OK, I think it'd be better to be consistent in both the cases but I don't insist on it.  Anyway, the difference in the examples is confusing as it is now, it should be clarified in the document.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+Inline comments are allowed, but discouraged. Inline comments are acceptable
+for very short comments, but shall never expand through multiple lines.
+To improve legibility, inline comments should start at least 2 spaces away from
+the code. Multiple inline comments in successive lines should be vertically
+aligned.
+
+```python
+    _MULTIPATHD = cmdutils.CommandPath("multipathd",
+                                       "/usr/sbin/multipathd",  # Fedora, EL7
+                                       "/sbin/multipathd")      # Ubuntu
```
mz-pdm, line 285: Doesn't the indentation here contradict the suggested guidelines? :-)

aesteve-rh: True, violates the "there should be no arguments on the first line" point. You caught me :)

I'll change it!

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+    from vdsm.foo import baz
+
+Should come before more specific imports like:
+
+    from vdsm.foo.bar import baz
+
+Accepted, but better to avoid:
+- Relative name imports
+- Import many modules in a comma separated list
+- Alias an imported module with non-standard abbreviation
```
nirs, line 128: This is actually very common in storage cod, for example:

    from vdsm.storage import exception as se

So we should describe this kind of imports. New code should use the same style - the storage exception module should be important in the same way in the storage package.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+You can lint the code before submitting a patch by doing:
+
+    make lint
+
+However, some code conventions are not explicitly enforced by the linter due
+to legacy code and inconsistent styles. But new code shall adhere to those
+rules, and old code, should be updated as possible when patched.
+
+In this section, we will cover some of the most important conventions and
+preferences for contributing to Vdsm.
```
nirs, line 108: Before we deal with details, we need to give the general rules that apply to everything:

When modifying existing module, use the style of the current module. When adding a new module, use the style used by other modules in this package.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+Use 4 spaces per indentation level.
+
+Continuation lines should align wrapped elements either vertically using
+Python's implicit line joining inside parentheses, brackets and braces, or
+using a hanging indent. When using a hanging indent, the following should
+be considered:
+- There should be no arguments on the first line.
+- Further indentation should be used to clearly distinguish itself as a continuation line.
+- Hanging indentation should stick to the 4-space rule.
+```python
```
nirs, line 141: This is not clear. It would be useful to discus special cases with examples:
- How to define functions with many arguments
- How to call function with many arguments
- How to call function with positional and keyword arguments
- When to use conditional expressions (a if condition else b)
- How to format list comprehension

doc/development.md
nirs: There is no problem to work around :-)

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+        condition_2):
+    do_something()
+
+# One line per argument for cases where arguments expand to more than one line.
+def long_function_name(
+        var_one,
+        var_two,
+        var_three,
+        var_four):
+    print(var_one)
```
nirs, line 158: We don't use this style. If a function definition does not fit in one line we do this:

```
    def __init__(self, domain, image, parent, capacity, format, type, voltype,
                 disktype, description="", legality=sc.ILLEGAL_VOL, ctime=None,
                 generation=sc.DEFAULT_GENERATION,
                 sequence=sc.DEFAULT_SEQUENCE):
```

But in general we don't have functions with so many arguments and it an anti-pattern to add such functions. They are hard to work with regardless of how you indent them. The coding style guide should discourage adding lot of arguments.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+
+# Hanging indents should add a level.
+foo = long_function_name(
+    var_one, var_two, var_three, var_four)
+```
+```python
+# Wrong:
+# Arguments on first line forbidden (even when vertically aligned).
+foo = long_function_name(var_one, var_two,
+                         var_three, var_four)
```
nirs, line 168: I would not describe this as wrong, and we have some legacy code using this style. We can describe that this is used in legacy code and should not be used in new code.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+
+# Further indentation required as indentation is not distinguishable.
+def long_function_name(
+    var_one, var_two, var_three, var_four):
+    print(var_one)
```
nirs, line 173: No need to show this, just show the recommended style.

doc/development.md
nirs: I don't think we have any code using hanging indent in list comprehension. If you to do this your code is too complex and should be simplified. This is another generic rule that should be discuss at the start of the section. Don't fight formatting complex code, write simple code that is easy to format.

doc/development.md
nirs: I don't think we have such code and there is no reason to use this formatting in this
example.

Either it fits in one line (best):

    [x for x, y in things if x]

Or you break on the condition(s):

    [x for x, y in things_that_do_not_fit()
     if x and y.startswith("foo")]

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+]
+```
+
+```python
+# Wrong
+# for-statement broken in multiple lines
+foo = [var for var
+       in bar_list if condition_1 and
+       condition_2]
+```
```
nirs, line 202: Don't show the bad examples, only the good examples. Will save time for everyone.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+
+### Docstrings
+
+Always use the three double-quote `"""` format for docstrings 
+(per [PEP 257](https://peps.python.org/pep-0257/)).
+A docstring should be organized as multiple summary lines (one physical line
+not exceeding 79 characters) terminated by a period. First and last line
+of the docstring shall be blank.
+
+Functions and classes must have a docstring, unless:
```
nirs, line 212: Replacing "must" with "should" can simplify this text. Something like "should have a docstring if the docstring will be helpful" is a good simple rule.

doc/development.md
nirs: I agree, "optional" and "Defaults to None" are not needed.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+    """
+```
+
+### Comment blocks and Inline comments
+
+Comments are accepted in tricky parts of the code that need extra explanation.
+A comment can expand through multiple lines before the operation starts.
+Comments should have normal punctuation and grammar. Start sentences with
+uppercase letter, end with periods, and use commas where necessary.
+Comments should never describe the code.
```
nirs, line 274: Comments are not only for tricky parts. They explain why the code is written a specific way, and add info that cannot be added by the code itself.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+
+Inline comments are allowed, but discouraged. Inline comments are acceptable
+for very short comments, but shall never expand through multiple lines.
+To improve legibility, inline comments should start at least 2 spaces away from
+the code. Multiple inline comments in successive lines should be vertically
+aligned.
```
nirs, line 280: flake8 already enforce the 2 spaces before inline comments so we don't need to repeat it here.

nirs (commented):
Lets try to make the coding guidlines much shorter, and focus on the general rules instead of specific details.

doc/development.md
aesteve-rh: I'd say I added this style from a review, but ok, I'll remove it. I agree with many arguments as an antipattern, and I'll add the comment. But it can be that does not fit a single line due to long variable names and default values being used, so it is still interesting to comment how to properly break the line.

doc/development.md
aesteve-rh: That's one of the main points of these coding guidelines, avoid perpetuating old style that should not be used in new code. I already mention in the beginning of the section that what we describe here are rules that cannot be enforced because we still use it in legacy code. I'll rewrite it a bit to make it more clear.

doc/development.md
aesteve-rh: Ok, I'll remove this example.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+Less specific imports:
+
+    from vdsm.foo import baz
+
+Should come before more specific imports like:
+
+    from vdsm.foo.bar import baz
+
+Accepted, but better to avoid:
+- Relative name imports
```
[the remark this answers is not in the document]
aesteve-rh: I think that absolute imports are more readable, makes easier to locate the imported modules.

At least Python 3 enforces explicit relative imports, which deals with the more problematic case. But still, I don't see a good reason to use relative imports either. The structure of the project is not as complex as to make it really appealing. In any case, in the guidelines they are not forbidden, but I think it is fine to discourage them.

Anyway, if you still think I should remove the relative import from the list to avoid, I'll do it, I don't really have a strong preference.

doc/development.md
```diff
@@ -89,4 +89,205 @@ merged, we need to test them on all supported distributions and
...
+- Import many modules in a comma separated list
```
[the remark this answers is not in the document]
aesteve-rh: That point can be used to invalidate most of the guidelines, lol.

I'll remove the item from the list.

doc/development.md
aesteve-rh: That can be one of the definitions of `tricky code`: code that needs to be explained further because the code itself does not provide the entire information.

If you dislike the word `tricky` I can look for something better. I think the idea is more or less clear, though: comments that are easy to follow with the code are redundant and unnecessary, only make comments that add value to the code.

nirs: Code that needs comments it not tricky. Maybe use "not obvious" instead?

aesteve-rh:
I'll push an updated version after https://github.com/oVirt/vdsm/pull/255 gets integrated :)

doc/development.md
aesteve-rh: I think I will relax it to a recommendation only, no need to enforce it really.

To work around it, I think would favor evaluating the conditions separately:
```python
condition_name = condition_1 and condition_2 [and ...]
if condition_name:
    do_something()
```
Which can be split into multiple statements, or multiple lines, easily.

doc/development.md
aesteve-rh: I think many of these points are already covered in the text and examples.

aesteve-rh:
Shortened the text and relaxed some statements + removed bad examples. Let's give it another iteration :)

doc/development.md
mz-pdm: I can still see "vdsm" rather than "Vdsm" here.

doc/development.md
```diff
@@ -120,3 +120,180 @@ configuration in the `.github/workflows/ci.yml` file.
...
+Code is linted with `flake8` and `pylint` based in the `/.pylintrc` file.
+You can trigger the linters check by pushing changes into your fork, which invokes:
+
+    make lint
+
+However, some code conventions are not explicitly enforced by the linters due
+to their limitations. Nonetheless, new code shall adhere to those
+rules, and old code should be updated as possible when patched.
+
+In general, is very important for new code to be consistent with the existing
```
mz-pdm, line 138: In general, it is ...

doc/development.md
```diff
@@ -120,3 +120,180 @@
...
+# Vertically aligned and extra 4 spaces for condition continuation
+foo = [var for var in bar_list
+       if condition_1
+           and condition_2
+           and condition_3]
+```
+
+### Docstrings
+
+Always use the three double-quote `"""` format for docstrings
```
mz-pdm, line 207: Extra space at the end of line.

doc/development.md
```diff
@@ -120,3 +120,180 @@
...
+        var_one, var_two, var_three,
+        var_four):
+    print(var_one)
+```
+
+To break comprehension Lists, Dicts, Sets as well as generator expressions,
+keep expressions that would be written in a single line together.
+Continuation lines stick to the same rules as above. Closing
+brace/bracket/parenthesis shall line up under the first character of the
+line start.
```
mz-pdm, line 195: But in the following example the closing bracket is at the end of line.  Which is all right, just confusing with respect to this sentence.  It might be clarified by saying e.g. "Closing brace/bracket/parenthesis, if placed on a separate line, shall line up ...".

doc/development.md
```diff
@@ -120,3 +120,180 @@
...
+List each parameter by name. It is strongly encouraged that the type of the
+parameter follows the name, either enclosed in parentheses, with clear wording,
+or in [type annotation](https://docs.python.org/3/library/typing.html) format).
+Then the description of the parameter
+separated by a colon. If the type is optional or can take multiple types,
+that can be specified with comma-separated lists or use
+type annotation syntax.
+If the description of the parameter is too long to
+fit in 79 characters, following hanging lines are indented with 4 extra
+spaces.
```
mz-pdm, line 239: Are the 4 spaces mandated by something generally accepted or suggested just here?  I find 4, rather than 2, spaces excessive, both visually and as for taking from the available line space.

doc/development.md
```diff
@@ -120,3 +120,180 @@
...
+
+    Returns:
+        _type_: _description_
+
+    Raises:
+        OSError: _description_
+    """
+```
+
+### Comment blocks and Inline comments
```
mz-pdm, line 274: Why is "Inline" capitalized?

doc/development.md
```diff
@@ -120,3 +120,180 @@
...
+
+Comments are accepted in non-obvious parts of the code that need extra explanation.
+A comment can expand through multiple lines before the operation starts.
+Comments should have normal punctuation and grammar. Start sentences with
+uppercase letter, end with periods, and use commas where necessary.
```
mz-pdm, line 279: uppercase letters, ...

mz-pdm (commented):
Thanks for the changes, looks mostly good, just some typos and a couple of minor comments.

doc/development.md
aesteve-rh: https://developers.google.com/edu/python/introduction?csw=1#indentation

Basically this: 4 spaces is widely accepted in the python community + is recommended by the PEP 8 style guide. But there are other options that use 2 spaces instead, and it is equally fine.

IMO whatever we decide, it should be mandated to keep the code consistent. Inconsistent indentation from within the same project is evil. I'd vote for 4 spaces, as is already the most commonly indentation used in the project files. But I'm open to debate.

aesteve-rh: Ah, you meant only the documentation (my bad). Here we can be more flexible, I assume. I don't have a strong opinion. I can relax it to a recommendation and comment that 2 spaces are also ok (as long as it is consistent with surrounding code).

mz-pdm: > I can relax it to a recommendation and comment that 2 spaces are also ok 

OK.

> (as long as it is consistent with surrounding code).

... surrounding docstrings.

doc/development.md
aesteve-rh: No reason. Fixed.

mz-pdm (dismissed):

aesteve-rh:
@nirs @tinez @mz-pdm Can you ack? If we still want to include this, anyway.

doc/development.md
```diff
@@ -157,3 +157,180 @@ To add SPDX header to a file with a non-default license:
...
+and/or semantics, but avoid implementation details.
+
+Certain aspects of a function shall be documented in special sections,
+listed below.
+
+#### Args: (or Arguments:)
+
+List each parameter by name. It is strongly encouraged that the type of the
+parameter follows the name, either enclosed in parentheses, with clear wording,
+or in [type annotation](https://docs.python.org/3/library/typing.html) format).
```
mz-pdm, line 269: Extra parenthesis at the end.

doc/development.md
```diff
@@ -157,3 +157,180 @@
...
+Then the description of the parameter
+separated by a colon. If the type is optional or can take multiple types,
+that can be specified with comma-separated lists or use
+type annotation syntax.
+If the description of the parameter is too long to
+fit in 79 characters, following hanging lines can be indented with 2 or 4
+extra spaces (as long as it is consistent with surrounding docstrings).
+
```
mz-pdm, line 277: Cosmetic: The paragraphs above should be reformatted for a better alignment.

mz-pdm (commented):
Looks good, just two cosmetic issues.

mz-pdm (approved):

Issue #237, opened 2022-06-07
lvm: check VG autoactivation attribute after updating the LVMCache


aesteve-rh:
VGs in LVM are created with disabled `autoactivation` in Vdsm to avoid unwanted activation of VG/LVs when the autoactivation option (`-aay`) is used. Autoactivation is used, e.g., during system boot.

However, this configuration can be overridden, either in the `lvm.conf` or by manually invoking a command that overwrites the VG settings. Vdsm must ensure that all VGs keep this attribute disabled.

When we update the VGs in the LVMCache, we should check for the setautoactivation attribute of the VGs. If any VG has the attribute enabled, log a warning and deactivate it.

PR #244, opened 2022-06-09
lvm: make SPM host check for VGs autoactivation field

aesteve-rh:
Add a check for the autoactivation attribute of the VGs. For each VGs in the LVMCache:

1. Obtain the autoactivation field of each VG
   a. If it fails (LVM exception), raise `VolumeGroupDoesNotExist`
2. Check if the retrieved autoactivation is enabled, and if so:
   a. Log a warning and try to disable it
   b. If it fails at disabling the option, raise a `VolumeGroupDisableOptionError` <-- __new exception__ (will it cause problems in the engine side raising a new exception type? @bennyz @barpavel )

Changing VGs can be done only by the SPM host. Other hosts cannot fix this issue.
Add a flow on the SPM checking VG configuration and fixing it:
- When starting the SPM. All VGs available in the host are checked.
  - If an error occurs (e.g., LVM error), log a warning but do not raise.
  - Since we check all VGs the error may be unrelated (e.g., stall LVMCache), and raising would break the whole SPM start flow.
- When activating a storage domain. Only the associated VG is checked.
  - If an error occurs, we raise.
  - In this case it will fail to activate the volume from the engine, and it’ll have to be retried. Either way, if the VG is not available in the LVM, the SD activation would fail nonetheless.

_Does it affect detach storage domain and attaching to an older system?_

No real change has done to the SD. It should not cause any trouble. The only thing to consider may be the new Exception added. But given that the new checks will not run in older systems, they will not trigger the Exception. So to my understanding it shall have back compatibility.

_How this change should be tested?_

Added lvm and blocksd tests to verify the new flow.

It may also be nice to verify it in a VM (not done yet):
- Have a cluster with hosts in maintenance mode
- Manually enable autoactivation in one of the SDs in the host (will the VG be available in maintenance mode?)
- Activate a host, it will start the SPM and log a warning.
- Detach the SD.
- Enable autoactivation for the detached SD.
- Reattach the SD, it will succeed and log a warning.

Depends on https://github.com/oVirt/vdsm/pull/234 (should be integrated first) -- _DONE_.
Fixes: https://github.com/oVirt/vdsm/issues/237

### DIFF
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -1440,6 +1440,62 @@ def getAllVGs():
     return _lvminfo.getAllVgs()  # returns list
 
 
+def check_single_vg_autoactivation(vg_name):
+    """
+    Retrieve the autoactivation field of vg_name and check if is disabled.
+    For each VG with autoactivation enabled, log a warning and disable it.
+    The autoactivation field should remain always disabled to avoid unexpected
+    VG activation.
+
+    Args:
+        vg_name (str): name of the VG to check.
+
+    Raises:
+        se.VolumeGroupDoesNotExist: in case it fails at retrieving the
+            autoactivation field of a VG.
+        se.VolumeGroupDisableOptionError: in case it fails at disabling the
+            autoactivation field of a VG.
+    """
+    cmd = ["vgs", "-o", "autoactivation", vg_name]
+    try:
+        out = _lvminfo.run_command(
+            cmd, devices=_lvminfo._getVGDevs([vg_name]))
+    except se.LVMCommandError as e:
+        log.warning(
+            "Retrieving VG %s autoactivation field failed", vg_name)
+        # Retrieving options for a single VG failed after a recent update.
+        # Assume VG does not exist and raise.
+        raise se.VolumeGroupDoesNotExist.from_error(vg_name, e)
+    else:
+        if 'enabled' in (line.strip() for line in out):
+            log.warning(
+                "Found autoactivation enabled for VG %s, disabling",
+                vg_name)
+            _disable_vg_autoactivation(vg_name)
+
+
+def check_vgs_autoactivation():
+    """
+    Loop through all VGs, and check if autoactivation field is disabled.
+    """
+    for vg in getAllVGs():
+        try:
+            check_single_vg_autoactivation(vg.name)
+        except se.VolumeGroupDoesNotExist as e:
+            log.warning("Failed to check VG %s autoactivation: %s", vg.name, e)
+        except se.VolumeGroupDisableOptionError as e:
+            log.warning(
+                "Failed to disable VG %s autoactivation: %s", vg.name, e)
+
+
+def _disable_vg_autoactivation(vg_name):
+    cmd = ["vgchange", "--setautoactivation", "n", vg_name]
+    try:
+        _lvminfo.run_command(cmd, devices=_lvminfo._getVGDevs((vg_name, )))
+    except se.LVMCommandError as e:
+        raise se.VolumeGroupDisableOptionError.from_lvmerror(e)
+
+
 # TODO: lvm VG UUID should not be exposed.
 # Remove this function when hsm.public_createVG is removed.
 def getVGbyUUID(vgUUID):
--- a/lib/vdsm/storage/sp.py
+++ b/lib/vdsm/storage/sp.py
@@ -42,6 +42,7 @@ from vdsm.storage import fileSD
 from vdsm.storage import fileUtils
 from vdsm.storage import guarded
 from vdsm.storage import image
+from vdsm.storage import lvm
 from vdsm.storage import mailbox
 from vdsm.storage import merge
 from vdsm.storage import misc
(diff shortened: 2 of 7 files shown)


nirs:
> Add a flow on the SPM checking VG configuration and fixing it:
> 
> * when starting the SPM
> * when activating a storage domain

This makes sense, but there we need to consider more aspects.

- Logging warning when finding a VG that need to fixed
- How errors are handled if fixing a VG fails? Do we fail SPM start or activating a storage domain? This can bring down entire system if our VG update code does not work in some condition.
- Does it affect detach storage domain and attaching to an older system?
- How this change should be tested?

I did not look in the code yet, but lets make the intent of this change more clear.

aesteve-rh:
Regarding these two points:

- Does it affect detach storage domain and attaching to an older system?

I do not have yet enough insight to be able to foresee the outcome of this scenario. Maybe we can discuss it together, or I can try to test it, enable all logs (now that I just learnt how to do it) and see how it goes.

- How this change should be tested?

The LVM test is the easy part. But for the SPM I am not sure if it would require for me to verify it in a real setup with VMs, or I can reproduce real scenarios adding new tests to the `blocksd_test.py`, I'll need to check.

aesteve-rh:
Updated text the Issue with some considerations following @nirs comment.

I will not ready up the PR yet, but maybe you (also @bennyz and @barpavel ) can take a look at the text and let me know if there are any weak points or fishy parts that I need to review/rework.

lib/vdsm/storage/lvm.py
```diff
@@ -1440,6 +1440,62 @@ def getAllVGs():
     return _lvminfo.getAllVgs()  # returns list
 
 
+def check_single_vg_autoactivation(vg_name):
+    """
+    Retrieve the autoactivation field of vg_name and check if is disabled.
```
barpavel, line 1445: `if is` --> `if it is`

lib/vdsm/storage/lvm.py
```diff
@@ -1440,6 +1440,62 @@ def getAllVGs():
...
+        se.VolumeGroupDisableOptionError: in case it fails at disabling the
+            autoactivation field of a VG.
+    """
+    cmd = ["vgs", "-o", "autoactivation", vg_name]
+    try:
+        out = _lvminfo.run_command(
+            cmd, devices=_lvminfo._getVGDevs([vg_name]))
+    except se.LVMCommandError as e:
+        log.warning(
+            "Retrieving VG %s autoactivation field failed", vg_name)
```
barpavel, line 1465: Maybe make it `'%s'` or `%r` to make the name easier found in the log message? Nir?
If so, relevant for other logs as well.

lib/vdsm/storage/lvm.py
```diff
@@ -1440,6 +1440,62 @@ def getAllVGs():
...
+        # Retrieving options for a single VG failed after a recent update.
+        # Assume VG does not exist and raise.
+        raise se.VolumeGroupDoesNotExist.from_error(vg_name, e)
+    else:
+        if 'enabled' in (line.strip() for line in out):
+            log.warning(
+                "Found autoactivation enabled for VG %s, disabling",
```
barpavel, line 1472: Suggestion: `disabling` --> `disabling it`

lib/vdsm/storage/sp.py
```diff
@@ -1205,6 +1210,11 @@ def activateSD(self, sdUUID):
         if dom.getDomainClass() == sd.DATA_DOMAIN:
             self._convertDomain(dom)
 
+        # If running as SPM, check that autoactivation is disabled for all VGs
+        # before activating the SD.
```
barpavel, line 1214: AFAIR `SD` is "engine" language, while `Pool` is used in VDSM :)
Nir, please express your opinion.

tests/storage/lvm_test.py
```diff
@@ -1694,6 +1694,32 @@ def test_vg_no_autoactivation(tmp_storage):
...
+    lvm.createLV(vg_name, lv2_name, 1024, activate=False)
+
+    # Enable autoactivation for the VG.
+    commands.run(
+        ["vgchange", "--setautoactivation", "y", "--devices", dev, vg_name])
+
+    # Check should detect the VG with enabled autoactivation and disable it.
+    lvm.check_vgs_autoactivation()
+
+    # LVM command to autoactivate all LVs of the VG, but is disabled.
```
barpavel, line 1717: `..., but is disabled.` I think the comment better be rephrased, or I didn't understand it, sorry then :)

barpavel:
@aesteve-rh Code written very nice & clean :):):)

barpavel:
Regarding new `VolumeGroupDisableOptionError` error.
I never had to add a new error, but looking at `CannotCreateLogicalVolume` as an example I see it's declared in `org.ovirt.engine.ui.frontend.VdsmErrors` class, `VdsmErrors.properties` and in few other places, so I guess there might be some work on engine side about it. @bennyz

tests/storage/lvm_test.py
aesteve-rh: True, can be better phrased, I'll update it :)

aesteve-rh:
Drop the PR because it is not relevant anymore.

PR #245, opened 2022-06-09
lvmlocal.conf: Disable multipath_wwids_file usage

nirs:
In RHEL 8.6, LVM added detection of multipath components using the
multipath wwids file (/etc/multipath/wwids). Unfortunately this
incorrectly detect blacklist multipath devices. This causes
vgimportdevices to skip the devices when adding to the devices file, and
the result is host failing to boot.

Since we always use lvm devices or filter, this feature is not helpful
for our use case, and can be safely disabled.

Bug-Url: https://bugzilla.redhat.com/2090169
Related-to: https://bugzilla.redhat.com/2076262


nirs:
@sandrobonazzola how can we test this with node upgrade? I think we need to build upgrade
with vdsm version including this fix, and try to upgrade existing node (e.g. 4.5.0, 4.4.x).

aesteve-rh (approved):

vjuranek (approved):

nirs:
@arachmani can you help with testing this change with node upgrade?

PR #248, opened 2022-06-13
virt: fix graphics configuration when resuming from hibernation

nyoxi:
Somehow libvirt drops `passwdValidUntil` attribute from domain XML when
when we pass XML that has no `passwd` attribute to restoreFlags(). Same
happens when restoring from snapshot with memory and we can use the same
method to mitigate the problem.

An alternative would be to pass `VIR_DOMAIN_SAVE_IMAGE_XML_SECURE` flag
to `saveImageGetXMLDesc()` call. That way libvirt would not hide the
`passwd` from us. But since our passwords to graphics console are
temporary this does not add any value. Also we would leak the console
password to the logs which is not favorable even though the password is
likely to be invalid at the time of resume.

Signed-off-by: Tomáš Golembiovský <tgolembi@redhat.com>

### DIFF
--- a/lib/vdsm/virt/vm.py
+++ b/lib/vdsm/virt/vm.py
@@ -2827,7 +2827,6 @@ class Vm(object):
                     if 'vmName' in self.conf:
                         srcDomXML = self._correct_disk_volumes_from_conf(
                             srcDomXML)
-                    srcDomXML = self._correctGraphicsConfiguration(srcDomXML)
                 else:
                     srcDomXML = self._connection.saveImageGetXMLDesc(fname)
                     # Modify CPU pinning -- remove pinning saved during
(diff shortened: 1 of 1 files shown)


lib/vdsm/virt/vm.py
```diff
@@ -2837,6 +2836,15 @@ def _run(self):
                     dom = cpumanagement.replace_cpu_pinning(
                         self, dom, self._initial_vcpupin)
                     srcDomXML = xmlutils.tostring(dom)
+                # We need to set `passwd` attribute to graphics devices
+                # otherwise libvirt would remove the `passwdValidTo` attribute
+                # and that would cause other errors in our validation code
+                # elesewhere. We might call `saveImageGetXMLDesc()` with
```
mz-pdm, line 2842: elsewhere

lib/vdsm/virt/vm.py
```diff
@@ -2837,6 +2836,15 @@
...
+                # `VIR_DOMAIN_SAVE_IMAGE_XML_SECURE` flag to retrive the
```
mz-pdm, line 2843: retrieve

nyoxi:
> Will it also work when SASL is used with VNC and no passwords in the domain XML should be used? I assume that yes since it has already been used for snapshots, just to be sure.

My guess would be that this flow is broken. It seems to me that `_correctGraphicsConfiguration()` is missing a check for such cases. But this is just my assumption and I may be missing something.

nyoxi:
> In which sense broken? The presence of expired passwords in the domain XML may be harmless in such a case. Would you verify what happens in such a case? I think this configuration is used when FIPS is enabled.

The way I understand the comments is that when `passwdValidTo` is not empty libvirt enforces password checking. Which is something you don't want with SASL. But I have to admit I never used this configuration and I don't have an environment o test it.

mz-pdm:
> The way I understand the comments is that when `passwdValidTo` is not empty libvirt enforces password checking. Which is something you don't want with SASL.

This doesn't sound very good but without this patch it seems to be even worse.  So let's merge this one and fix the rest later if needed.

> But I have to admit I never used this configuration and I don't have an environment o test it.

It's not that difficult, I can help with setting it up if needed.  And it's available out of the box in OST environment.  Please don't give up on checking it, let's make it working completely when we are already on it.

mz-pdm (approved):
OST failed for unrelated reasons and passed in #251 followup.

PR #251, opened 2022-06-14
virt: Replace CPU pinning when resuming snapshot with memory 

nyoxi:
When resuming a live snapshot with memory the domain XML contains CPU
pinning that has been used at the time the snapshot was taken. Such
pinning is likely to be invalid when the snapshot is resumed. Engine
passes new CPU pinning configuration in API and it should be honored.


mz-pdm (commented):
OK, let's just wait until #248 is merged.

mz-pdm (approved):

Issue #252, opened 2022-06-15
Fix connect_all error log 


bennyz:
https://github.com/oVirt/vdsm/blob/d43d31f5a1f165bcff68ac9b4d38645c2761015f/lib/vdsm/storage/storageServer.py#L597-L598

Message is missing "not", and  should be: `"Could not configure connection to %s and iface %s"`, or maybe switch to f-strings

aesteve-rh:
Logging package is [optimized](https://docs.python.org/3/howto/logging.html#optimization) for % formatting, the advantage being that formatting the parameters is deferred until the log is actually going to be printed, and only if it is. 

Otherwise the proposed fix looks fine.

PR #254, opened 2022-06-15
storageServer: fix and improve iscsi connect error logging

aesteve-rh:
When seting up an iscsi node for a connection fails, the error logged had a typo.

Fix and improve it by adding the reason of the error to the logs.

Fixes: https://github.com/oVirt/vdsm/issues/252

### DIFF
--- a/lib/vdsm/storage/storageServer.py
+++ b/lib/vdsm/storage/storageServer.py
@@ -602,6 +602,9 @@ class IscsiConnection(Connection):
             else:
                 logins.append(con)
 
+        if not logins:
+            return results
+
         # Run login to nodes in parallel. This operations happen on remote
         # iscsi server and if the some targets are not available, the operation
         # can take quite some time (by default 120 seconds) and can cause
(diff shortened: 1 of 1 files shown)


nirs (approved):
It would be nice to have a test for the case when setting up all nodes fails.

aesteve-rh:
> It would be nice to have a test for the case when setting up all nodes fails.

Yes, that would be nice. The hsm_connect_test has a lot of fakery in it, the Connection factory is overwritten to only return faked connections, so the code that is modified in this branch is completely uncovered.

I don't know if it is like that because it is complex to test or because it is covered by OST, but I'll check and, if possible, add some tests with real data.

nirs:
> I don't know if it is like that because it is complex to test or because it is covered by OST, but I'll check and, if possible, add some tests with real data.

OST does not test negative flows, so this code is not covered. What we need is a test in the storage server layer, simulating all the error conditions and making sure the code does the right hing.

Ideally we would have functional test with a real iscsi server in the same way we test
qemu-img by running, without faking anything.

PR #255, opened 2022-06-15
Clarified the development and tests md.

rokkbert:
Clarified the order of things in the development docs.

Bug-Url: not a bug, has no associated ticket
Signed-off-by: Rob Linden rlinden@redhat.com

### DIFF
--- a/doc/development.md
+++ b/doc/development.md
@@ -2,13 +2,6 @@
 
 ## Environment setup
 
-Fork the project on https://github.com/oVirt/vdsm.
-
-Clone your fork:
-
-    sudo dnf install -y git
-    git@github.com:{user}/vdsm.git
-
 Enable oVirt packages for Fedora:
 
     sudo dnf copr enable -y nsoffer/ioprocess-preview
@@ -24,34 +17,65 @@ Update the system after enabling all repositories:
 
     sudo dnf update -y
 
+Fork the project on https://github.com/oVirt/vdsm.
+
+Clone your fork:
+
+    sudo dnf install -y git
+    git clone git@github.com:{your_username}/vdsm.git
+
 Install additional packages for Fedora, CentOS, and RHEL:
 
     sudo dnf install -y `cat automation/check-patch.packages`
 
-Create virtual environment for vdsm:
+Generate the Makefile (and configure script):
+
+    ./autogen.sh --system --enable-timestamp
+
+Now you can create the virtual environment
+(https://docs.python.org/3/library/venv.html), which is necessary to run the
+tests later. This needs to be done only once:
 
     make venv
 
 
 ## Building Vdsm
 
-To configure sources (run `./configure --help` to see configuration options):
+Before building, it is recommended to recreate the Makefile because it
+contains version numbers, which might have changed by updating the local
+repository:
 
-    git clean -xfd
     ./autogen.sh --system --enable-timestamp
+
+To build Vdsm:
+
     make
 
-To test Vdsm (refer to tests/README for further tests information):
-
-    make check
-
-To create an RPM:
+To create the RPMs:
 
     make rpm
 
-To upgrade your system with local build's RPM:
+To upgrade your system with local build's RPM (before you do this you should
+activate maintenance mode for Vdsm):
 
     make upgrade
+
+
+## Running the tests
+
+To run tests, first enter the virtual environment:
+
+    source ~/.venv/vdsm/bin/activate
+
+Then start some tests with tox, for example the networking tests:
+
+    tox -e network
+
+To exit the virtual environment afterwards:
+
+    deactivate
+
+For more information about testing see [/tests/README.md](/tests/README.md).
 
 
 ## Making new releases
@@ -90,3 +114,9 @@ architectures.
 
 When you push patches to GitHub, CI will run its tests according to the
 configuration in the `.github/workflows/ci.yml` file.
+
+
+## Advanced Configuration
+
+Before running `make` you could use `./configure` to set some (rarely used) options.
+To see the list of options: `./configure -h`.
--- a/tests/README.md
+++ b/tests/README.md
@@ -13,15 +13,26 @@ Note: Not all tests have been converted to tox and/or pytest yet.
 
 ## Running the tests
 
-Before running vdsm tests, activate the environment:
+Before running Vdsm tests, activate the virtual environment (which you need to
+create once, with `make venv`, as described in
+[/doc/development.md](/doc/development.md)):
 
     source ~/.venv/vdsm/bin/activate
 
-When done, you can deactivate the environment:
+When done with testing, you can deactivate the virtual environment:
 
     deactivate
 
-Before running storage tests, make sure you have setup the user storage
+To run tests from only one module, for example `lib`:
+
+    tox -e lib
+
+The test verbosity can be increased. Any arguments after `--` are passed
+through to pytest:
+
+    tox -e lib -- -vv
+
+Before running storage tests, make sure you have set up the user storage
 accordingly. To create storage you can do:
 
     make storage
@@ -32,20 +43,29 @@ for more information.
 
 Run storage tests:
 
+    tox -e storage -- -vv --no-cov storage
+
+Note that for some test modules (like `storage` or `virt`), _if_ there
+are any extra pytest arguments (after the `--`) then the search path at
+the end needs to be specified explicitly, because the configuration in
+`tox.ini` doesn't take care of it in that case and unintended tests from
+other tox environments would be collected and run too.  
+Here we also disabled the coverage check, which is configured in
+`tox.ini`, because running subsets of the tests can't meet the desired
+coverage goal anyway and will just produce a useless warning. In this case
+some tests can only run as root. The same applies when we run the tests
+from just one file. But note that you also don't get any coverage reports
+in that case.
+
+You can also use environment variables to pass options to pytest, then you
+don't have to worry about the search path:
+
+    export PYTEST_ADDOPTS="--no-cov -vv"
     tox -e storage
 
-Increasing test verbosity:
+To run the tests from just a single file:
 
-    tox -e storage -- -vv
-
-You can also use environment variables:
-
-    export PYTEST_ADDOPTS=-vv
-    tox -e storage
-
-Running specific storage tests modules:
-
-    tox -e storage -- storage/image_test.py
+    tox -e storage -- --no-cov storage/image_test.py
 
 Alternatively, all tests but storage tests can be run by doing:
 


aesteve-rh (changes requested):
Thanks! We are getting a nice development document for new contributors :)

Looks good, see some minor comments inline.

doc/development.md
```diff
@@ -35,24 +35,46 @@
...
-To configure sources (run `./configure --help` to see configuration options):
+To configure sources (create Makefiles and configure script): 
 
     git clean -xfd
     ./autogen.sh --system --enable-timestamp
-    make
 
-To test Vdsm (refer to tests/README for further tests information):
+(It is recommended to start fresh after an update, from the previous step, before
+building RPMs, because the generated Makefile also contains the release number.)
```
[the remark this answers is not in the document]
nirs: I would keep this simple for the basic build flow. We can discuss rebasing in a section about preparing and updating prs.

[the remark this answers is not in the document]
rokkbert: This document doesn't really talk about how to commit/rebase/push/merge things back to the upstream repo (yet). I assume that the process isn't that VDSM-specific, but rather should be the same for all of oVirt? So I'm not sure into how much detail we should go about that here.

doc/development.md
```diff
@@ -7,7 +7,7 @@ Fork the project on https://github.com/oVirt/vdsm.
 Clone your fork:
```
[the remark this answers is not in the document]
rokkbert: Which commit are You referring to?

rokkbert:
All the commits in this PR should be squashed when merging, I think. They are not interesting enough.

doc/development.md
```diff
@@ -35,24 +35,46 @@ Create virtual environment for vdsm:
...
     git clean -xfd
     ./autogen.sh --system --enable-timestamp
-    make
 
-To test Vdsm (refer to tests/README for further tests information):
+(It is recommended to start fresh after an update, from the previous step, before
+building RPMs, because the generated Makefile also contains the release number.)
 
-    make check
+Now you can create the virtual environment (https://docs.python.org/3/library/venv.html),
```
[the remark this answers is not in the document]
nirs: We don't need to guess, you can preview this page here:
https://github.com/oVirt/vdsm/blob/bc93d3bfa262f8ac047d09e8f66054526adb16fe/doc/development.md

If you don't want to provide special text for the url, there is point in using url syntax.

doc/development.md
```diff
@@ -35,24 +35,46 @@ Create virtual environment for vdsm:
...
-To configure sources (run `./configure --help` to see configuration options):
+To configure sources (create Makefiles and configure script): 
 
     git clean -xfd
     ./autogen.sh --system --enable-timestamp
-    make
 
-To test Vdsm (refer to tests/README for further tests information):
+(It is recommended to start fresh after an update, from the previous step, before
+building RPMs, because the generated Makefile also contains the release number.)
```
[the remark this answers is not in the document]
nirs: It is also not clear what is "start fresh" and "the previous step", and it will break easily when someone will add a new step before this step.

It think it is enough to explain the needed steps, like:

    Configure vdsm:
  
        ./autogen.sh ...

And add a note right after why this is needed:

   NOTE: if you don't configure before building rpms you may
   get packages with wrong release string, and upgrading will
   fail.

doc/development.md
```diff
@@ -35,24 +35,46 @@ Create virtual environment for vdsm:
...
 
-    make check
+Now you can create the virtual environment (https://docs.python.org/3/library/venv.html),
+which is necessary to run the tests later. This needs to be done only once:
+
+    make venv
+
+To see configuration options run `./configure -h`. Configure things, if necessary,
+then start make to build Vdsm:
```
nirs: Changing the configuration is advanced topic that always never needed. Lets keep things
simpler here, and keep this info in another section.

doc/development.md
```diff
@@ -35,24 +35,46 @@ Create virtual environment for vdsm:
...
+
+Note how the prompt gets a `(vdsm)` prefix.
+After that you can:
+
+    make check
+
+To leave the virtual environment:
+
+    deactivate
+
```
nirs, line 73: Testing vdsm needs a new section. This section is about building vdsm.

tests/README.md
```diff
@@ -13,11 +13,11 @@ Note: Not all tests have been converted to tox and/or pytest yet.
 
 ## Running the tests
 
-Before running vdsm tests, activate the environment:
+Before running Vdsm tests, activate the environment (which you need to create once, with `make venv`, as described in [doc/development.md](doc/development.md)):
```
nirs, line 16: Please keep lines wrapped. The text should look nice in your editor since
developer will check this file locally.

doc/development.md
```diff
@@ -35,24 +35,46 @@ Create virtual environment for vdsm:
...
     make rpm
 
 To upgrade your system with local build's RPM:
 
     make upgrade
 
+To test Vdsm you need to activate the virtual environment first.
+(refer to [tests/README.md](tests/README.md) for further testing information):
+
+    . ~/.venv/vdsm/bin/activate
```
nirs, line 63: Use source, "." works but cryptic.

doc/development.md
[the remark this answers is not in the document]
mz-pdm: We needn't be that specific, this is not a git(hub) tutorial.  I'd just either drop the remark about release numbers or add a remark about making sure to be on the top of the upstream branch.  The point is that if you are behind the upstream branch, you may be on an old tag, resulting in building rpm's with an old version.

nirs:
I looked only at the first commit, will review everything later.

doc/development.md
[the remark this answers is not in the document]
aesteve-rh: Every commit should have some text (at least a why) and a signoff (for now). In this case I was referring to the first commit, the other commits are fixes over the first one, so they should probably be squashed.

Personally, I would've amended in this case, so I don't have to squash later. In general, there is a lot of `rebase` involved in the review process :)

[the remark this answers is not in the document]
rokkbert: Ok, it did have a short message and it was simple :)
I've been told not to amend (or force-push ;-) ) in other projects, but I agree it would make sense in this case. Will keep it in mind for next time.

tests/README.md
rokkbert: Ok. What is our recommended line length?

doc/development.md
rokkbert: So we should not mention the venv here, but only in the tests/README.md?

tests/README.md
nirs: In general 79 chars, but for long text using shorter lines may be better.
for example in commit message I use 72 chars.

doc/development.md
rokkbert: You mean we need not mention that there are options here? Or we don't talk about configure at all since it's optional?

doc/development.md
nirs: We need to add a new section in this document about running the tests. Entering the virtual env is not related to building vdsm.

doc/development.md
nirs: There is no need to talk about configure in this section. Maybe in another "advanced" section at the end of the document.

doc/development.md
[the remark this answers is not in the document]
rokkbert: Interesting, that you can reach the commit-content like this :) Also here, in its branch: https://github.com/rokkbert/vdsm/blob/dev-docs/doc/development.md

doc/development.md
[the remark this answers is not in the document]
rokkbert: I though the point was that even if you are up-to-date the Makefile would not necessarily be regenerated and could keep the old number. But now I see that it _does_ seem to be created new each time I run autogen.
So I don't really understand the point, because that you get old packages when you build an old version of the code seems kind of obvious. What, then, _is_ the purpose of `git clean -xfd`?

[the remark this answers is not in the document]
mz-pdm: > I though the point was that even if you are up-to-date the Makefile would not necessarily be regenerated and could keep the old number. But now I see that it _does_ seem to be created new each time I run autogen.
> So I don't really understand the point, because that you get old packages when you build an old version of the code seems kind of obvious. 

OK, then I would drop the change of this piece of the document.

> What, then, _is_ the purpose of `git clean -xfd`?

There might be some argument beyond "just cleanup to be sure", but GitHub doesn't make easy to find PR comments, so maybe Albert or Nir can clarify it.

rokkbert:
> > All the commits in this PR should be squashed when merging, I think. They are not interesting enough. Maybe I rebase-squash everything, once all comments are resolved?
> 
> If the commit should be squashed please squash and push again to update the PM. We don't squash prs in vdsm during merge.

Ok.

> You should split the commits so each commit is doing one logical change and is easy to review and revert if needed. Every commit should have good commit message explaining why the change is needed.

Do You think that it's worth keeping commits (and their messages) for documentation-changes like this?

nirs:
> > You should split the commits so each commit is doing one logical change and is easy to review and revert if needed. Every commit should have good commit message explaining why the change is needed.
> 
> Do You think that it's worth keeping commits (and their messages) for documentation-changes like this?

It depends on the change, and usually less important in documentation changes.

mz-pdm:
> Do You think that it's worth keeping commits (and their messages) for documentation-changes like this?

To a reasonable extent yes.  Besides what Nir wrote, commit messages are sometimes valuable sources of information why a change was done, which may make sense also for documentation changes.

doc/development.md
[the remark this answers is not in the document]
nirs: In vdsm, engine, ovirt-imageio, and ioprocess we always send good set of changes. We don't add a new commit to fix an issue in a previous commit. We fix issue by "git rebase -i" and editing the commit that need to change.

doc/development.md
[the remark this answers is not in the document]
nirs: Running "git clean -dxf" is known to work for creating good build. One known issue
this solves is that the generated Makefile contains the RELEASE string
("{timestamp}.git{commit-hash}"). If you create a build with the same Makefile,
you always get the same release string with the wrong release and upgrading the 
vdsm does not work.

doc/development.md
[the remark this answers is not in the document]
nirs: I used the "View file" menu in the "---" menu on the top right of the file in the commit view.

doc/development.md
nirs: It may be POSIX and standard, but it is still cryptic and should not be used unless source is not available (this seems to be the case in FreeBSD).

Regardless, we should really use same instructions as in python venv documenation:
https://docs.python.org/3/library/venv.html#creating-virtual-environments

doc/development.md
nirs: We need all basic information for developers in one place. tests/README can handle stuff which is rarely needed, or converted to other documents in doc/.

doc/development.md
```diff
@@ -35,21 +35,21 @@ Create virtual environment for vdsm:
...
 Now you can create the virtual environment (https://docs.python.org/3/library/venv.html),
 which is necessary to run the tests later. This needs to be done only once:
 
     make venv
 
 To see configuration options run `./configure -h`. Configure things, if necessary,
-then start make to build Vdsm:
+then build Vdsm:
 
     make
```
nirs, line 51: The changes in this commit should be squashed into the first commit if we need them.

doc/development.md
```diff
@@ -28,10 +28,6 @@ Install additional packages for Fedora, CentOS, and RHEL:
 
     sudo dnf install -y `cat automation/check-patch.packages`
 
-Create virtual environment for vdsm:
-
-    make venv
-
```
nirs, line 34: Why remove this? this belongs to the initial setup that someone must do once, unlike
activating the virtual env that must be done before running the tests.

nirs (changes requested):

doc/development.md
rokkbert: I got the impression that you guys didn't like to mix that in here, because it needs to be done only for running the tests, and not building. Or does it matter, at what time it is executed? I can also leave it in, of course, but I don't think the tests-doc would be complete without mentioning it again, redundantly.

doc/development.md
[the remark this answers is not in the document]
nirs: I tested now with patch on current master, and running autogen.sh was enough to regenerate the makefile.

Before running autogen.sh:

```
$ grep 'RELEASE =' Makefile.before 
PACKAGE_BASE_RELEASE = 1
PACKAGE_RELEASE = 1
```

After running autogen.sh:

```
$ grep 'RELEASE =' Makefile
PACKAGE_BASE_RELEASE = 202206201318.gite763a6b19
PACKAGE_RELEASE = 202206201318.gite763a6b19
```

So we can remove the git clean from the instructions.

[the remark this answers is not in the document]
rokkbert: I'm not sure what the deciding factor is, but in my tests the Makefile was not touched by running autogen.sh again.

doc/development.md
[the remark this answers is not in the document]
aesteve-rh: I still don't see the commit message nor the signoff. It is in the PR but is missing from the commit.

doc/development.md
```diff
@@ -28,24 +28,23 @@ Install additional packages for Fedora, CentOS, and RHEL:
...
 ## Building Vdsm
 
-To configure sources (run `./configure --help` to see configuration options):
+To prepare sources (create Makefiles and configure script):
 
-    git clean -xfd
     ./autogen.sh --system --enable-timestamp
-    make
 
-To test Vdsm (refer to tests/README for further tests information):
```
aesteve-rh, line 44: With this, we lose the reference to the tests/README, which is not mentioned either in the main README.
And we mix creating a virtual environment, which is meant for testing mainly, with the `Building Vdsm` section.

Following Nir's suggestion here: https://github.com/oVirt/vdsm/pull/255#discussion_r898273342
I think it would be nice to have a `Testing Vdsm` section (or similar name) with very basic, everyday commands (e.g., make venv once, then activate, a simple test run with tox, and deactivate, that's all), and refer to the tests/README for further information.

tests/README.md
```diff
@@ -13,11 +13,12 @@ Note: Not all tests have been converted to tox and/or pytest yet.
...
 ## Running the tests
 
-Before running vdsm tests, activate the environment:
+Before running Vdsm tests, activate the virtual environment (which you need to
+create once, with `make venv`, as described in [doc/development.md](doc/development.md)):
 
     source ~/.venv/vdsm/bin/activate
 
-When done, you can deactivate the environment:
+When done with testing, you can deactivate the environment:
```
aesteve-rh, line 21: WIth the test README I thought me discussed adding an example for pytest flags `--log-cli-level=DEBUG`, maybe also  `--cov-fail-under 0`.

But we can leave it for another PR, more specific for tests, so that we can move this forward.

doc/development.md
rokkbert: I agree. But it probably shouldn't be the storage tests then, for the simple example, because those are more complicated.

rokkbert: There's also a lot of testing-related make-targets, and it's not obvious, which ones should actually be used when, I think.

aesteve-rh: AFAIK the testing make-targets are meant mostly for the pipelines, developers will rely more on tox (or pytest) directly, since we run only what tests our change set.
In any case, I would not put them here.

tests/README.md
rokkbert: I added --no-cov as an example, I think that is simpler than setting the limit to 0.

doc/development.md
rokkbert: I added a basic test example in a seperate section, including the link to the tests/README

doc/development.md
rokkbert: I squashed them all.

doc/development.md
rokkbert: Ok, I didn't add anything about the make targets, I just left in the ones that were already described.

doc/development.md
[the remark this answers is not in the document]
rokkbert: Is it ok now, @aesteve-rh ? Otherwise I don't understand what I'm missing.

aesteve-rh (dismissed):
Nice!

tests/README.md
```diff
@@ -13,11 +13,12 @@ Note: Not all tests have been converted to tox and/or pytest yet.
...
 ## Running the tests
 
-Before running vdsm tests, activate the environment:
+Before running Vdsm tests, activate the virtual environment (which you need to
+create once, with `make venv`, as described in [../doc/development.md](../doc/development.md)):
 
     source ~/.venv/vdsm/bin/activate
 
-When done, you can deactivate the environment:
+When done with testing, you can deactivate the environment:
```
nirs, line 21: This info was added now to doc/developmenet.md - do we need to repeat it here?

tests/README.md
```diff
@@ -47,6 +48,12 @@ Running specific storage tests modules:
 
     tox -e storage -- storage/image_test.py
 
+If you run only some tests then the configured coverage goals (see `tox.ini`)
+can not be met, so it's better to disable the coverage check and avoid the
+useless warning. Any arguments after `--` are passed through to pytest:
```
nirs, line 53: Explaining why we need to use "--" should be done in the first example using it, maybe as an example.

We should really add --no-cov to all example running tests from single module, not as separate section. We can explain why on the first example using this.

nirs (changes requested):

doc/development.md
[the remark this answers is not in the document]
rokkbert: That's true. I'll try to include that as well.

doc/development.md
```diff
@@ -28,24 +28,23 @@ Install additional packages for Fedora, CentOS, and RHEL:
...
-    make venv
-
 
 ## Building Vdsm
 
-To configure sources (run `./configure --help` to see configuration options):
+To prepare sources (create Makefiles and configure script):
 
-    git clean -xfd
     ./autogen.sh --system --enable-timestamp
```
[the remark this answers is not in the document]
rokkbert: Do You mean the "git clean" call? (And if so: didn't we just verify that it is unnecessary to do that?) Otherwise I don't understand the comment, because I didn't change the autogen command.

doc/development.md
```diff
@@ -28,24 +28,23 @@ Install additional packages for Fedora, CentOS, and RHEL:
...
-    make
 
-To test Vdsm (refer to tests/README for further tests information):
+Now you can create the virtual environment (https://docs.python.org/3/library/venv.html),
+which is necessary to run the tests later. This needs to be done only once:
+
+    make venv
```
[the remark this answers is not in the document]
rokkbert: I don't understand what follows from that? Is my wording incorrect or missing something? Or is this the wrong place to put it?

doc/development.md
```diff
@@ -54,6 +53,23 @@ To upgrade your system with local build's RPM:
...
+
+Then start some tests with tox, for example the networking tests:
+
+    tox -e network
+
+To exit the virtual environment afterwards:
+
+    deactivate
+
+For more information about testing see [../tests/README.md](../tests/README.md).
```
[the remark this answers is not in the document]
rokkbert: Actually I think I tried that and it didn't work. Is that supposed to be some github-magic, that turns that into an absolute link? I'll try it again.

tests/README.md
rokkbert: I think it's ok to duplicate it as a reminder, because the setup-info in the other document might have been read at a much earlier time, and now the user looks for testing details in this document and might have forgotten about it.

doc/development.md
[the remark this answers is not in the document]
rokkbert: It needs to start with the / in order to work. But that is better than relative paths.

doc/development.md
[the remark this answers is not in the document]
nirs: No, the `.autogen.sh ...` line.

doc/development.md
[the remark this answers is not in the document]
nirs: I just copied the URL to the page, no magic.

    [tests/README.md](https://github.com/oVirt/vdsm/blob/master/tests/README.md)

doc/development.md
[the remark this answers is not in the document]
nirs: I mean that we don't need this here, it should be in the "Enviroment setup" section where it was before.

tests/README.md
nirs: OK, lets keep this duplicated for now.

doc/development.md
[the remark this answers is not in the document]
rokkbert: The reason it was moved away from there was that it isn't possible to do it before the Makefile is created. It's confusing for the reader who follows instructions from top to bottom if a command just fails because the instructions give the commands in an order in which they can't be executed.

doc/development.md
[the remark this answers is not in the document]
rokkbert: I don't think that's a good idea, because then the link doesn't work inside a fork of the repo. It jumps back to the original. Do we want that?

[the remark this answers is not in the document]
nirs: Well if you find a way to do a link that works both in a fork and in the main repo it would be better.

doc/development.md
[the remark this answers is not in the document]
nirs: This is why we need to add .autogen.sh to "Enviroment setup":

- add custom repos
- install development tools
- clone the git repo
- autogen.sh
- create the env

What's wrong with this flow?

Then in the section about building vdsm we can have:

- autogen.sh (should not be needed, but currently required)
- build the rpms
- install the rpms

And in running the tests:

- enter the env
- run the tests

tests/README.md
nirs: --no-cov does not collect coverage, which is bad. We always want to collect coverage so it is easy to inspect the coverage.

What we want to disable is failing the tests if coverage is lower than the limit, since this work only when running all the tests as root.

So I think this should be fixed by using:

   --cov-fail-under=0

as default value, so developers do not need to specify it, and enable the actual limit in the CI, when we run all the tests.

This can be done by using default in tox.ini:
https://github.com/oVirt/vdsm/blob/9bf9a30230ae318a7798a25b5ab362ce33a006a7/tox.ini#L97

We change the default to 0, and inject the value using
https://github.com/oVirt/vdsm/blob/master/automation/tests.sh

This is not in the scope of this PR, but can be the next step if you have time for this.

tests/README.md
```diff
@@ -32,20 +43,29 @@ for more information.
...
-
-Increasing test verbosity:
+    tox -e storage -- -vv --no-cov storage
 
-    tox -e storage -- -vv
+Note that for some test modules (like `storage` or `virt`), _if_ there
+are any extra pytest arguments (after the `--`) then the search path at
+the end needs to be specified explicitly, because the configuration in
+`tox.ini` doesn't take care of it in that case and too many tests would be
+collected.
```
mz-pdm, line 52: It should be clarified what "search path" is, best accompanied with an example.  Also, "too many tests would be collected" sounds confusing -- can we run "too many" tests?

And there is an extra space after the last sentence.

mz-pdm (commented):
Looks good to me, except for the paragraph commented inside.

doc/development.md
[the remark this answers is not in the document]
rokkbert: I think `[/tests/README.md](/tests/README.md)` is fine, it works everywhere and it looks correct. Without the leading / it feels wrong, because we are in a subfolder that we need to get out of.

doc/development.md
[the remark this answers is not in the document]
rokkbert: Nothing wrong with that except maybe for the duplication of the autogen. But I think we can't avoid all the little issues, so I do it like that now.

tests/README.md
rokkbert: Makes sense. I agree it's getting out of scope though.

tests/README.md
rokkbert: I'm afraid I don't know how to clarify that well because I'm not experienced enough with tox/pytest.

"too many tests" means "more tests than the user actually wanted to specify". If I think I was asking for the storage tests only, but then other tests run as well, because it doesn't work like that, then that is "too many tests". How could we say it better? Maybe "unintended tests"?

The two spaces after "collected." are for the line break.

mz-pdm: > "too many tests" means "more tests than the user actually wanted to specify". If I think I was asking for the storage tests only, but then other tests run as well, because it doesn't work like that, then that is "too many tests". How could we say it better? Maybe "unintended tests"?

"unintended test" would be definitely better; or maybe something like "tests out of the specified collection would be run" could be used (+ the example how to specify the path to work around the problem).

> The two spaces after "collected." are for the line break.

Ah, I see.

nirs: "too many tests" can be replaced by "tests from other tox env".

aesteve-rh (approved):

tests/README.md
rokkbert: > (+ the example how to specify the path to work around the problem)

The example was:
`tox -e storage -- -vv --no-cov storage`
Do You think that is sufficient?

mz-pdm: Yes, I understand now what's meant there, thanks.

mz-pdm (approved):

doc/development.md
```diff
@@ -28,32 +21,63 @@ Install additional packages for Fedora, CentOS, and RHEL:
...
+(https://docs.python.org/3/library/venv.html), which is necessary to run the
+tests later. This needs to be done only once:
 
     make venv
 
 
 ## Building Vdsm
 
-To configure sources (run `./configure --help` to see configuration options):
+Before building, it is recommended to recreate the Makefile because it
```
tinez, line 44: Isn't this an issue only when building RPMs? If so maybe we can mention it in the RPM section.

doc/development.md
```diff
@@ -28,32 +21,63 @@ Install additional packages for Fedora, CentOS, and RHEL:
 
     sudo dnf install -y `cat automation/check-patch.packages`
 
-Create virtual environment for vdsm:
+Fork the project on https://github.com/oVirt/vdsm.
```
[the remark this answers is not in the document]
michalskrivanek: I consider automation/*  dead. Does that list in check-patch.packages still make any sense?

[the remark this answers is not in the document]
nirs: We still use scripts from automation in github ci, and the package list is still correct.
We have #43 to remove this directory.

But of course installing the packages can be done only after cloning the source so the clone should move before that command.

doc/development.md
nirs: This section is about building rpms, what else do we build?

doc/development.md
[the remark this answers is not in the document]
rokkbert: I tried to follow the suggestion from https://github.com/oVirt/vdsm/pull/255#discussion_r908783004 but I see the problem now.

doc/development.md
tinez: > what else do we build?

Build-time generated py modules needed for tests etc. The RPM part is mentioned only in line 54 of this PR. For `make; tox -e storage` cycles we don't need to run autogen each time.

nirs: But this section is about building and installing rpms. We have a section about running
the tests. Maybe the heading should be renamed to make it more clear.

Normally make is not needed for running the tests, it is used once when preparing
the source and you need to invoke make only if you change the makefiles or the templates
generated sources. I guess the "running the tests" section need to explain about this.

nirs (approved):
Good enough, time to merge.

tinez (approved):
LGTM

PR #256, opened 2022-06-15
hsm: connectStorageServer receive device list

aesteve-rh:
When connecting a device to a host, add an option for the
caller to specify the list of devices to wait for in
StoragePool.connectStorageServer.

By default, Vdsm uses udevadm settle and 5 seconds timeout
to make sure devices are settled before continuing after
setting up an iscsi device, but that is still not robust
enough and can break due to a race condition.

By telling Vdsm which devices must exist on the host
when connecting to a server for making a direct LUN
available, we can make the API more robust.

Bug-Url: https://bugzilla.redhat.com/2074029


aesteve-rh:
Right now this is too big of a change, so let's drop it.

PR #262, opened 2022-06-21
[WIP] lvm: refactor proposal for LVM command formatting

aesteve-rh:
LVMCache holds a runner instance and a great part of the LVM command formatting responsibility. This is like that mostly for convenience, as the cache runs commands internally to keep itself updated when we operate with it, and uses cached devices to construct the command. But also adds a handy additional layer, with a semaphore to control the maximum concurrent commands and a potential re-try with refreshed devices.

However,  LVMCache code is complex enough and does not need the additional complexity of being in charge of formatting the commands. Leads to unconnected, inconsistent code that clearly does not belong together (adding a bunch of methods that are basically purely static as they do not interact with any member of the class).

This patch moves the LVMRunner (and then removes it) and all the formatting functions to a new module called `lvmcmd`. This change is transparent for functions that were using the `run_command` from the LVMCache to get cached devices. The method is still available, but it invokes `lvmcmd` for running the command, instead of using the LVMRunner instance. Ultimately, `lvm` module will keep mainly the LVMCache and all the functions to interact with it, and `lvmcmd` will be in charge of running and formatting all commands.

### DIFF
--- a/lib/vdsm/storage/lvm.py
+++ b/lib/vdsm/storage/lvm.py
@@ -1096,7 +1096,7 @@ def changelv(vg, lvs, attrs):
     """
     cmd = lvmcmd.lvchange(vg, normalize_args(lvs), attrs)
     try:
-        _lvminfo.run_command(tuple(cmd), devices=_lvminfo._getVGDevs((vg, )))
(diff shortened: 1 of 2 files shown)


aesteve-rh:
I'll close this. If I get the time to get into this again, I will document it first. In the meantime, I'll keep it local or in my fork.

PR #265, opened 2022-06-22
Fix migration of VM with dedicated CPU

nyoxi:


ahadas (dismissed):

mz-pdm:
Looks OK, but please improve commit messages:

- "raise exception properly" should be more specific, e.g. "raise, not return, exception in cpumanagement._siblings"

- We replace CPU pinning in _srcDomXML and libvirt.VIR_MIGRATE_PARAM_DEST_XML.  It's worth to mention that both the XMLs are different and passed to different functions and that means we have to replace CPU pinnings in them separately.

nyoxi:
> Looks OK, but please improve commit messages:

Done, please check the commit message if the change is adequate.

ahadas (approved):
I tend to miss the commit messages on github.. so first, good that others don't :) and second, the changes looked good also before but the commit message made it more clear

mz-pdm (approved):

PR #269, opened 2022-06-23
hook-log-console: Don't require exact vdsm match

tinez:
There's no reason to require exactly the same vdsm version
for the hook to work. It causes issues when installing on top of
node/rhvh, so let's use a more relaxed requirement.

### DIFF
--- a/vdsm.spec.in
+++ b/vdsm.spec.in
@@ -502,7 +502,7 @@ The VM must be pinned to the host.
 %package hook-log-console
 Summary:        Log VM's serial console to a file
 BuildArch:      noarch
-Requires:       %{name} == %{version}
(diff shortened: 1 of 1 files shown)


didib:
I just copied this from hook-localdisk without thinking too much... incidentally, hook-localdisk is the only other one that requires an exact match (not sure why, irrelevant for now). Sorry.

didib (approved):

mz-pdm (approved):

tinez:
> I just copied this from hook-localdisk without thinking too much... incidentally, hook-localdisk is the only other one that requires an exact match (not sure why, irrelevant for now). Sorry.

My bad - should've spotted that in the first review.

PR #271, opened 2022-06-29
hooks: Add log-firmware

didib:
This is a combination of the interface of log-console and the
functionality of qemucmdline, hard-coded to log the firmware output to a
log file. qemucmdline can't be used directly for HostedEngine, because
we want a separate log file per VM, and qemucmdline does not have
internally functionality to allow doing that, relying instead on the
engine to supply correct params somehow. For HostedEngine, we can't rely
on the engine, as it does not exist at this point.

Change-Id: I5e2324009daed5136ab7884226cce0feee2e0c96
Signed-off-by: Yedidyah Bar David <didi@redhat.com>

### DIFF
--- a/vdsm_hooks/log_firmware/README
+++ b/vdsm_hooks/log_firmware/README
@@ -1,0 +1,51 @@
+log_firmware vdsm hook
+=================================
+This hook appends a VM's firmware debug logging to a log file on the host.
+
+Installation:
+* Use engine-config to set the appropriate custom properties as such:
+
+Custom property for a VM:
+    sudo engine-config -s "UserDefinedVMProperties=log_firmware=^(on|off)$"
+
+* Prepare the log directory on the host:
+    sudo mkdir -p /var/log/qemu-firmware
+    sudo chown qemu /var/log/qemu-firmware
+    sudo chcon -t qemu_var_run_t /var/log/qemu-firmware
+    sudo chmod 700 /var/log/qemu-firmware
+
+Usage:
+
+The hook can be enabled using one of two ways:
+
+1. Set the VM property 'log_firmware' to 'on', for specific VMs, as needed, via
+the engine. Set it to 'off', to explicitly disable the hook.
+
+2. Set the environment variable 'log_firmware_vm_regexp', in
+/etc/sysconfig/vdsm, to a regular expression matching the VM names for which
+you want to enable the hook. Matching is done using Python's re.match().
+
+If log_firmware is set to 'off', for some VM, the hook will be disabled,
+regardless of the value of log_firmware_vm_regexp or whether the name matches.
+
+If 'log_firmware' is not 'on', for some VM, and 'log_firmware_vm_regexp' does
+not match, the hook will be disabled. So by default, with both variables
+being empty/unset, the hook is disabled.
+
+In principle, a future version might enable it by default, based on feedback
+from users.
+
+Using log_firmware_vm_regexp is useful, in particular, for enabling the hook
+for a HostedEngine VM, for which using 'log_firmware' is less practical, as
+the engine might be down exactly when you want to use it (to help diagnose
+why it's down...).
+
+The logs are written to the directory set with the environment variable
+log_firmware_dir, or to /var/log/qemu-firmware if it's not set.
+If you do set it, you should prepare the directory as detailed above.
+The logs' filenames, inside this directory, are {vm_name}-firmware.log,
+where {vm_name} is replaced with the VM's name.
+
+TODO: I'd like to be able to be able to write log files to
+/var/log/libvirt/qemu, like other logs, but that's not easy.
+See also: https://gerrit.ovirt.org/c/lago-ost/+/114150
--- a/vdsm_hooks/log_firmware/before_vm_start.py
+++ b/vdsm_hooks/log_firmware/before_vm_start.py
@@ -1,0 +1,90 @@
+#!/usr/bin/python3
+#
+# Copyright 2022 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+#
+# Refer to the README and COPYING files for full details of the license
+#
+
+import os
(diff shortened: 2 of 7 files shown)


vdsm_hooks/log_firmware/README
```diff
@@ -0,0 +1,42 @@
+log_firmware vdsm hook
+=================================
+This hook appends a VM's firmware to a log file on the host.
```
mz-pdm, line 3: ... VM's firmware *log output* to a log file ...

mz-pdm (commented):
OK, but could it perhaps make sense to merge log_console and log_firmware hooks into a single one, to avoid having more and more hooks?

didib:
> OK, but could it perhaps make sense to merge log_console and log_firmware hooks into a single one, to avoid having more and more hooks?

I think some cases would want one and not the other, but do not feel strongly about this. Do you have a specific interface in mind? Have 4 different env vars like current patch? Only the existing two of log-console, but extend their functionality? Something else?

didib:
We still do not have many users of log-console (only OST, AFAIK), so we can rename it as well, if you want.

didib:
> > I think some cases would want one and not the other,
> 
> Yes, what I meant was that the options would be still independent, just available in a single hook because they serve a similar purpose (logging some extra information from VMs). Something like that cat, mv, cp are similar utilities in a single package. But if you think it would be more confusing than useful, I'm OK to have them separate, I just don't like the number of vdsm-hook-* packages :-).

I agree, just do not feel strongly either way...

As I said, we can still rename log-console, if you want. I'd personally not find it very nice to keep it called 'log-console' and _also_ make it log firmware output...

didib:
> Sure, a better name would have be invented. Maybe log-vm-outputs?

Slightly too general, IMO. One might expect from such a hook to also e.g. log all outgoing network traffic.

I considered 'log-early-boot' or 'debug-early-boot', but 'log-console' logs all console output, not only during boot... And can be practically useful also for diagnosing lockups etc. during normal use.

didib:
Works for me, but is not ready yes - requires:
```
chcon -t qemu_var_run_t /var/log/firmware
```

didib:
> Works for me, but is not ready yes - requires:
> 
> ```
> chcon -t qemu_var_run_t /var/log/firmware
> ```

Do you think it's enough, considering this is just a hook, to just add something like this to the README?

mz-pdm:
> Do you think it's enough, considering this is just a hook, to just add something like this to the README?

I'd say it's enough, at least until somebody complains :-).

But you mention /var/log/libvirt/qemu/{vm_name}-firmware.log as the log location in the README -- wouldn't it be a generally better location than /var/log/firmware and wouldn't it avoid the problem?

didib:
> But you mention /var/log/libvirt/qemu/{vm_name}-firmware.log as the log location in the README

Indeed, I changed the code but not the README yet.

> -- wouldn't it be a generally better location than /var/log/firmware and wouldn't it avoid the problem?

/var/log/libvirt/qemu has selinux permissions for logging stuff from qemu, not sure exactly how, but apparently not directly. Perhaps only to virtlogd, which qemu logs through indirectly somehow. Anyway, the firmware log seems to be written by qemu itself, and writing there fails. /var/log/firmware was just an attempt to make it work, for my current need ([2101787](https://bugzilla.redhat.com/show_bug.cgi?id=2101787), not a concrete suggestion for review.

Assuming that writing to /var/log/libvirt/qemu as in the README is hard (didn't try yet - will likely require playing with the selinux policy in a more involved way than I want to spend time on, see above about virtlogd), I am going to suggest a different location. If something else under /var/log/libvirt will work for me, I think that would be my choice. Otherwise, I don't mind that much - perhaps /var/log/qemu-firmware or whatever. We just need to come up with something clear and with a low risk for collisions...

tinez:
Given the amount of overlap between this hook and the console one I'd definitely vote for merging these two. `log-early-boot` sounds good IMO. Also, I'd stick with logging to `/var/log/libvirt/qemu/{vm_name}-firmware.log` for consistency. If the hooks gains users other than OST we can think about customizations and dealing with selinux, but for now I don't think it's worth spending our time on this.

vdsm_hooks/log_firmware/README
didib: Changed to "... VM's firmware debug logging to a log ...". I hope this is clear enough...

michalskrivanek:
lgtm as is, we don't have to spend too much time on it until we have a need

didib:
Fixed a bug in the README: s;/var/lib/libvirt/qemu-firmware;/var/log/qemu-firmware;

didib:
Rebased

vdsm_hooks/log_firmware/before_vm_start.py
```diff
@@ -0,0 +1,90 @@
...
+
+    domxml = hooking.read_domxml()
+    vm_name = domxml.getElementsByTagName('name')[0].firstChild.data
+    hooking.log(f'vm_name: {vm_name}')
+
+    if (
+        log_firmware != 'on'
+        and (
+            not log_firmware_vm_regexp
+            or not re.match(log_firmware_vm_regexp, vm_name)
```
mz-pdm, line 46: is / is not None would be idiomatic but no big deal.

mz-pdm (approved):

vdsm_hooks/log_firmware/before_vm_start.py
didib: I admit I never heard that this is considered idiomatic.

See e.g. the start of [Match Objects](https://docs.python.org/3/library/re.html#match-objects):
```
Match objects always have a boolean value of True. Since [match()](https://docs.python.org/3/library/re.html#re.Pattern.match) and [search()](https://docs.python.org/3/library/re.html#re.Pattern.search) return None when there is no match, you can test whether there was a match with a simple if statement:

match = re.search(pattern, string)
if match:
    process(match)
```

Searching, rather naively, the current master branch of vdsm, I see a few such cases, but not many:
```
[ybardavi@ybardavi vdsm]$ grep -rw 'if .*\.match\>' * | wc -l
16
[ybardavi@ybardavi vdsm]$ grep -rw 'if .*\.match\>.*None' * | wc -l
5
```

didib: IIUC you refer to the case where a user is setting log_firmware_vm_regexp to an empty string.

It's indeed not documented in the README, but I actually think the code does the sane thing and ignores it, instead of matching all possible VM names. I did consider this case when writing and deciding on `not log_firmware_vm_regexp` instead of `log_firmware_vm_regexp is None`.

I obviously agree with you that as a programmer, I should always be aware of the difference and consciously pick one or the other. Not always this is clear from the context/comments, sorry...

mz-pdm: > IIUC you refer to the case where a user is setting log_firmware_vm_regexp to an empty string.

Yes.

> It's indeed not documented in the README, but I actually think the code does the sane thing and ignores it, instead of matching all possible VM names. I did consider this case when writing and deciding on `not log_firmware_vm_regexp` instead of `log_firmware_vm_regexp is None`.

OK.

I already approved this PR, waiting whether @tinez would like to add anything.

PR #273, opened 2022-06-30
docker: add vdsm user to container image

aesteve-rh:
Create an unprivileged vdsm user in the container image to enable running tests as non-root.

This could be achieved by using other unprivileged user, e.g., `nobody`, but creating vdsm allows creating tests specific for vdsm user that run both in the pipelines and in the test VMs.

### DIFF
--- a/docker/Dockerfile.centos-8
+++ b/docker/Dockerfile.centos-8
@@ -60,6 +60,9 @@ RUN dnf copr enable -y ovirt/ovirt-master-snapshot centos-stream-8 \
         xfsprogs \
     && dnf clean all
 
+# Create vdsm user with groups kvm,qemu,sanlock
+RUN useradd --system -m -u 36 -N -G 36,107,179 vdsm
+
 # Add gdb python support.
 RUN debuginfo-install -y python3 \
     && dnf clean all
--- a/docker/Dockerfile.centos-8
+++ b/docker/Dockerfile.centos-8
@@ -1,8 +1,8 @@
 FROM quay.io/centos/centos:stream8
 
 # Add runtime dependencies.
-RUN dnf install -y \
-        http://resources.ovirt.org/pub/yum-repo/ovirt-release-master.rpm \
+RUN dnf copr enable -y ovirt/ovirt-master-snapshot centos-stream-8 \
+    && dnf install -y ovirt-release-master \
     && dnf update -y \
     && dnf install -y \
         autoconf \
@@ -49,7 +49,7 @@ RUN dnf install -y \
         python3-sanlock \
         python3-six \
         python3-yaml \
-        qemu-img-6.0.0 \
+        qemu-img \
         redhat-rpm-config \
         rpm-build \
         sanlock \
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -30,5 +30,5 @@ $(targets):
 
 push:
 	for name in $(targets); do \
-		podman push $(prefix)-$$name ovirtorg/$(prefix)-$$name; \
+		podman push $(prefix)-$$name quay.io/ovirt/$(prefix)-$$name; \
 	done


.github/workflows/ci.yml
```diff
@@ -5,11 +5,27 @@ env:
 jobs:
   lint:
     runs-on: ubuntu-latest
-    container: quay.io/ovirt/vdsm-test-centos-8
+    container: quay.io/alesgar/vdsm-test-centos-8
```
nirs, line 8: This is temporary change for testing I guess?

docker/Dockerfile.centos-8
```diff
@@ -2,7 +2,7 @@ FROM quay.io/centos/centos:stream8
 
 # Add runtime dependencies.
 RUN dnf install -y \
-        http://resources.ovirt.org/pub/yum-repo/ovirt-release-master.rpm \
+        http://resources.ovirt.org/pub/yum-repo/ovirt-release44.rpm \
```
nirs, line 5: This does not look right.

docker/Dockerfile.centos-8
```diff
@@ -60,6 +60,9 @@ RUN dnf install -y \
         xfsprogs \
     && dnf clean all
 
+# Create vdsm user with groups kvm,qemu,sanlock
+RUN useradd --system -m -u 36 -N -G 36,107,179 vdsm
```
nirs, line 64: This like the only change needed in this pr, added in all dockerfiles.

docker/Makefile
```diff
@@ -30,5 +30,5 @@ $(targets):
 
 push:
 	for name in $(targets); do \
-		podman push $(prefix)-$$name ovirtorg/$(prefix)-$$name; \
+		podman push $(prefix)-$$name ovirt/$(prefix)-$$name; \
```
nirs, line 33: This push url is missing quay.io/ prefix, and should really be in its own commit.

.github/workflows/ci.yml
```diff
@@ -5,11 +5,27 @@ env:
...
+      # Required to create loop devices.
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run storage tests
+      run: |
+        ./automation/setup-storage.sh
+        chown -R vdsm:kvm ./ /var/tmp/vdsm*
+        su -c "tox -e storage -- --cov-fail-under 60 \
+           -m 'not (integration or slow or stress or root)' storage" -- vdsm
```
nirs, line 28: I think we need:
- tests-storage
- tetsts-storage-root

Both can use the same tests-storage.sh script, with additional argument so the script can run tests as root or as regular user.

.github/workflows/ci.yml
aesteve-rh: Right. All changes in the ci.yml and the setup-storage script will go away. I have another branch for the test split, this branch is only for the image update. Mentioned it in the PR text.

docker/Dockerfile.centos-8
aesteve-rh: release-master was not available, and this was the most recent rpm I saw in the same resources folder. Maybe we need to upload new rpms? Not sure.

docker/Dockerfile.centos-8
aesteve-rh: 👍

.github/workflows/ci.yml
aesteve-rh: Was a test for running with vdsm user, I will remove the change before merging. I have it covered in a different branch and is as you suggest, two jobs for storage testing.

docker/Makefile
aesteve-rh: Right, I’ll separate it.

docker/Dockerfile.centos-8
nirs: The issue is old links in the docker files. The right way to install is explained
in ovirt master snapshot copr repo. You need to update all the docker files to consume
the right repos.
https://copr.fedorainfracloud.org/coprs/ovirt/ovirt-master-snapshot/

aesteve-rh: Ah I see, I should've known.

I'll update them :+1:

nirs (changes requested):

docker/Dockerfile.centos-8
```diff
@@ -49,7 +49,7 @@ RUN dnf install -y \
         python3-sanlock \
         python3-six \
         python3-yaml \
-        qemu-img-6.0.0 \
+        qemu-img-6.1.0 \
```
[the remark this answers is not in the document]
aesteve-rh: Done, and tested the new image in the pipelines :)

nirs (changes requested):

aesteve-rh:
Done, just pushed it. I'll trigger a run with a follow-up branch I have with updates in the CI :)

Issue #128, opened 2022-04-12
Run the tests as unprivileged user


nirs:
Most of vdsm code run as vdsm:kvm, but vdsm tests run as root by default in the CI.
This causes several problems:
- Tests behave differently when run locally on developer machine and in the CI
- Some tests cannot succeed when run as root, and are skipped in the CI. The tests and the code are likey to break because they are tested only locally.
- Error in code that would fail in runtime running as vdsm:kvm are not detected in the tests, running as root:root.
- We need special configuration to run the containers with privileges - this should be done only for tests that must run as root.

Fix:
- Separate the tests that need to run as root from the tests that need to run as vdsm:kvm
- Run the tests as vdsm:kvm or another unprivileged user
- Run tests that needs root in a privileged container

In general only code that run via supervdsm need to be tested as root. Unfortunately the code running as root is mixed with other code in vdsm.

PR #276, opened 2022-07-05
ci: split storage ci jobs for root and non-root testing

aesteve-rh:
Split storage CI jobs between unprivileged `vdsm` user tests (mark non-root) and root tests.

## Comparison

With the numbers below we end up executing 50 more tests that were skipped before and safe more than 2 minutes in execution time.

Numbers are not averaged, only the last run is considered.

### Before

__All storage tests__
collected 2702 items / 101 deselected / 2601 selected
Required test coverage of 68% reached. Total coverage: 68.87%
`= 2472 passed, 123 skipped, 101 deselected, 5 xfailed, 1 xpassed, 1 warning in 371.31s (0:06:11) =`

### After

__Non-root tests__
collected 2700 items / 235 deselected / 2465 selected
Required test coverage of 61% reached. Total coverage: 61.87%
`= 2402 passed, 59 skipped, 235 deselected, 4 xfailed, 1 warning in 164.30s (0:02:44) =`

__Root tests__
collected 2700 items / 2566 deselected / 134 selected
Required test coverage of 47% reached. Total coverage: 47.76%
`= 120 passed, 12 skipped, 2566 deselected, 1 xfailed, 1 xpassed, 1 warning in 211.60s (0:03:31) =`

__Total__
`= 2522 passed, 71 skipped, 5 xfailed, 1 xpassed, in max time 211.60s (0:03:31) =`

Fixes: https://github.com/oVirt/vdsm/issues/128

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run linters
       run: ./automation/lint.sh
+  # TODO: rename to 'test-storage-user' once the github settings allow it.
   test-storage:
     env:
       TRAVIS_CI: 1
@@ -20,8 +21,20 @@ jobs:
       options: --privileged
     steps:
     - uses: actions/checkout@v2
-    - name: Run storage tests
-      run: ./automation/tests-storage.sh
+    - name: Run storage tests as vdsm
+      run: ./automation/tests-storage.sh vdsm
+  test-storage-root:
+    env:
+      TRAVIS_CI: 1
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/ovirt/vdsm-test-centos-8
+      # Required to create loop devices.
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run storage tests as root
+      run: ./automation/tests-storage.sh root
   tests:
     env:
       TRAVIS_CI: 1
--- a/Makefile.am
+++ b/Makefile.am
@@ -149,10 +149,6 @@ venv:
 tests: tox
 	tox -e "tests,lib,network,virt,gluster,hooks"
 
-.PHONY: tests-storage
-tests-storage: tox
-	tox -e "storage"
-
 .PHONY: tests-storage-user
 tests-storage-user: tox
 	tox -e "storage-user"
--- a/Makefile.am
+++ b/Makefile.am
@@ -153,6 +153,14 @@ tests: tox
 tests-storage: tox
 	tox -e "storage"
 
+.PHONY: tests-storage-user
+tests-storage-user: tox
+	tox -e "storage-user"
+
+.PHONY: tests-storage-root
+tests-storage-root: tox
+	tox -e "storage-root"
+
 .PHONY: storage
 storage:
 	python3 tests/storage/userstorage.py setup
--- a/automation/tests-storage.sh
+++ b/automation/tests-storage.sh
@@ -1,5 +1,17 @@
 #!/bin/sh -e
 
+print_help() {
+    echo "Usage: $0 USER"
+    echo ""
+    echo "  Helper script to setup and run storage tests as USER."
+    echo ""
+    echo "Examples:"
+    echo "  Run tests as root user"
+    echo "    $ $0 root"
+    echo "  Run tests as vdsm user"
+    echo "    $ $0 vdsm"
+}
+
 create_loop_devices() {
     local last=$(($1-1))
     local min
@@ -12,28 +24,51 @@ create_loop_devices() {
 }
 
 setup_storage() {
-    python3 tests/storage/userstorage.py setup
+    # Configure lvm to ignore udev events, otherwise some lvm tests hang.
+    mkdir -p /etc/lvm
+    cp docker/lvmlocal.conf /etc/lvm/
+
+    # Make sure we have enough loop device nodes. Using 16 devices since with 8
+    # devices we have random mount failures.
+    create_loop_devices 16
+
+    # Build vdsm.
+    ./autogen.sh --system
+    make
+
+    # Setup user storage during the tests.
+    make storage
 }
 
 teardown_storage() {
-    python3 tests/storage/userstorage.py teardown \
+    make clean-storage \
         || echo "WARNING: Ingoring error while tearing down user storage"
 }
 
-# Configure lvm to ignore udev events, otherwise some lvm tests hang.
-mkdir -p /etc/lvm
-cp docker/lvmlocal.conf /etc/lvm/
+# Process user argument
+user=$1
+if [ -z "$user" ]; then
+    echo "ERROR: user required"
+    print_help
+    exit 1
+fi
 
-# Make sure we have enough loop device nodes. Using 16 devices since with 8
-# devices we have random mount failures.
-create_loop_devices 16
+echo "Running tests as user $user"
 
-# Build vdsm.
-./autogen.sh --system
-make
-
-# Setup user stoage during the tests.
-trap teardown_storage EXIT
 setup_storage
+if [ "$user" != "root" ]; then
+    # Change ownership of current and storage folders
+    # to allow non-privileged access.
+    chown -R $user ./ /var/tmp/vdsm*
+fi
+# Teardown storage before exit.
+trap teardown_storage EXIT
 
-make tests-storage
+if [ "$user" = "root" ]; then
+    # Run only tests marked as root
+    make tests-storage-root
+else
+    # Run tests not marked as root as $user
+    # Use 'su' instead of 'sudo' in order to preserve the environment.
+    su $user -c "make tests-storage-user"
+fi
--- a/tox.ini
+++ b/tox.ini
@@ -97,6 +97,36 @@ commands =
     --cov-fail-under={env:STORAGE_COVERAGE:68} \
     {posargs:storage}
 
+[testenv:storage-user]
+passenv = {[base]passenv}
+setenv =
+    {[base]setenv}
+    COVERAGE_FILE=.coverage-storage-user
+deps = {[base]deps}
+changedir = {[base]changedir}
+commands =
+    python profile {envname} pytest -m {[base]markers}" and not root" \
+    --durations=20 \
+    --cov=vdsm.storage \
+    --cov-report=html:htmlcov-storage-user \
+    --cov-fail-under={env:STORAGE_USER_COVERAGE:61} \
+    {posargs:storage}
+
+[testenv:storage-root]
+passenv = {[base]passenv}
+setenv =
+    {[base]setenv}
+    COVERAGE_FILE=.coverage-storage-root
+deps = {[base]deps}
+changedir = {[base]changedir}
+commands =
+    python profile {envname} pytest -m {[base]markers}" and root" \
+    --durations=20 \
+    --cov=vdsm.storage \
+    --cov-report=html:htmlcov-storage-root \
+    --cov-fail-under={env:STORAGE_ROOT_COVERAGE:47} \
+    {posargs:storage}
+
 [testenv:virt]
 passenv = {[base]passenv}
 setenv =
--- a/tox.ini
+++ b/tox.ini
@@ -81,22 +81,6 @@ commands =
         network/integration \
         network/unit
 
-[testenv:storage]
-passenv = {[base]passenv}
-setenv =
-    {[base]setenv}
-    COVERAGE_FILE=.coverage-storage
-deps = {[base]deps}
-changedir = {[base]changedir}
-# TODO: Remove --ignore
-commands =
-    python profile {envname} pytest -m {[base]markers} \
-    --durations=20 \
-    --cov=vdsm.storage \
-    --cov-report=html:htmlcov-storage \
-    --cov-fail-under={env:STORAGE_COVERAGE:68} \
-    {posargs:storage}
-
 [testenv:storage-user]
 passenv = {[base]passenv}
 setenv =
--- a/tests/storage/blockdev_test.py
+++ b/tests/storage/blockdev_test.py
@@ -59,6 +59,7 @@ def loop_device(tmpdir):
 class TestZero:
 
     @requires_root
+    @pytest.mark.root
     def test_entire_device(self, loop_device):
         # Write some data to the device.
         with directio.open(loop_device.path, "r+") as f:
@@ -78,6 +79,7 @@ class TestZero:
             assert data == ZERO
 
     @requires_root
+    @pytest.mark.root
     def test_size(self, loop_device):
         # Write some data to the device.
         with directio.open(loop_device.path, "r+") as f:
@@ -94,6 +96,7 @@ class TestZero:
             assert data == DATA
 
     @requires_root
+    @pytest.mark.root
     @pytest.mark.parametrize("size", [
         sc.BLOCK_SIZE_4K,
         250 * sc.BLOCK_SIZE_4K
@@ -154,6 +157,7 @@ class TestDiscard:
             assert data == DATA
 
     @requires_root
+    @pytest.mark.root
     def test_supported(self, loop_device):
         # If the loop device backing file is on a file system that does not
         # support discard, discard is not supported.
--- a/tests/storage/blocksd_test.py
+++ b/tests/storage/blocksd_test.py
@@ -1616,6 +1616,7 @@ def test_dump_sd_metadata(
 
 
 @requires_root
+@pytest.mark.root
 @pytest.mark.parametrize("domain_version", [5])
 def test_create_illegal_volume(domain_factory, domain_version, fake_task,
                                fake_sanlock):
@@ -1644,6 +1645,7 @@ def test_create_illegal_volume(domain_factory, domain_version, fake_task,
 
 
 @requires_root
+@pytest.mark.root
 def test_reduce_volume_called(domain_factory, fake_task, fake_sanlock):
     sd_uuid = str(uuid.uuid4())
     dom = domain_factory.create_domain(sd_uuid=sd_uuid, version=5)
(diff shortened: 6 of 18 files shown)


tests/storage/nbd_test.py
```diff
@@ -132,6 +132,7 @@ def nbd_env(monkeypatch):
 
 @broken_on_ci
 @requires_privileges
+@pytest.mark.root
```
nirs, line 135: These tests do not need root so they should not be marked for "root". They must run as non privileged user, and access systemd via supervdsm.

Please remove the marker from this module.

.github/workflows/ci.yml
```diff
@@ -20,8 +20,26 @@ jobs:
...
     steps:
     - uses: actions/checkout@v2
-    - name: Run storage tests
-      run: ./automation/tests-storage.sh
+    - name: Run storage tests not marked as root
+      run: |
+        ./automation/setup-storage.sh
+        chown -R vdsm:kvm ./ /var/tmp/vdsm*
+        su -c "./automation/tests-storage.sh" -- vdsm
+        make clean-storage
```
nirs, line 28: This logic should move into automation/tests-storage.sh.

The github yaml should contain only code reuired to run on github. Anything else should be in automation/tests-storage.sh. This way if tomorrow you want to run the tests n gitlab (most tests already run on gitlab), we can change only the yaml file.

I think this can be something like:

    ./automation/tests-storage.sh vdsm

.github/workflows/ci.yml
```diff
@@ -20,8 +20,26 @@ jobs:
...
+    container:
+      image: quay.io/ovirt/vdsm-test-centos-8
+      # Required to create loop devices.
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run storage tests marked as root
+      run: |
+        ./automation/setup-storage.sh
+        ./automation/tests-storage.sh
```
nirs, line 42: And this can be:

    ./automation/tests-storage.sh

So automation/tests-storage.sh will prepare the environment and run tox as vdsm if a user was specified.

nirs (changes requested):

tests/storage/nbd_test.py
aesteve-rh: That's why I put them in an isolated commit, I could smell that this may not be correct.

I understand the explanation, but then they are picked by the "not root" job, which will skip them because vdsm is not a privileged user. No difference, as these are broken in the ci, but that's the reasoning of the change.

I'll remove the commit though.

.github/workflows/ci.yml
```diff
@@ -20,7 +20,19 @@ jobs:
       options: --privileged
     steps:
     - uses: actions/checkout@v2
-    - name: Run storage tests
+    - name: Run storage tests not marked as root as vdsm user
```
nirs, line 23: The name is little bit too long; Maybe "Run storage tests as vdsm" and "Run storage tests as root"?

.github/workflows/ci.yml
```diff
@@ -20,7 +20,19 @@ jobs:
...
+      run: ./automation/tests-storage.sh vdsm -u=vdsm
```
nirs, line 24: Do we really need a named parameter?

I think it will much simpler to have:

    tests-storage.sh vdsm
    tests-storage.sh root

And fail the script if you call it without arguments.

automation/tests-storage.sh
```diff
@@ -1,5 +1,26 @@
...
+    echo "Run root tests as user root (default args)"
+    echo "  $ $0"
+    echo "Run root tests as user root (explicit)"
+    echo "  $ $0 root -u=root"
+    echo "Run root tests as non-root (forbidden)"
+    echo "  $ $0 root -u=vdsm"
+    echo "Run non-root tests as user root"
+    echo "  $ $0 vdsm -u=root"
+    echo "Run non-root tests as user vdsm"
+    echo "  $ $0 vdsm -u=vdsm"
```
nirs, line 21: Nice docs, but this looks over-engineered for a helper script.

automation/tests-storage.sh
```diff
@@ -12,28 +33,107 @@ create_loop_devices() {
...
+            * )
+
+                echo "Unknown option: $1"
+                print_help
+                exit 1
+            ;;
+
+        esac
+        shift
+done
```
nirs, line 102: Do we really need 35 lines to set the mode?

Why not one line?

    mode=${$1:?mode required}

Or:

    mode=$1
    if [ "$mode" != "vdsm" -a "$mode" != "root" ]; then
        show usage...
    fi

automation/tests-storage.sh
```diff
@@ -12,28 +33,107 @@ create_loop_devices() {
...
+
+# Defaults if not provided
+if [ ! -z ${user+x} ]; then
+    user=root
+fi
+if [ ! -z ${mode+x} ]; then
+    mode=root
+fi
```
nirs, line 110: Just fail if not provided.

automation/tests-storage.sh
```diff
@@ -12,28 +33,107 @@ create_loop_devices() {
...
+echo "Running $mode tests as user $user"
 
-# Setup user stoage during the tests.
-trap teardown_storage EXIT
 setup_storage
+if [ "$user" != "root" ]; then
+    # Change ownership of current and storage folders
+    # to allow non-privileged access.
+    chown -R vdsm:kvm ./ /var/tmp/vdsm*
+fi
```
nirs, line 125: Keeping both user and mode seems too complicated. Can we use only
user variable?

automation/tests-storage.sh
```diff
@@ -12,28 +33,107 @@ create_loop_devices() {
...
+# Teardown storage before exit.
+trap teardown_storage EXIT
 
-make tests-storage
+if [ "$mode" = "root" ]; then
+    # root mode
+    # Run only tests marked as root
+    tox -e storage -- --cov-fail-under 47 \
```
nirs, line 132: The coverage value is better managed as configuration value instead of hiding
in the body of the script.

The implementation should be as a separate tox env instead of duplicating
most of the testenv:storage here.

Implementing the logic here means there is no good way to run the storage 
tests directly with tox - you must use this wrapper script. If we have a tox
env that does everything, a developer can run the tests as normal user
using:

    $ tox -e storage-user

And the root tests as:

    # tox -e storage-root

So the only thing the script has to do is to change the user
before running tox and run the right target.

automation/tests-storage.sh
```diff
@@ -12,28 +33,107 @@ create_loop_devices() {
...
+if [ "$user" != "root" ]; then
+    # Change ownership of current and storage folders
+    # to allow non-privileged access.
+    chown -R vdsm:kvm ./ /var/tmp/vdsm*
+fi
+# Teardown storage before exit.
+trap teardown_storage EXIT
 
-make tests-storage
+if [ "$mode" = "root" ]; then
```
nirs, line 129: This is same as:

    "$user" != "root"

nirs (changes requested):

automation/tests-storage.sh
aesteve-rh: Ok, it makes sense to me.

I just wasn't sure if this is something we wanted in tox.ini as these environments are practically only going to be used in the pipelines. But I'll add them.

.github/workflows/ci.yml
aesteve-rh: It is how we are going to use, so the proposal makes sense.

But from the conversation we had, I understood that we wanted to differentiate 4 cases:
- Running root tests as root
- Running root tests as non-root (no possible)
- Running non-root tests as root
- Running non-root tests as vdsm (or other user?)

Even we may want to run all tests (as in `make tests-storage`). We cannot differentiate 4 cases with a single argument, hence my approach. This is more flexibility than what we require at the moment, but eases to extend the pipelines in the future.

And having root as default argument (not required) I inferred it from https://github.com/oVirt/vdsm/pull/276#discussion_r913947817

What I would do is, basically, we give the user to the script if we want to run as non-root (by default run as root). And run only the tests that match (i.e., root runs root tests, non-root runs non-root tests). This was also my first attempt at this (I didn't post it).

In this case the root user cannot run non-root tests, but again, matches how is gonna be used in the pipelines, so should be fine.

automation/tests-storage.sh
aesteve-rh: Simplifying the CLI as in https://github.com/oVirt/vdsm/pull/276#discussion_r915292609 will simplify the code to parse arguments, and make it more similar to what you wrote.

In this approach, I just didn't want to assume that the mode is first argument and user second, so I parsed them in a loop.

automation/tests-storage.sh
nirs: Why tox env are not for local use? Today we use `tox -e storage` and it skip the tests that needs root and fail when coverage is too low because the root tests were skipped.

With this change we can have:
- tox -e storage or tox -e storage-user
- tox -e storage-root

Both will run only the relevant tests and give correct coverage results. Looks like great improvement even if this was not used in the CI.

One issue with separate targets is how to get complete coverage of all tests - both user and root tests, but this can solved by another make target combining both tox envs and generating combined coverage.

The root target is basically workaround since almost no code in storage needs root, we actually need supervdsm running to run code as root, but we don't have this test infa now. When we will have it, we could remove the storage-root target and have a single target.

michalskrivanek:
I still do not follow why do you need any differentiation for CI. Do we need to run twice, once as vdsm and second time as root? 

For local you should use the same way as CI, I do not see any value in non-containerized runs. Anyone can run a container today

automation/tests-storage.sh
aesteve-rh: I didn’t mean it as it is not good, but I though that in general we would keep using the storage env for local testing. But yeah, on a second thought, this will useful for both the CIs and local.

The make target to combine coverage reports I will leave for the next patch.

.github/workflows/ci.yml
```diff
@@ -19,8 +19,20 @@ jobs:
...
       options: --privileged
     steps:
-    - uses: actions/checkout@v2
-    - name: Run storage tests
+    - uses: actions/checkout@v3
+    - name: Run storage tests not marked as root as vdsm user
+      run: ./automation/tests-storage.sh vdsm -u=vdsm
+  test-storage-root:
+    env:
+      TRAVIS_CI: 1
```
michalskrivanek, line 27: are the errors/skips still relevant? we don't run it in travis any more

.github/workflows/ci.yml
```diff
@@ -19,8 +19,20 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/ovirt/vdsm-test-centos-8
```
michalskrivanek, line 30: I'd really love to limit the amount of images we manage. For CI leveraging the https://github.com/oVirt/buildcontainer would be best, both for building and running tests

.github/workflows/ci.yml
```diff
@@ -19,8 +19,20 @@ jobs:
...
+      # Required to create loop devices.
+      options: --privileged
+    steps:
+    - uses: actions/checkout@v3
```
michalskrivanek, line 34: I don't want to diverge, all other oVirt projects use v2, Perhaps push separately as a global change everywhere?
If you do want to touch it then I'd suggest to take into account the recent discussion https://lists.ovirt.org/archives/list/devel@ovirt.org/thread/7SEFKOASOATTMO2NK2SBWMOV4CV6LZOS/#WZ3W6BII34CTQXXLBYJB6W6ECCWEGM4J

aesteve-rh: Ok. Then I will then just remove the commit for this patch. I don't see a reason why not to update it globally to v3 if we decide to take the action suggested in the discussion. Seems a nice chance to do so.

.github/workflows/ci.yml
aesteve-rh: Looks like legacy env, true. I will remove them for the updated parts in this patch, so that it is not carried over. But if removing it from the CIs is the right action to take, I would do it in a separate patch.

.github/workflows/ci.yml
aesteve-rh: Sounds reasonable. We can bring this up in the next vdsm meeting, but changing the image is out of the scope of this patch. For this change, the image of the new job is consistent with the other jobs in the CI.

.github/workflows/ci.yml
aesteve-rh: https://github.com/oVirt/vdsm/issues/279

tox.ini
```diff
@@ -97,6 +97,36 @@ commands =
     --cov-fail-under={env:STORAGE_COVERAGE:68} \
     {posargs:storage}
 
+[testenv:storage-user]
+passenv = {[base]passenv}
+setenv =
+    {[base]setenv}
+    COVERAGE_FILE=.coverage-storage
```
nirs, line 104: We alway name the coverage report after the test name, so `.coverage-storage-user`

.github/workflows/ci.yml
```diff
@@ -20,8 +20,20 @@ jobs:
       options: --privileged
```
nirs, line 20: naming this job test-storge-user will be more consistent and clear.

automation/tests-storage.sh
```diff
@@ -12,28 +24,50 @@ create_loop_devices() {
...
-make
+echo "Running tests as user $user"
 
-# Setup user stoage during the tests.
-trap teardown_storage EXIT
 setup_storage
+if [ "$user" != "root" ]; then
+    # Change ownership of current and storage folders
+    # to allow non-privileged access.
+    chown -R vdsm:kvm ./ /var/tmp/vdsm*
```
nirs, line 62: This hard codes "vdsm:kvm", but the user gave us the name of the user, so why not:

   chown -R %user:$user ...

This means that for running the tests as youself you will call:

    tests-storage.sh $USER

automation/tests-storage.sh
```diff
@@ -12,28 +24,50 @@ create_loop_devices() {
...
+# Teardown storage before exit.
+trap teardown_storage EXIT
 
-make tests-storage
+if [ "$user" = "root" ]; then
+    # Run only tests marked as root
+    make tests-storage-root
+else
+    # Run tests not marked as root as $user
+    su -c "make tests-storage-user" -- ${user}
```
nirs, line 72: Maybe:

    sudo -u $user make tests-storage-user

nirs (changes requested):
Nice!

Makefile.am
```diff
@@ -153,6 +153,14 @@ tests: tox
 tests-storage: tox
 	tox -e "storage"
```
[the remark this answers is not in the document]
aesteve-rh: And the environment?

.github/workflows/ci.yml
aesteve-rh: Ok, but then I need the project settings to change `tests-storage` from the required checks before merging for the new jobs.

automation/tests-storage.sh
aesteve-rh: It does not preserve the $PATH, to make it work it would have to be:
```
sudo -u $user env "PATH=$PATH" make tests-storage-user
```
I don't mind which option goes here.

aesteve-rh: It also loses all other env variables, which ends up making the test to ignore the `broken_on_ci` fixtures and fail: https://github.com/aesteve-rh/vdsm/runs/7355846727

I think it is better if we stick with the current strategy.

nirs: Good points, we want to keep the environment. Would be nice to add a comment why we use su instead of sudo.

automation/tests-storage.sh
nirs: Right, we should really use vdsm:kvm. But I agree that using vdsm or current user should be good enough for the tests, and we can always enhance this to use vdsm:kvm, and let the user pass both user and group. Maybe if the user does not pass a group we can default to the same value (correct for current user). The CI can pass vdsm:kvm to get more real environment.

Makefile.am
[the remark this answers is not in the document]
nirs: Sure, storage-user and storage-root replaced storage.

.github/workflows/ci.yml
nirs: Of course we need to adapt the call calling this to do this change.

tox.ini
```diff
@@ -97,6 +97,36 @@ commands =
...
+deps = {[base]deps}
+changedir = {[base]changedir}
+commands =
+    python profile {envname} pytest -m {[base]markers}" and not root" \
+    --durations=20 \
+    --cov=vdsm.storage \
+    --cov-report=html:htmlcov-storage \
+    --cov-fail-under={env:STORAGE_COVERAGE:61} \
```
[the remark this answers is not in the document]
nirs: They were used in the CI so we can control the expected coverage when CI cannot run some tests that can run locally. But even if they are not used it is good idea to keep this so we can override the value.

Makefile.am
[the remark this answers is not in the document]
aesteve-rh: Then I'll remove the tox environment in an additional commit.

automation/tests-storage.sh
```diff
@@ -12,28 +24,50 @@ create_loop_devices() {
...
```
nirs, line 72: Would be nicer as:

   su $user -c "command ..."

Makefile.am
```diff
@@ -149,9 +149,13 @@ venv:
 tests: tox
 	tox -e "tests,lib,network,virt,gluster,hooks"
 
-.PHONY: tests-storage
-tests-storage: tox
-	tox -e "storage"
```
nirs, line 154: This break the CI in case we revert the next patch. Please move this change to the next commit. Every commit must be correct.

nirs (changes requested):

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run linters
       run: ./automation/lint.sh
-  test-storage:
+  test-storage-user:
```
nirs, line 13: Please keep test-storage for now, since github setting is configured to require this build, and we don't have permissions to change it.

Add a comment to rename it once we change github settings to require test-storage-user and test-storage-root.

nirs (approved):

PR #288, opened 2022-07-24
network: bridge_opts - test priority option recognized

erav:
Change-Id: I120ae79d46006f4a7c0eafc319cceb03b8c030b0

Signed-off-by: Eitan Raviv <eraviv@redhat.com>

### DIFF
--- a/tests/network/functional/bridge_test.py
+++ b/tests/network/functional/bridge_test.py
@@ -67,12 +67,11 @@ class TestBridge(object):
 
     @nftestlib.parametrize_legacy_switch
     def test_add_bridge_with_custom_opts(self, adapter, switch, nic0):
+        opts = 'multicast_snooping=0 multicast_router=0 priority=1'
         NET_ATTRS = {
             'nic': nic0,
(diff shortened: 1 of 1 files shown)


tests/network/functional/rollback_test.py
```diff
@@ -235,7 +235,7 @@ def test_setup_invalid_bridge_opts_fails(adapter, nic0):
     net_attrs = {
         'nic': nic0,
         'switch': 'legacy',
-        'custom': {'bridge_opts': 'foo=0'},
+        'custom': {'bridge_opts': 'foo=0 priority=1'},
```
almusil, line 238: This tests failure so I wouldn't add the priority here.

erav: after rebase on the revert patches the two tests pass for me locally

almusil: They do, but it does not make sense to test priority in this test if it fails on the foo option nevertheless

almusil (approved):

PR #293, opened 2022-08-04
blockVolume: fix over allocation of small volumes

aesteve-rh:
For block sparse volumes, if no initial_size is
defined, Vdsm allocates one chunk (2.5G), without
considering the virtual size of the volume.
This results in wasted space if the requested
volume is smaller.

Instead, we can allocate the capacity when
it is smaller than one chunk, to avoid wasting
space for small volumes.
Note than any size smaller than one extent
(128M) will be rounded to the extent size.

Bug-Url: https://bugzilla.redhat.com/2094576

### DIFF
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -364,7 +364,7 @@ class BlockVolumeManifest(volume.VolumeManifest):
             else:
                 chunk_size_mb = config.getint("irs",
                                               "volume_utilization_chunk_mb")
-                alloc_size = chunk_size_mb * MiB
+                alloc_size = min(chunk_size_mb * MiB, capacity)
         else:
             # Preallocated raw or qcow2
             alloc_size = initial_size if initial_size else capacity
--- a/tests/storage/blocksd_test.py
+++ b/tests/storage/blocksd_test.py
@@ -866,7 +866,6 @@ def test_volume_metadata(tmp_storage, tmp_repo, fake_access, fake_rescan,
 
 @requires_root
 @pytest.mark.root
-@pytest.mark.xfail(reason='Vdsm over-allocates small volumes', strict=False)
 def test_cow_small_volume(domain_factory, fake_task, fake_sanlock):
     """
     Test added to verify fix for https://bugzilla.redhat.com/2094576.
--- a/tests/storage/blocksd_test.py
+++ b/tests/storage/blocksd_test.py
@@ -31,6 +31,8 @@ import string
 
 import pytest
 
+from vdsm import utils
+from vdsm.config import config
 from vdsm.common.units import MiB, GiB
 from vdsm.storage import blockSD
 from vdsm.storage import clusterlock
@@ -54,7 +56,7 @@ from storage.storagefakelib import FakeDomainMonitor
 from storage.storagefakelib import FakeTaskManager
 
 TESTDIR = os.path.dirname(__file__)
-
+CHUNK_SIZE_MB = config.getint("irs", "volume_utilization_chunk_mb")
 
 Chain = namedtuple("Chain", ["base", "internal", "top"])
 
@@ -860,6 +862,49 @@ def test_volume_metadata(tmp_storage, tmp_repo, fake_access, fake_rescan,
         data = f.read(sc.METADATA_SIZE)
     data = data.rstrip(b"\0")
     assert data == md.storage_format(domain_version, CAP=md.capacity)
+
+
+@requires_root
+@pytest.mark.root
+@pytest.mark.xfail(reason='Vdsm over-allocates small volumes', strict=False)
+def test_cow_small_volume(domain_factory, fake_task, fake_sanlock):
+    """
+    Test added to verify fix for https://bugzilla.redhat.com/2094576.
+    Small cow sparse volumes should not over-allocate beyond its capacity.
+    Small volumes in this context mean `capacity < chunk_size`.
+
+    To avoid slowness of creating loop devices and storage domains for every
+    test, avoid parametrized test, just create as many volumes as needed
+    to test, and check them in one execution.
+    """
+    # Tested volumes capacity
+    vol_sizes = [
+        CHUNK_SIZE_MB * MiB,          # capacity == chunk size
+        (CHUNK_SIZE_MB - 1) * MiB,    # capacity < chunk size
+        sc.VG_EXTENT_SIZE - 1 * MiB,  # capacity < extent size
+    ]
+    sd_uuid = str(uuid.uuid4())
+    dom = domain_factory.create_domain(sd_uuid=sd_uuid, version=5)
+
+    img_uuid = str(uuid.uuid4())
+    vol_info = [(str(uuid.uuid4()), size) for size in vol_sizes]
+
+    for vol_uuid, capacity in vol_info:
+        dom.createVolume(
+            imgUUID=img_uuid,
+            capacity=capacity,
+            volFormat=sc.COW_FORMAT,
+            preallocate=sc.SPARSE_VOL,
+            diskType=sc.DATA_DISKTYPE,
+            volUUID=vol_uuid,
+            desc="Small volume",
+            srcImgUUID=sc.BLANK_UUID,
+            srcVolUUID=sc.BLANK_UUID)
+
+        vol = dom.produceVolume(img_uuid, vol_uuid)
+        assert vol.getCapacity() == capacity
+        initial_size = utils.round(capacity, sc.VG_EXTENT_SIZE)
+        assert vol.getVolumeSize() == initial_size
 
 
 @requires_root
--- a/tests/storage/blockvolume_test.py
+++ b/tests/storage/blockvolume_test.py
@@ -45,8 +45,9 @@ from testlib import permutations, expandPermutations
 from testlib import VdsmTestCase
 
 
+CHUNK_SIZE_MB = 1024
 CONFIG = make_config([
-    ('irs', 'volume_utilization_chunk_mb', '1024'),
+    ('irs', 'volume_utilization_chunk_mb', str(CHUNK_SIZE_MB)),
     ('irs', 'volume_utilizzation_precent', '50'),
 ])
 
@@ -71,9 +72,9 @@ class TestBlockVolumeSize(VdsmTestCase):
         [(sc.PREALLOCATED_VOL, sc.COW_FORMAT, GiB, None), GiB],
         # Sparse, cow, capacity config.volume_utilization_chunk_mb - 1,
         # No initial size.
-        # Expected 1024 MiB allocated (config.volume_utilization_chunk_mb)
-        [(sc.SPARSE_VOL, sc.COW_FORMAT, (CONFIG.getint(
-            "irs", "volume_utilization_chunk_mb") - 1) * MiB, None), GiB],
+        # Expected 1023 MiB allocated (config.volume_utilization_chunk_mb - 1)
+        [(sc.SPARSE_VOL, sc.COW_FORMAT, (CHUNK_SIZE_MB - 1) * MiB, None),
+         (CHUNK_SIZE_MB - 1) * MiB],
         # Sparse, cow, capacity 4 GiB, initial size 952320 B.
         [(sc.SPARSE_VOL, sc.COW_FORMAT, 4 * GiB, 952320),
          int(952320 * blockVolume.QCOW_OVERHEAD_FACTOR)],


lib/vdsm/storage/blockVolume.py
```diff
@@ -365,6 +365,10 @@ def calculate_volume_alloc_size(
                 chunk_size_mb = config.getint("irs",
                                               "volume_utilization_chunk_mb")
                 alloc_size = chunk_size_mb * MiB
+                if alloc_size > capacity:
+                    # If the chunk size is bigger than the virtual size,
+                    # allocate the extent size instead.
+                    alloc_size = cls.align_size
```
nirs, line 371: This will allocate 128M if you try to create 512M volume. What we really want is to allocate one chunk, but limit the chunk size to the capacity:

    alloc_size = max(chunk_size_mb * MiB, capacity)

The comment is also not helpful, it describes the code instead of why the code exist. Since this is trivial change, I don't think a comment is needed.

We need to add new test verifying that we allocated the right size in this case. The best way to do such changes it to add first the test, reproducing the bug. The test must fail to prove that the test is correct. Then add the fix which will fix the failure, proving that the fix is correct.

If the fix causes other tests to fail, it is best to fix the other failing tests assuming minimal allocation of one chunk (even if capacity < chunk) before you add the fix.

tests/storage/merge_test.py
```diff
@@ -367,8 +367,8 @@ def make_env(
             sd_type='block',
             format='raw',
             prealloc=sc.SPARSE_VOL,
-            chain_len=2):
-        size = 2 * GiB
+            chain_len=2,
+            size=2 * GiB):
```
nirs, line 371: Can you explain this change?

tests/storage/merge_test.py
```diff
@@ -501,7 +501,11 @@ def failing_rebase(*args, **kw):
             assert subchain.top_vol.getLegality() == sc.ILLEGAL_VOL
 
     def test_reduce_chunked_internal(self):
-        with self.make_env(sd_type='block', format='cow', chain_len=4) as env:
+        with self.make_env(
+                sd_type='block',
+                format='cow',
+                chain_len=4,
+                size=3 * GiB) as env:
```
nirs, line 508: Why did you add size here?

tests/storage/merge_test.py
```diff
@@ -522,28 +526,6 @@ def test_reduce_chunked_internal(self):
                 ('reduce', (base_vol.optimal_size(),), {}),
             ]
 
-    def test_reduce_chunked_leaf(self):
```
nirs, line 525: The assumption that we don't need to reduce after merge is based on wrong assumption
about creating volume with one chunk. If you create the volume with initial_size=capacity the volume will be reduced after merge.

Changing merge logic because because allocation bug was fixed is not good. We want to keep the existing behavior and fix the allocation bug.

tests/storage/merge_test.py
```diff
@@ -577,7 +559,11 @@ def test_reduce_skipped(self, sd_type, format, prealloc):
             assert len(calls) == 0
 
     def test_reduce_failure(self):
-        with self.make_env(sd_type='block', format='cow', chain_len=4) as env:
+        with self.make_env(
+                sd_type='block',
+                format='cow',
+                chain_len=4,
+                size=3 * GiB) as env:
```
nirs, line 566: Not clear why you added size here.

nirs (changes requested):

lib/vdsm/storage/blockVolume.py
aesteve-rh: `What we really want is to allocate one chunk, but limit the chunk size to the capacity` -> that would be min, no?
```python
    alloc_size = min(chunk_size_mb * MiB, capacity)
```
And that would make the allocation to be smaller than an extent if the capacity is smaller (as is the case in BZ). I can max it later with the align size, maybe?

And in this case, we already had a test covering the usecase in `blockvolume_test`: `TestBlockVolumeSize`.

tests/storage/merge_test.py
aesteve-rh: With this implementation, capacities smaller than 2.5G (chunk size), allocated only one extent, and weren't reducing after merge. I changed it so that the size was bigger than the chunk size to get the reduce call in the test.

lib/vdsm/storage/blockVolume.py
nirs: Yes I sometimes get min/max wrong, and this should be min().

Getting alloc_size < align_size is fine! lvm rounds the value up to the extent size
and create volume code adjust the size properly before writing the metadata.

You can add max(align_size, alloc_size) to align the value before calling lvm if you
think it makes the code more clear, but this may be needed in other cases, so best
do this later in all cases and not only for the case fixed here.

tests/storage/merge_test.py
nirs: This changes are best done in a separate patch, with commit message explaining why the change is needed. This makes the actual change smaller and safer and keeps the rule that every commit should have exactly one logical change.

lib/vdsm/storage/blockVolume.py
aesteve-rh:
```
Getting alloc_size < align_size is fine! lvm rounds the value up to the extent size
and create volume code adjust the size properly before writing the metadata.
```
That's good to know! Then I think the max is not really needed. Maybe a comment or a specific test would suffice. Thanks for the clarification.

tests/storage/merge_test.py
aesteve-rh: `If you create the volume with initial_size=capacity the volume will be reduced after merge.` Unfortunately, that's not what I observe in the test.

The test creates block volumes using `calculate_volume_alloc_size`, which results in initial_size=capacity.
https://github.com/aesteve-rh/vdsm/blob/aesteve/small-volumes-blocksd/lib/vdsm/storage/blockVolume.py#L333

On finalize, we obtain the optimal_size and the actual_size (in this case, actual_size=capacity):

Optimal_size:
In this test we create COW volumes, so we obtain the optimal size in `optimal_cow_size`:
https://github.com/aesteve-rh/vdsm/blob/aesteve/small-volumes-blocksd/lib/vdsm/storage/blockVolume.py#L421
In this case what prevails for the optimal size calculation is this line:
https://github.com/aesteve-rh/vdsm/blob/aesteve/small-volumes-blocksd/lib/vdsm/storage/blockVolume.py#L448
Which results in `optimal_size = int(capacity * 1.1)` <- this is not in the code, but is closely equivalent to lines 448 and 449 (it also aligns the result to the extent size).

Therefore, optimal_size is always bigger than actual_size (because `int(capacity * 1.1) > capacity`), and the volume is never reduced:
https://github.com/aesteve-rh/vdsm/blob/aesteve/small-volumes-blocksd/lib/vdsm/storage/merge.py#L286
Few actual examples
- 2 GiB volumes: 2.2 GiB optimal_size > 2 GiB actual_size.
- 1 GiB volumes: 1.1 GiB optimal_size > 1 GiB actual_size.
- 1 MiB volumes: 128 MiB optimal_size = 128 MiB actual_size

aesteve-rh: On the other hand, for sizes bigger than a chunk:

`calculate_volume_alloc_size` will give an initial_size=chunk_size, for any chosen capacity in the test.

In `optimal_cow_size` the most relevant part is this one:
https://github.com/aesteve-rh/vdsm/blob/aesteve/small-volumes-blocksd/lib/vdsm/storage/blockVolume.py#L445
Which comes from `required_size (current data offset) + chunk_size` (aligned to the next extent_size).

Thus, for all these cases, we always obtain the same result:
- 3 GiB volumes (or any volume > 2.5 GiB): 2.625 GiB optimal_size > 2.5 GiB chunk size

And volume is never reduced.

tests/storage/merge_test.py
aesteve-rh: I have removed this change, was not required after changing the `alloc_size` to the capacity in:
```python
    alloc_size = min(chunk_size_mb * MiB, capacity)
```

tests/storage/merge_test.py
aesteve-rh: Finally removed this change.

tests/storage/blockvolume_test.py
```diff
@@ -72,8 +72,7 @@ class TestBlockVolumeSize(VdsmTestCase):
         # Sparse, cow, capacity config.volume_utilization_chunk_mb - 1,
         # No initial size.
         # Expected 1024 MiB allocated (config.volume_utilization_chunk_mb)
-        [(sc.SPARSE_VOL, sc.COW_FORMAT, (CONFIG.getint(
-            "irs", "volume_utilization_chunk_mb") - 1) * MiB, None), GiB],
+        [(sc.SPARSE_VOL, sc.COW_FORMAT, (CHUNK_SIZE_MB - 1) * MiB, None), GiB],
```
nirs: Not sure this is better - it is shorter code but the current code is about the actual vdsm apis used, so it is more clear what is going on. Adding a constant hide things.

tests/storage/blockvolume_test.py
```diff
@@ -51,9 +53,43 @@
 ])
 
 
+@contextmanager
+def make_volume(size, format=sc.RAW_FORMAT, prealloc=sc.SPARSE_VOL):
+    img_id = make_uuid()
+    vol_id = make_uuid()
+    # TODO fix make_volume helper to create the qcow image when needed
+    with fake_env(storage_type='block') as env:
```
nirs, line 59: Using fake_env is very likey to create wrong tests. This env try to fake a block device with file and fake lvm, and may give wrong results.

This should be really tested in blocksd_test, using real storage domain and running real createVolume() calls.

tests/storage/merge_test.py
```diff
@@ -522,6 +522,7 @@ def test_reduce_chunked_internal(self):
                 ('reduce', (base_vol.optimal_size(),), {}),
             ]
 
+    @pytest.mark.xfail(reason='leafs never allocate more than optimal')
```
nirs, line 525: Marking test as xfail because it does not work after the fix does not make sense.

If the tests does not work now change it to test the real behavior. I wonder why this test succeeded before. We should test here real volume size (e.g. 10g).

The fix should not affect merge tests, it it breaks them the tests probably not testing real use case.

tests/storage/merge_test.py
nirs: Testing 2G volume is not interesting, real disk are much bigger - changing the test to test 10g volume.

It is possible that this test used small volume size, and relied on the wrong allocation (2.5g) to test the reduce, and now when we create 2g (or 2.2) we don't need to reduce.

aesteve-rh: As I tried to expose in this comment: https://github.com/oVirt/vdsm/pull/293#discussion_r941206435

Any volume size bigger than a chunk will not get reduced, since:
- vdsm allocates one chunk -> `actual_size = chunk_size`
- optimal_size will be `chunk_size + offset` for leaf volumes
- optimal_size > actual_size

It definitively relied in the over-allocation to trigger the reduce. The test only creates the volumes and invokes `finalize()`, so if `calculate_volume_alloc_size` does not allocate more than the `optimal_size` it just never gets reduced.

tests/storage/blockvolume_test.py
aesteve-rh: For a single occurrence is ok, for multiple occurrences it made the permutations more difficult to read.

I think the constant declared in consecutive lines with the `CONFIG`, makes it clear that the permutations use the direct size, and the test shall obtain the same size from the `CONFIG` structure that is monkeypatched.

Which btw made me realize that there is a typo in the `CONFIG`, `volume_utilizzation_precent` is wrong.

tests/storage/blockvolume_test.py
aesteve-rh: I'm ok with adding a `blocksd_test`, but I'd still keep this one, it is consistent with the other tests in the module, checking size, allocation, max_size, and optimal_size with the same strategy and code style.

tests/storage/merge_test.py
aesteve-rh: That's why I removed it in the first place, xfail just shows that it is indeed failing in the pipelines.

It was working because we were overallocating for small volumes and the test levarages that. Bigger volumes (e.g., 10g) will not work, with current code or previous code.

In general what is making this test fail (vs other similar test) is that leaf volumes have an increased optimal_size, which makes it consistently bigger than the allocated size: 
https://github.com/aesteve-rh/vdsm/blob/aesteve/small-volumes-blocksd/lib/vdsm/storage/blockVolume.py#L432

tests/storage/merge_test.py
aesteve-rh: One way to make it pass is forcing over-allocation through fakery, e.g.:
```python
        monkeypatch.setattr(
            blockVolume.BlockVolumeManifest,
            "calculate_volume_alloc_size",
            lambda *x: 4 * GiB)
```
I guess we could also write and then free some space in the volume, so that it can be reduced.

But probably this test does not make sense anymore and can be removed.
- Sparse leaf volumes will never be over-allocated with the bug-fix in this branch, so will never reduce.
- Preallocated volumes are not allowed to be reduced.

tests/storage/merge_test.py
aesteve-rh: I'll close this thread to not to diverge from the other [thread](https://github.com/oVirt/vdsm/pull/293#discussion_r938228290), which is a related discussion (aka what to do with the test that now fails no matter which size we use).

tests/storage/merge_test.py
aesteve-rh: Fixed with https://github.com/oVirt/vdsm/pull/316

tests/storage/blockvolume_test.py
```diff
@@ -71,9 +104,9 @@ class TestBlockVolumeSize(VdsmTestCase):
         [(sc.PREALLOCATED_VOL, sc.COW_FORMAT, GiB, None), GiB],
         # Sparse, cow, capacity config.volume_utilization_chunk_mb - 1,
         # No initial size.
-        # Expected 1024 MiB allocated (config.volume_utilization_chunk_mb)
-        [(sc.SPARSE_VOL, sc.COW_FORMAT, (CONFIG.getint(
-            "irs", "volume_utilization_chunk_mb") - 1) * MiB, None), GiB],
+        # Expected 1023 MiB allocated (config.volume_utilization_chunk_mb - 1)
```
barpavel, line 107: Not sure whether the comment should be updated (in lines 105 & 107): `config.volume_utilization_chunk_mb` --> `CHUNK_SIZE_MB`.

tests/storage/blocksd_test.py
```diff
@@ -862,6 +863,42 @@ def test_volume_metadata(tmp_storage, tmp_repo, fake_access, fake_rescan,
...
+    (CHUNK_SIZE_MB + 1) * MiB - sc.VG_EXTENT_SIZE,
+    (CHUNK_SIZE_MB - 1) * MiB,
+])
+def test_cow_volume_small_size(
+        domain_factory, capacity, fake_task, fake_sanlock):
+    sd_uuid = str(uuid.uuid4())
+    dom = domain_factory.create_domain(sd_uuid=sd_uuid, version=5)
+
+    img_uuid = str(uuid.uuid4())
+    vol_id = str(uuid.uuid4())
```
barpavel, line 883: Suggestion:
`vol_id` --> `vol_uuid`.

tests/storage/blockvolume_test.py
```diff
@@ -44,16 +44,49 @@
...
-    ('irs', 'volume_utilization_chunk_mb', '1024'),
+    ('irs', 'volume_utilization_chunk_mb', str(CHUNK_SIZE_MB)),
     ('irs', 'volume_utilizzation_precent', '50'),
 ])
 
 
+@contextmanager
+def make_volume(size, format=sc.RAW_FORMAT, prealloc=sc.SPARSE_VOL):
+    img_id = make_uuid()
+    vol_id = make_uuid()
```
barpavel, line 57: Maybe better use `img_uuid` & `vol_uuid`?

barpavel (approved):

tests/storage/blockvolume_test.py
aesteve-rh: ok, this is actually moved code, not created new, so I moved it as is. But it is ok, I can update the name of these variables in the process.

tests/storage/blockvolume_test.py
```diff
@@ -44,16 +44,49 @@
...
+    # TODO fix make_volume helper to create the qcow image when needed
```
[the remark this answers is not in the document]
aesteve-rh: Same as before, I moved this line as is. I'm fine with a small update though :)

tests/storage/blockvolume_test.py
aesteve-rh: Both are true, and leaving the config in the comment hints for the source of the `CHUNK_SIZE_MB`, so I think we can keep it as is?

tests/storage/blocksd_test.py
```diff
@@ -54,7 +55,7 @@
 from storage.storagefakelib import FakeTaskManager
 
 TESTDIR = os.path.dirname(__file__)
-
+CHUNK_SIZE_MB = blockSD.config.getint("irs", "volume_utilization_chunk_mb")
```
nirs, line 58: Any reason you use blockSD.config instead of importing config and fetching the value from the config?

tests/storage/blocksd_test.py
```diff
@@ -862,6 +863,43 @@ def test_volume_metadata(tmp_storage, tmp_repo, fake_access, fake_rescan,
     assert data == md.storage_format(domain_version, CAP=md.capacity)
 
 
+@requires_root
+@pytest.mark.root
+@pytest.mark.xfail(reason='Vdsm over-allocates one chunk', strict=False)
```
nirs: vdsm over-allocates small volumes

tests/storage/blocksd_test.py
```diff
@@ -862,6 +863,43 @@ def test_volume_metadata(tmp_storage, tmp_repo, fake_access, fake_rescan,
...
+@pytest.mark.parametrize("capacity", [
+    # capacity
+    1 * MiB,
+    512 * MiB,
+    sc.VG_EXTENT_SIZE,
+    CHUNK_SIZE_MB * MiB - sc.VG_EXTENT_SIZE,
+    (CHUNK_SIZE_MB + 1) * MiB - sc.VG_EXTENT_SIZE,
+    (CHUNK_SIZE_MB - 1) * MiB,
+])
+def test_cow_volume_small_size(
```
nirs, line 877: test_cow_small_volume?

A comment explaining what is small volume can help (e.g. capacity < chunk size)

tests/storage/blocksd_test.py
```diff
@@ -862,6 +863,43 @@ def test_volume_metadata(tmp_storage, tmp_repo, fake_access, fake_rescan,
...
+@pytest.mark.root
+@pytest.mark.xfail(reason='Vdsm over-allocates one chunk', strict=False)
+@pytest.mark.parametrize("capacity", [
+    # capacity
+    1 * MiB,
+    512 * MiB,
+    sc.VG_EXTENT_SIZE,
+    CHUNK_SIZE_MB * MiB - sc.VG_EXTENT_SIZE,
+    (CHUNK_SIZE_MB + 1) * MiB - sc.VG_EXTENT_SIZE,
+    (CHUNK_SIZE_MB - 1) * MiB,
```
nirs, line 875: Not sure why we need so many tests. The interesting cases are:
- capacity == chunk size -> allocate capacity
- capacity == chunk size - 1 -> allocate capacity
- capacity < extent size -> allocate extent size
- capacity > chunk size -> allocated chunk (but this must be already covered by other tests)

(Allocations are always rounded up to extent size)

These tests are slow, so minimizing the number of tests is very important. Using parameters for this tests is really bad - we will create the loop devices and storage domain for creating a single volume for every test!

All the cases should be tested with the same storage domain. You can create one test that create 3 volumes instead of using parameters.

tests/storage/blockvolume_test.py
```diff
@@ -51,9 +53,43 @@
...
+        [CHUNK_SIZE_MB * MiB - sc.VG_EXTENT_SIZE],
+        [(CHUNK_SIZE_MB + 1) * MiB - sc.VG_EXTENT_SIZE],
+        [(CHUNK_SIZE_MB - 1) * MiB],
+    ])
+    @MonkeyPatch(blockVolume, 'config', CONFIG)
+    def test_cow_volume_small_size(self, capacity):
+        with make_volume(capacity, format=sc.COW_FORMAT) as vol:
+            self.assertEqual(vol.getCapacity(), capacity)
+            initial_size = utils.round(capacity, sc.VG_EXTENT_SIZE)
+            self.assertEqual(vol.getVolumeSize(), initial_size)
```
nirs, line 88: This repeats the blocksd tests in a bad way. I would delete the duplicate tests here instead of wasting time on the fake storage.

tests/storage/blocksd_test.py
aesteve-rh: No reason other than to avoid having the extra import. But does no harm and will make it more clear, so I will change it. As it is, the real path to config gets obscured.

tests/storage/blocksd_test.py
aesteve-rh: Ok, I did put the three usecases together, avoided using parametrized test, and commented the whole test to help to read it.

tests/storage/blockvolume_test.py
aesteve-rh: Removed.

tests/storage/blockvolume_test.py
```diff
@@ -73,7 +73,8 @@ class TestBlockVolumeSize(VdsmTestCase):
         # No initial size.
         # Expected 1024 MiB allocated (config.volume_utilization_chunk_mb)
         [(sc.SPARSE_VOL, sc.COW_FORMAT, (CONFIG.getint(
-            "irs", "volume_utilization_chunk_mb") - 1) * MiB, None), GiB],
+            "irs", "volume_utilization_chunk_mb") - 1) * MiB, None),
+         (CONFIG.getint("irs", "volume_utilization_chunk_mb") - 1) * MiB],
```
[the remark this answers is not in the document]
aesteve-rh: I removed the constant because you seemed not convinced before, and after removing all the additional cases and test, my justification for having it (same as yours, more clear test data) felt not as relevant anymore.

I can bring it back, no problem. Lets will wait for the ost test to finish, and I will push the update.

nirs (approved):

aesteve-rh:
Last OST run finished successfully: https://rhv-devops-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/ds-ost-baremetal-gh/1619/

@nirs or @mz-pdm this can be merged if everything else looks ok. And then we'll be ready for the next build :)

Issue #151, opened 2022-04-22
Replace vdsm version of userstorage with pip's userstorage


aesteve-rh:
### Description

VDSM uses an outdated version of userstorage stored locally in the project (i.e., `tests/storage/userstorage.py`).

### Proposal

Change VDSM version of userstorage to be used as in [imageio](https://github.com/oVirt/ovirt-imageio), where userstorage is a requirement, thus installed with pip in the venv, updateable, and ready-to-use from command line directly.

PR #294, opened 2022-08-10
Replace userstorage

aesteve-rh:
Vdsm uses an outdated version of userstorage
stored locally in the project.

Change it so that it uses the installed
package from PyPI.

Fixes: https://github.com/oVirt/vdsm/issues/151


aesteve-rh:
https://github.com/aesteve-rh/vdsm/actions/runs/2832404676

This is a test-run in my fork, using the containers uploaded to my `quay.io`, and they passed.
If everything else looks correct, I will update the containers to make the pipelines pass.

tests/storage/conftest.py
```diff
@@ -146,8 +148,8 @@ def tmp_storage(monkeypatch, tmpdir):
 @pytest.fixture(
     scope="module",
     params=[
-        userstorage.PATHS["mount-512"],
-        userstorage.PATHS["mount-4k"],
+        BACKENDS["mount-512"],
+        BACKENDS["mount-4k"],
```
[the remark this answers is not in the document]
aesteve-rh: Right, I noticed the wrapper, but I wasn't sure if I should include it in this patch or not. Fixed!

aesteve-rh:
https://github.com/aesteve-rh/vdsm/pull/1
Again, to show that it passes the pipelines in its current status.

automation/common.sh
```diff
@@ -182,14 +182,11 @@ run_tests_storage() {
     create_loop_devices 16
```
nirs, line 182: We don't use this now, and there is no need to change it. The only change needed is to delete the entire directory (#43).

tests/storage/filesd_metadata_test.py
```diff
@@ -88,20 +88,21 @@
 _SHA_CKSUM=769baef6c65aeef08cf6d177c2c44046b2aac877
 """,
 }
+BACKENDS = userstorage.load_config("storage.py").BACKENDS
```
nirs, line 92: This will work only if the current directory is "tests/storage" is this really true?

tests/storage/filesd_metadata_test.py
```diff
@@ -88,20 +88,21 @@
...
         pytest.param(
-            (userstorage.PATHS["file-4k"], sc.ALIGNMENT_1M),
+            (BACKENDS["file-4k"], sc.ALIGNMENT_1M),
             id="file-4k-1m"),
         pytest.param(
-            (userstorage.PATHS["file-4k"], sc.ALIGNMENT_2M),
+            (BACKENDS["file-4k"], sc.ALIGNMENT_2M),
             id="file-4k-2m"),
     ],
 )
```
nirs, line 109: I already commented - we need to call setup() and teardown() for every test. Using the parameters is not enough. We need to add a wrapper calling setup and teardown.

tests/storage/conftest.py
```diff
@@ -154,10 +155,8 @@ def tmp_storage(monkeypatch, tmpdir):
     ids=str,
 )
 def tmp_mount(request):
-    mount = request.param
-    if not mount.exists():
-        pytest.xfail("{} storage not available".format(mount.name))
-    return mount
+    with Backend(request.param) as backend:
+        yield backend
```
nirs, line 159: This is fine, but the previous patch is broken without this. We cannot commit broken
patches fixed by the next patch. The backend wrapper should be added first and when the code is ready we can replace the builtin userstorage with the upstream version.

tests/storage/filesd_metadata_test.py
```diff
@@ -108,14 +109,11 @@
...
-        pytest.xfail("{} storage not available".format(storage.name))
-
-    with open(storage.path, "w") as f:
-        f.truncate(0)
-
-    yield Storage(storage.path, storage.sector_size, alignment)
-    oop.stop()
+    with Backend(storage) as backend:
+        with open(backend.path, "w") as f:
+            f.truncate(0)
```
nirs, line 114: The purpose of the backend is to avoid such code. Using the backend enstures that the volume is cleaned up before/after test.

tests/storage/outofprocess_test.py
```diff
@@ -61,13 +62,11 @@ def oop_cleanup():
...
-        pytest.xfail("{} storage not available".format(storage.name))
-    tmpdir = os.path.join(storage.path, "tmp")
-    os.mkdir(tmpdir)
-    yield tmpdir
-    shutil.rmtree(tmpdir)
+    with Backend(request.param) as backend:
+        tmpdir = os.path.join(backend.path, "tmp")
+        os.mkdir(tmpdir)
+        yield tmpdir
+        shutil.rmtree(tmpdir)
```
nirs, line 69: This looks like generic helper that many tests could use - providing a temporary directory backed by user storage. Looks like good addition to the userstorage library.

For now I would add something like:

    with Backend(request.param) as backend:
        yield backend.tmpdir()

tests/storage/storage_backend.py
```diff
@@ -0,0 +1,47 @@
...
+    simplify fixtures.
+    """
+
+    def __init__(self, storage, can_detect_sector_size=True):
+        if not storage.exists():
+            pytest.xfail(f"Storage {storage.name} is not available")
+
+        self._storage = storage
+        self.path = storage.path
+        self.url = urllib.parse.urlparse(f"file:{storage.path}")
```
nirs, line 38: url is useful in imageio when everything works with url, but not in vdsm.

tests/storage/storage_backend.py
```diff
@@ -0,0 +1,47 @@
...
+        self.block_size = storage.sector_size
+        self.can_detect_sector_size = can_detect_sector_size
+
+    def __enter__(self):
+        self._storage.setup()
+        return self
+
+    def __exit__(self, *args):
+        self._storage.teardown()
```
nirs, line 47: We can add here:

    @contextmanager
    def tmpdir(self):
        yield tempfile.TemporaryDirectory(dir=self.path)

Of course this works only with "mount" user storage, so it may need to be smarter.

tests/storage/xlease_test.py
```diff
@@ -151,13 +150,12 @@ def close(self):
...
-    tv = TemporaryVolume(backend, alignment, block_size=storage.sector_size)
-
-    yield tv
-    tv.close()
+    with Backend(storage):
+        backend = xlease.DirectFile(storage.path)
+        tv = TemporaryVolume(
+            backend, alignment, block_size=storage.sector_size)
+        yield tv
+        tv.close()
```
nirs, line 158: Looks like good way to extend the backend.

tests/storage/README.md
```diff
@@ -45,8 +49,7 @@ environment are skipped automatically. The tests must deal with the
 missing storage configuration by detecting that the storage does not
 exist.
 
-To add new storage configurations edit the STORAGE list in the
-userstorage.py tool.
+To add new storage configurations edit the BACKENDS list in storage.py.
 
 
 ## How user storage works?
```
nirs, line 55: Maybe delete this and point to the upstream documentation?

If this is not documented upstream, maybe move this documentation
upstream.

nirs (changes requested):

tests/storage/filesd_metadata_test.py
aesteve-rh: Yes, tox path defaults to `tests/storage` for the `storage` environment. If we overwrite defaults and select a given test inside this folder, in this case `storage/filesd_metadata_test.py`, it will also work.

tests/storage/storage_backend.py
aesteve-rh: Sounds good. But as is it does not work. This yields a `TemporaryDirectory` instance, therefore we cannot do this:
```python
with Backend(storage) as backend:
    yield backend.tmpdir()
```
As it just gives the instance. We want to enter the context and get the directory path, so we need to do:
```python
with Backend(storage) as backend:
    with backend.tmpdir() as tmpdir:
        return tmpdir
```
But in this case it does not execute the TemporaryDirectory `__enter__` method, since we wrapped the context manager class with the `contextmanager` decorator. So it is not what we want either: we do not get the directory name, but the class instance; and it will not be cleaned up at exit.

Instead, I just return the class without the `@contextmanager` decorator.

tests/storage/xlease_test.py
aesteve-rh: This is very specific for xlease testing, no?

tests/storage/README.md
aesteve-rh: True, it is documented upstream. I'll link it.

tests/storage/xlease_test.py
nirs: yes, it shows we can extend the backend easily.

tests/storage/storage_backend.py
nirs: Sounds good.

nirs (changes requested):

tests/storage/storage_backend.py
```diff
@@ -0,0 +1,48 @@
...
+    simplify fixtures.
+    """
+
+    def __init__(self, storage):
+        if not storage.exists():
+            pytest.xfail(f"Storage {storage.name} is not available")
+
+        self._storage = storage
+        self.path = storage.path
+        self.block_size = storage.sector_size
```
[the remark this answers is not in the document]
aesteve-rh: That's something that crossed my mind on the [PR ](https://github.com/nirs/userstorage/pull/11)in userstorage (having `block_size` as an attribute, aliasing sector_size). Sounds good :+1:

aesteve-rh:
@nirs this should be ready to review (and hopefully merge)

tests/storage/conftest.py
```diff
@@ -56,8 +57,11 @@
 from . import tmpfs
 from . import tmprepo
 from . import tmpstorage
-from . import userstorage
 
+# Mark tests with xfail if userstorage is missing
+userstorage.missing_handler = pytest.xfail
```
nirs, line 62: Blank line here will help.

tests/storage/outofprocess_test.py
```diff
@@ -53,19 +54,15 @@ def oop_cleanup():
...
-    storage = request.param
-    if not storage.exists():
-        pytest.xfail("{} storage not available".format(storage.name))
-    tmpdir = os.path.join(storage.path, "tmp")
-    os.mkdir(tmpdir)
-    yield tmpdir
-    shutil.rmtree(tmpdir)
+    with request.param:
+        with request.param.tmpdir() as tmpdir:
+            yield tmpdir
```
nirs, line 65: It would be more consistent and more clear like this:

    backend = request.param
    with backend:
        with backend.tmpdir() as tmpdir:
            yield tmpdir

tests/storage/sdm_copy_data_test.py
```diff
@@ -66,21 +66,20 @@
...
     ],
     ids=str,
 )
 def user_mount(request):
-    mount = request.param
-    if not mount.exists():
-        pytest.xfail("{} storage not available".format(mount.name))
-    return mount
+    with request.param:
+        yield request.param
```
nirs, line 82: Previous code tried to make it more clear by using:

    mount = request.param

We can can do the same, or use:

    backend = request.param

tests/storage/xlease_test.py
```diff
@@ -64,16 +66,14 @@ def pwrite(self, offset, buf):
...
     ],
     ids=str,
 )
 def user_storage(request):
-    storage = request.param
-    if not storage.exists():
-        pytest.xfail("{} storage not available".format(storage.name))
-    return storage
+    with request.param:
+        yield request.param
```
nirs, line 76: Again this removes the explaining variable (storage) - this change is not required.
I think using "backend" like most test will help to make the code easier to work with.

nirs (approved):

PR #296, opened 2022-08-11
spec: Use vdsm.spec directly rather than generating it

mz-pdm:
We currently generate vdsm.spec from vdsm.spec.in.  This means that
whatever we want to do regarding building Vdsm packages, we must run
auto* stuff first and having all the necessary tools available before
we start with the building process, which is not always convenient or
even possible.

The only thing that is replaced in vdsm.spec.in when generating
vdsm.spec from it is @PACKAGE_RELEASE@.  If we put it to a separate
file included into the spec using %include then we needn’t generate
vdsm.spec anymore and can use it directly.  Let’s add vdsm-version.in
file for this purpose and let’s use vdsm.spec instead of vdsm.spec.in.

While @PACKAGE_RELEASE@ must be still replaced in the included file,
it can be done not only using autotools but also manually by running

  something-simple > vdsm-version

For example:

  { echo -n '%global vdsm_release '; ./build-aux/pkg-version --release; } > vdsm-version

Or

  echo '%global vdsm_release 1' > vdsm-version

for tagged releases.


mz-pdm:
ping

michalskrivanek:
lgtm

tinez:
Generally I'm ok with the taken direction, but could you elaborate on what problem are we specifically trying to solve?
I think the same can be achieved by `sed 's/@PACKAGE_RELEASE@/1' vdsm.spec.in > vdsm.spec` if needed.
Some thoughts I have on this PR:
- would it be beneficial to not require the presence of `vdsm-release` at all by having `%{!?vdsm_release: %include vdsm-release}` in the spec?
- in tarballs we already have a `VERSION` file, so seeing that one along with `vdsm-version` and `vdsm-version.in` makes things a bit confusing

mz-pdm:
> Generally I'm ok with the taken direction, but could you elaborate on what problem are we specifically trying to solve?

It should ease automated d/s builds but I think having a less or more directly usable .spec file in the repo is good generally.  @galitf can provide details about the particular needs for d/s builds.

> I think the same can be achieved by `sed 's/@PACKAGE_RELEASE@/1' vdsm.spec.in > vdsm.spec` if needed.

Yes (just a missing ending slash in `s/.../.../`?).  But, @galitf, IIRC you need a .spec file in the upstream repo, right?

> Some thoughts I have on this PR:
> - would it be beneficial to not require the presence of `vdsm-release` at all by having `%{!?vdsm_release: %include vdsm-release}` in the spec?

A very good idea!

> - in tarballs we already have a `VERSION` file, so seeing that one along with `vdsm-version` and `vdsm-version.in` makes things a bit confusing

Right, thanks for pointing this out.

@galitf may need to include additional lines in the spec file but I think this is a separate problem that can be solved either by adding another `%include` (although the `%include` in `%{!?vdsm_release: %include vdsm-release}` could be perhaps abused for this) or (preferably) by maintaining a modified .spec in a separate d/s branch.

mz-pdm:
How about this simplified version? Does it miss anything we need?

As why we cannot generate the release dynamically from the spec file, it's explained in https://github.com/oVirt/vdsm/commit/d9dc07f3c422bc5cc4fcc80d98cea36cad1ad5c1: "A spec file doesn't know the location of the source directory when generating a SRPM and thus can't run pkg-version script from it."

This simplified version may not make @galitf particularly happy because it doesn't allow including additional stuff, but this can be resolved either by adding the lines to the .spec file from the repo or by another patch, based on the particular needs.

PR #301, opened 2022-08-23
containers: Don't use arch in COPR chroot name

tinez:
Turns out there's no need to specify the architecture in COPR's chroot
names, so let's not do this.


mwperina (approved):
+1

erav (approved):

Issue #292, opened 2022-08-01
Fix creating a preallocated qcow2 disk


aesteve-rh:
Running the command:
```
$ ./ovirt-img upload-disk -c engine --storage-domain nfs-01 /data/tmp/fedora-34.qcow2 --preallocated
```
...does not create preallocated file (with qemu-img info):
Virtual Size: 50 GiB
Actual Size: 23 GiB

So, for qcow2 format we don't do preallocation even if engine asked for preallocated format.
Needs investigation. There may be a good reason why we don't preallocate, but it may also be an error.

aesteve-rh:
## For block storage:

```python
if preallocate == sc.SPARSE_VOL:
...
else:
    # Preallocated qcow2
    alloc_size = initial_size if initial_size else capacity
```

So upload-disk was (should not in the latest version) requesting a new preallocated qcow2 disk with `initial_size` 23 GiB, and `virtual_size` 50 GiB. With this request, Vdsm takes into consideration the `initial_size` that engine has sent in the request and creates the disk, even for preallocated disks.

## For file storage (as in the initial message):

`preallocate` is not considered at all for COW volumes (?).

PR #304, opened 2022-08-26
fileVolume: consider preallocation for qcow

aesteve-rh:
File volume create calls do not consider
preallocation policy for qcow2 format, which
are thin provisioned regardless.

Vdsm should do as commanded by the engine and
create a preallocated volume, both for raw and
qcow2 volumes.

Fixes: https://github.com/oVirt/vdsm/issues/292

### DIFF
--- a/lib/vdsm/storage/fileVolume.py
+++ b/lib/vdsm/storage/fileVolume.py
@@ -521,11 +521,9 @@ class FileVolume(volume.Volume):
             cls.log.info("Request to create COW volume %s with capacity = %s",
                          vol_path, capacity)
 
-            operation = qemuimg.create(vol_path,
-                                       size=capacity,
-                                       format=sc.fmt2str(sc.COW_FORMAT),
-                                       qcow2Compat=dom.qcow2_compat())
-            operation.run()
+            cls._create_image(vol_path, capacity,
+                              format=sc.fmt2str(sc.COW_FORMAT),
+                              qcow2_compat=dom.qcow2_compat())
         else:
             # Create hardlink to template and its meta file
             cls.log.info("Request to create snapshot %s/%s of volume %s/%s "
@@ -556,17 +554,7 @@ class FileVolume(volume.Volume):
 
     @classmethod
     def _allocate_volume(cls, vol_path, size, preallocate):
-        # Always create sparse image, since qemu-img create uses
-        # posix_fallocate() which is inefficient and harmful.
-        op = qemuimg.create(vol_path, size=size, format=qemuimg.FORMAT.RAW)
-
-        # This is fast but it can get stuck if storage is inaccessible.
-        with vars.task.abort_callback(op.abort):
-            with utils.stopwatch(
-                    "Creating image {}".format(vol_path),
-                    level=logging.INFO,
-                    log=cls.log):
-                op.run()
+        cls._create_image(vol_path, size, sc.fmt2str(sc.RAW_FORMAT))
 
         # If the image is preallocated, allocate the rest of the image
         # using fallocate helper. qemu-img create always writes zeroes to
@@ -582,6 +570,21 @@ class FileVolume(volume.Volume):
                         level=logging.INFO,
                         log=cls.log):
                     op.run()
+
+    @classmethod
+    def _create_image(cls, vol_path, size, format, qcow2_compat=None):
+        # Always create sparse image, since qemu-img create uses
+        # posix_fallocate() which is inefficient and harmful.
+        op = qemuimg.create(vol_path, size=size,
+                            format=format, qcow2Compat=qcow2_compat)
+
+        # This is fast but it can get stuck if storage is inaccessible.
+        with vars.task.abort_callback(op.abort):
+            with utils.stopwatch(
+                    "Creating image {}".format(vol_path),
+                    level=logging.INFO,
+                    log=cls.log):
+                op.run()
 
     @classmethod
     def _set_permissions(cls, vol_path, dom):
--- a/lib/vdsm/storage/fileVolume.py
+++ b/lib/vdsm/storage/fileVolume.py
@@ -495,7 +495,13 @@ class FileVolume(volume.Volume):
 
         cls._truncate_volume(vol_path, 0, vol_id, dom)
 
-        cls._allocate_volume(vol_path, capacity, preallocate=preallocate)
+        cls._create_image(vol_path, capacity, sc.fmt2str(sc.RAW_FORMAT))
+
+        # If the image is preallocated, allocate the rest of the image
+        # using fallocate helper. qemu-img create always writes zeroes to
+        # the first block so we should skip it during preallocation.
+        if preallocate == sc.PREALLOCATED_VOL:
+            cls._preallocate_volume(vol_path, capacity, offset=4 * KiB)
 
         # Forcing volume permissions in case qemu-img changed the permissions.
         cls._set_permissions(vol_path, dom)
@@ -551,16 +557,6 @@ class FileVolume(volume.Volume):
             if e.errno == errno.EEXIST:
                 raise se.VolumeAlreadyExists(vol_id)
             raise
-
-    @classmethod
-    def _allocate_volume(cls, vol_path, size, preallocate):
-        cls._create_image(vol_path, size, sc.fmt2str(sc.RAW_FORMAT))
-
-        # If the image is preallocated, allocate the rest of the image
-        # using fallocate helper. qemu-img create always writes zeroes to
-        # the first block so we should skip it during preallocation.
-        if preallocate == sc.PREALLOCATED_VOL:
-            cls._preallocate_volume(vol_path, size, offset=4 * KiB)
 
     @classmethod
     def _create_image(cls, vol_path, size, format, qcow2_compat=None):
--- a/lib/vdsm/storage/fileVolume.py
+++ b/lib/vdsm/storage/fileVolume.py
@@ -450,14 +450,18 @@ class FileVolume(volume.Volume):
         """
         Class specific implementation of volumeCreate.
         """
-        if volFormat == sc.RAW_FORMAT:
-            return cls._create_raw_volume(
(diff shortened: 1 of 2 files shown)


lib/vdsm/storage/fileVolume.py
```diff
@@ -552,11 +554,13 @@ def _truncate_volume(cls, vol_path, capacity, vol_id, dom):
...
     @classmethod
-    def _allocate_volume(cls, vol_path, size, preallocate):
+    def _allocate_volume(
+            cls, vol_path, size, preallocate, vol_format, cow_compat=None):
         try:
             # Always create sparse image, since qemu-img create uses
             # posix_fallocate() which is inefficient and harmful.
-            op = qemuimg.create(vol_path, size=size, format=qemuimg.FORMAT.RAW)
+            op = qemuimg.create(
+                vol_path, size=size, format=vol_format, qcow2Compat=cow_compat)
```
nirs, line 563: This creates a sparse image with virtual_size=size...

```
$ qemu-img create -f qcow2 /data/tmp/test.qcow2 1g
Formatting '/data/tmp/test.qcow2', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=1073741824 lazy_refcounts=off refcount_bits=16
```

```
$ ls -lhs /data/tmp/test.qcow2
196K -rw-r--r--. 1 nsoffer nsoffer 193K Aug 27 17:43 /data/tmp/test.qcow2
```

But in line 577 we run:

```
            # If the image is preallocated, allocate the rest of the image
            # using fallocate helper. qemu-img create always writes zeroes to
            # the first block so we should skip it during preallocation.
            if size > 4096 and preallocate == sc.PREALLOCATED_VOL:
                op = fallocate.allocate(vol_path, size - 4096, offset=4096)
```

Which write zeros to the image after offset 4096, overwriting data in offset 4k-196k, corrupting the qcow2 image.

Lets try this:

```
$ helpers/fallocate $((1024*1024*1024 - 4096)) --offset 4096 /data/tmp/test.qcow2 
```

```
$ ls -lhs /data/tmp/test.qcow2
1.1G -rw-r--r--. 1 nsoffer nsoffer 1.0G Aug 27 17:47 /data/tmp/test.qcow2
```

```
$ qemu-img info /data/tmp/test.qcow2 
image: /data/tmp/test.qcow2
file format: qcow2
virtual size: 1 GiB (1073741824 bytes)
disk size: 1 GiB
cluster_size: 65536
Format specific information:
    compat: 1.1
    compression type: zlib
    lazy refcounts: false
    refcount bits: 16
    corrupt: false
    extended l2: false
```

```
$ qemu-img check /data/tmp/test.qcow2 
Image end offset: 262144
```

So we were lucky, and there was no data after offset 4096, but this is unsafe. Future version of qcow2 can write data after offset 4096, and this change will corrupt the image.

We can make this safe using:

    offset = qemuimg.check(vol_path, format="qcow2")["offset"]

    # qemu-img offset is usually not aligned to 4k, and fallocate uses
    # direct I/O requiring alignment to logical block size. Using 4k
    # alignment ensures this works for any storage.
    offset = utils.round(offset, 4 * KiB)

    op = fallocate.allocate(vol_path, size - offset, offset=offset)
    ...

Reporting image end offset in "qemu-img check" is public API that is safe to use for truncating or reducing block devices to the end of the image, so it will always be safe for allocating additional space after the end of the image.

This works only for qcow2 format, for raw format st_size is always the virtual size,
and we cannot overwrite any data. Actually we could always allocated the entire size since new raw empty image must be empty.

But fixing this issue by modifying the code to create raw volume is the wrong approach. You may introduce regressions in creating raw images by adding support for preallocating qcow2, and the code becomes complicated and hard to understand. The best way is to have separate functions (allocate_raw_volume, allocatte_qcow2_volume) with simple and clear logic, and share some low level helpers like (run_fallocate_task) which both functions need.

tests/storage/localfssd_test.py
```diff
@@ -682,6 +682,52 @@ def test_volume_create_cow_sparse_with_parent(user_domain, local_fallocate):
...
+        path=path,
+        format=qemuimg.FORMAT.QCOW2,
+        virtual_size=PREALLOCATED_VOL_SIZE,
+        qemu_info=qemu_info)
+
+    # Actual size is bigger than than preallocated size as it includes
+    # the qcow2 header, which is one cluster size plus the variable sized
+    # tables. We could calculate the value for the the preallocated size in
+    # this check, but checking than the actual size is bigger should suffice.
+    assert qemu_info['actual-size'] > PREALLOCATED_VOL_SIZE
```
nirs, line 715: This will not be correct for qcow2 image since we don't use qemu to do the preallocation. For preallocated qcow2 we can test that

    actual size >= virutal size.

This check may work on some file systems (xfs) and fail on others. qemu-img configure xfs to use extent size of 1m, which may cause it to report bigger actual size. We don't want to assume this for any file system.

lib/vdsm/storage/blockVolume.py
```diff
@@ -366,7 +365,6 @@ def calculate_volume_alloc_size(
                                               "volume_utilization_chunk_mb")
                 alloc_size = chunk_size_mb * MiB
         else:
-            # Preallocated qcow2
```
nirs, line 369: How is this related to preallocation on file volumes? If this is unrelated to the change, best submit this as separate pr to avoid confusing reviewers.

nirs (changes requested):
Nice! we always assumed that we don't have a good way to do this but this pr show that it is simple and you covered this change with a good test.

lib/vdsm/storage/fileVolume.py
aesteve-rh: Ah, is true. I should be more careful with the qcow2 format, and its header :facepalm: 

I will try what you proposed and post it.

Alternatively, we may want to consider using the preallocation `falloc` in the `qemu-img create` call: https://github.com/oVirt/vdsm/blob/master/lib/vdsm/storage/qemuimg.py#L218

This comment states that is inefficient: https://github.com/oVirt/vdsm/blob/master/lib/vdsm/storage/fileVolume.py#L557
But I found [benchmarks](https://www.jamescoyle.net/how-to/2060-qcow2-physical-size-with-different-preallocation-settings) where only the `full` setting seems to be really slow. `falloc` is similar to what we do, but (I assume) already taking into account the format and qcow2 header.

nirs: preallocation=falloc is fine on modern storage but harmful on NFS 3, this is why we don't use it and we have our own fallocate helper. Check the commits adding our fallocate helper to understand why.

lib/vdsm/storage/fileVolume.py
```diff
@@ -580,9 +579,7 @@ def _allocate_volume(cls, vol_path, size, preallocate):
                             level=logging.INFO,
                             log=cls.log):
                         op.run()
-        except exception.ActionStopped:
-            raise
-        except Exception as e:
+        except cmdutils.Error as e:
```
nirs: This is little bit more fragile - if code in qemuimg.create() or fallocate() changes and another error will be raised this check will fail, but I think this is better than the current code.

lib/vdsm/storage/fileVolume.py
```diff
@@ -450,14 +450,18 @@ def _create(cls, dom, imgUUID, volUUID, capacity, volFormat, preallocate,
...
+                return cls._create_raw_volume(
+                    dom, volUUID, capacity, volPath, initial_size, preallocate)
+            else:
+                return cls._create_cow_volume(
+                    dom, volUUID, capacity, volPath, initial_size, volParent,
+                    imgUUID, srcImgUUID, srcVolUUID, add_bitmaps=add_bitmaps,
+                    bitmap=bitmap)
+        except cmdutils.Error as e:
+            cls.log.error("Unexpected error: %s", e, exc_info=True)
+            raise se.ImagePathError(volPath) from e
```
nirs, line 464: Looks good after using the right exception for this flow.

lib/vdsm/storage/fileVolume.py
```diff
@@ -583,6 +571,21 @@ def _allocate_volume(cls, vol_path, size, preallocate):
                         log=cls.log):
                     op.run()
 
+    @classmethod
+    def _run_create_image(cls, vol_path, size, format, cow_compat=None):
```
nirs, line 574: In the next patch you create _preallocate_volume helper - for consistency we can name this _create_image.

cow_compat should be qcow2_compat - we want to use the same name for this in the entire project.

lib/vdsm/storage/fileVolume.py
```diff
@@ -528,9 +529,19 @@ def _create_cow_volume(
             cls.log.info("Request to create COW volume %s with capacity = %s",
                          vol_path, capacity)
 
+            vol_format = sc.fmt2str(sc.COW_FORMAT)
```
nirs, line 532: vol_format is usually used for sc.{COW,RAW}_FORMAT. "format" would be a better name.

And we should also use format in the caller to _run_create_image().

nirs (changes requested):

lib/vdsm/storage/fileVolume.py
```diff
@@ -582,9 +582,9 @@ def _allocate_volume(cls, vol_path, size, preallocate):
                         op.run()
         except exception.ActionStopped:
             raise
-        except Exception:
-            cls.log.error("Unexpected error", exc_info=True)
-            raise se.VolumesZeroingError(vol_path)
+        except Exception as e:
+            cls.log.error("Unexpected error: %s", e, exc_info=True)
+            raise se.ImagePathError(vol_path) from e
```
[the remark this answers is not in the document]
aesteve-rh: I see. `se.VolumeCreationError` is even already catched from `volume.create` method (among others), and it sounds accurate. `se.ImagePathError` was not. I will change it.

There are many exceptions with similar names and is not easy to discern which one to use, but in this case I should've seen the exceptions that were already handled within the same context.

nirs (approved):

PR #305, opened 2022-08-29
blockVolume: remove misleading comments

aesteve-rh:
In calculate_volume_alloc_size, there are
a couple comments that hint for the volume
format being qcow.

However, the code that is executed below the
preallocated comment, can be reached
also with a raw volume.

Edit the comment to include preallocated
raw option to avoid confusion.

### DIFF
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -366,7 +366,7 @@ class BlockVolumeManifest(volume.VolumeManifest):
                                               "volume_utilization_chunk_mb")
                 alloc_size = chunk_size_mb * MiB
         else:
-            # Preallocated qcow2
+            # Preallocated raw or qcow2
             alloc_size = initial_size if initial_size else capacity
 
         return alloc_size


lib/vdsm/storage/blockVolume.py
```diff
@@ -354,7 +354,6 @@ def calculate_volume_alloc_size(
                                                    initial_size)
 
         if preallocate == sc.SPARSE_VOL:
-            # Sparse qcow2
```
nirs, line 357: No, this is always qcow2 - there is no raw-sparse on block storage.

lib/vdsm/storage/blockVolume.py
```diff
@@ -366,7 +365,6 @@ def calculate_volume_alloc_size(
                                               "volume_utilization_chunk_mb")
                 alloc_size = chunk_size_mb * MiB
         else:
-            # Preallocated qcow2
```
nirs, line 369: This comment is good, but it is missing the raw-preallocated case. Instead of removing it, we need to list the possible cases:

     Preallocated raw or qcow2

nirs (changes requested):

lib/vdsm/storage/blockVolume.py
aesteve-rh: ok

nirs (approved):

nirs (approved):

Issue #258, opened 2022-06-20
vdsm-tool: move logs from stdout print to logging


aesteve-rh:
Currently, `vdsm-tool` logs messages directly to the stdout using the sys module. However, this solution is very limiting:
- Don't have severity levels, which limit control over what messages are printed, and make the programmer responsible for manually adding the severity in the message.
- Don't have unified formatter. Messages can have inconsistent formats and lack information.
- Messages are always printed from the vdsm-tool modules. For parts where the `vdsm-tool` acts as a wrapper for another vdsm module, any internal logging the vdsm module might do must be deferred to the `vdsm-tool`. Alternatively, with logging we can pass the object as a parameter to allow other modules to log messages directly to the `vdsm-tool` output, with consistent formatting.

### Proposal

Use logging, instantiate it in the `__init__` module and set up a format. From there we can import it in all submodules and use it instead of `sys.stdout`, with the proper severity level, and give it as parameter in wrapper modules.

PR #307, opened 2022-09-01
vdsm-tool: use standard logging

aesteve-rh:
Update logging to have default INFO verbosity.
There is no reason to silence ERROR and WARNING,
and INFO is a good starting point for any CLI.
Therefore, only one verbosity increase is required,
for DEBUG level.

Update logging formatter to beautify and avoid logging
unuseful information.

Change file logging strategy:
- File logging always logs to DEBUG level
- Stdout logging is enabled even if file logging
  is set.

Use the global logger for all vdsm-tool
sub-modules. All logs will use standard
logging with consistent formatting.

Fixes: #258 

### DIFF
--- a/lib/vdsm/tool/__init__.py
+++ b/lib/vdsm/tool/__init__.py
@@ -4,7 +4,37 @@
 from __future__ import absolute_import
 from __future__ import division
 import functools
+import logging
 import os
+
+
+LOGGER_NAME = "vdsm-tool"
+
+
+def setup_logging(log_file, verbosity, append):
+    level = logging.DEBUG if verbosity else logging.INFO
+    root_logger = logging.getLogger(LOGGER_NAME)
+    root_logger.setLevel(level)
+
+    if root_logger.handlers:
+        return root_logger
+
+    if log_file is not None:
+        mode = append and 'a' or 'w'
+        file_handler = logging.FileHandler(log_file, mode=mode)
+        file_handler.setFormatter(
+            logging.Formatter(
+                "%(asctime)s [%(levelname)-7s] %(message)s",
+                "%m/%d/%Y %I:%M:%S %p"))
+        # File logging always at DEBUG level
+        file_handler.setLevel(logging.DEBUG)
+        root_logger.addHandler(file_handler)
+
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter("%(levelname)-7s | %(message)s"))
+    root_logger.addHandler(handler)
+
+    return root_logger
 
 
 class expose(object):
--- a/lib/vdsm/tool/config_lvm_filter.py
+++ b/lib/vdsm/tool/config_lvm_filter.py
@@ -6,6 +6,7 @@ from __future__ import division
 from __future__ import print_function
 
 import argparse
+import logging
 import textwrap
 
 from vdsm.common.config import config
@@ -14,6 +15,7 @@ from vdsm.storage import lvmdevices
 from vdsm.storage import lvmfilter
 from vdsm.storage import mpathconf
 
+from . import LOGGER_NAME
 from . import expose
 from . import common
 
@@ -25,6 +27,8 @@ _NAME = 'config-lvm-filter'
 # rc=2 will be exited from parse_args() in case of invalid usage.
 CANNOT_CONFIG = 3
 NEEDS_CONFIG = 4
+
+log = logging.getLogger(LOGGER_NAME)
 
 
 @expose(_NAME)
@@ -46,7 +50,7 @@ def main(*args):
     """
     args = parse_args(args)
 
-    print("Analyzing host...")
+    log.info("Analyzing host...")
 
     config_method = config.get("lvm", "config_method").lower()
 
@@ -55,8 +59,8 @@ def main(*args):
     elif config_method == "devices":
         return config_with_devices(args)
     else:
-        print("Unknown configuration method %s, use either 'filter' or "
-              "'devices'." % config_method)
+        log.info("Unknown configuration method %s, use either 'filter' or "
+                 "'devices'.", config_method)
         return CANNOT_CONFIG
 
 
@@ -76,7 +80,7 @@ def config_with_filter(args):
 
     # This is the expected condition on a correctly configured host.
     if advice.action == lvmfilter.UNNEEDED:
-        print("LVM filter is already configured for Vdsm")
+        log.info("LVM filter is already configured for Vdsm")
         return
 
     # We need to configure LVM filter.
@@ -107,7 +111,7 @@ def config_with_devices(args):
 
     # Check if the lvm devices is already configured.
     if lvmdevices.is_configured():
-        print("LVM devices already configured for vdsm")
+        log.info("LVM devices already configured for vdsm")
         return
 
     mounts = lvmfilter.find_lvm_mounts()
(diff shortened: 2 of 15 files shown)


aesteve-rh:
## Examples

Before:
```
# vdsm-tool is-configured
Managed volume database is already configured
Current revision of multipath.conf detected, preserving
lvm is configured for vdsm
sanlock is configured for vdsm
libvirt is already configured for vdsm
```

After:
```
# vdsm-tool is-configured
INFO    | lvm is configured for vdsm
INFO    | libvirt is already configured for vdsm
INFO    | Managed volume database is already configured
INFO    | Current revision of multipath.conf detected, preserving
INFO    | sanlock is configured for vdsm
```

Before:

```
# vdsm-tool is-configured --module multipath
Current revision of multipath.conf detected, preserving
WARNING: Invalid configuration: 'user_friendly_names' is enabled in multipath configuration:
   overrides {
     no_path_retry 16
     user_friendly_names yes
   }
This configuration is not supported and may lead to storage domain corruption.
```

After:
```
# vdsm-tool is-configured --module multipath
INFO    | Current revision of multipath.conf detected, preserving
WARNING | Invalid configuration: 'user_friendly_names' is enabled in multipath configuration:
WARNING |   overrides {
WARNING |     no_path_retry 16
WARNING |     user_friendly_names yes
WARNING |   }
WARNING | This configuration is not supported and may lead to storage domain corruption.
```

aesteve-rh:
```
# vdsm-tool check-volume-leases
WARNING | Make sure there are no running storage operations.

Do you want to check volume leases? [yes,NO] yes
INFO    | Checking active storage domains. This can take several minutes, please wait.
ERROR   | The storage pool is not connected. Please make sure the host is active before running the tool.
```

lib/vdsm/tool/__init__.py
```diff
@@ -20,9 +20,39 @@
...
 
+LOGGER_NAME = "vdsm-tool"
+
+
+def setup_logging(log_file, verbosity, append):
+    level = logging.DEBUG if verbosity else logging.INFO
+    root_logger = logging.getLogger(LOGGER_NAME)
+    root_logger.setLevel(level)
+
+    if root_logger.handlers:
```
mz-pdm, line 35: Ignoring log_file?

lib/vdsm/tool/__init__.py
```diff
@@ -20,9 +20,39 @@
...
+        return root_logger
+
+    if log_file is not None:
+        mode = append and 'a' or 'w'
+        file_handler = logging.FileHandler(log_file, mode=mode)
+        file_handler.setFormatter(
+            logging.Formatter(
+                "%(asctime)s [%(levelname)-7s] %(message)s",
+                "%m/%d/%Y %I:%M:%S %p"))
+        # File logging always at DEBUG level
```
mz-pdm, line 45: Why?

lib/vdsm/tool/__init__.py
```diff
@@ -20,9 +20,39 @@
...
+        file_handler.setLevel(logging.DEBUG)
+        root_logger.addHandler(file_handler)
+
+    handler = logging.StreamHandler()
+    handler.setFormatter(logging.Formatter("%(levelname)-7s | %(message)s"))
+    root_logger.addHandler(handler)
```
mz-pdm, line 51: Why is stdout logging enabled now when file logging is enabled?

lib/vdsm/tool/config_lvm_filter.py
```diff
@@ -163,74 +167,65 @@ def parse_args(args):
...
-    print("""\
-This filter allows LVM to access the local devices used by the
-hypervisor, but not shared storage owned by Vdsm. If you add a new
-device to the volume group, you will need to edit the filter manually.
-    """)
+    log.info("This filter allows LVM to access the local devices used by the")
+    log.info("hypervisor, but not shared storage owned by Vdsm. "
+             "If you add a new")
+    log.info("device to the volume group, "
+             "you will need to edit the filter manually.")
```
mz-pdm, line 187: Why to split it (here and elsewhere) to multiple logging messages?  For visual effects?

mz-pdm (commented):
The cleanup (verbosity, using loggers) is good.  I'm not sure about the formatting changes -- I find the original format better (human) readable and I don't know whether anything in deployment may rely on the format.

lib/vdsm/tool/config_lvm_filter.py
aesteve-rh: Multiline strings with logging look pretty bad, and it is consensus to try to avoid them. They look misaligned and makes it harder to parse.

In this example, if we keep it in a single string, it would look like:

```
INFO    | This filter allows LVM to access the local devices used by the
hypervisor, but not shared storage owned by Vdsm. If you add a new
device to the volume group, you will need to edit the filter manually.
```

lib/vdsm/tool/__init__.py
aesteve-rh: This is a save-guard in case it gets configured twice, not to add extra handlers.

lib/vdsm/tool/__init__.py
aesteve-rh: File logging is often used for reports and bug hunting, so default `DEBUG` level seemed appropriate. 

Taking into account that previously most messages were printed directly to stdout and not through logging, file logs were probably pretty empty. Now it will print all logs, which can be very useful in, e.g., an error occurred in the update-tool.

lib/vdsm/tool/__init__.py
aesteve-rh: I can see how this may be a questionable change.

The reasoning in my head was that as a user, probably is nice to keep the logs in the console even with file logging enabled. File logging will provide more detailed logs, to share and parse, but if something goes wrong we will still get the messages in the console without having to open the log file.

Again, I can see also how this may be questionable, depending on how/when file logging is used, so I can revert it.

lib/vdsm/tool/__init__.py
mz-pdm: I see, OK.  It looked like a case when an external configuration is used (e.g. in other places in Vdsm), but it's not the case here.

lib/vdsm/tool/__init__.py
mz-pdm: OK, thanks for clarification.  But again, are we sure nothing relies on the current content of the file?

Please add the reasoning behind this change to the commit message.

lib/vdsm/tool/__init__.py
mz-pdm: OK, makes sense, as long as it doesn't break anything that might be looking at console errors when file logging is enabled.

Please add the reasoning behind this change to the commit message.

lib/vdsm/tool/__init__.py
aesteve-rh: `are we sure nothing relies on the current content of the file?` No, that I am not sure, maybe @tinez knows?

I assume not, since, as I said, log file was probably pretty empty before. But there was some logging in network commands, so it could be that those were actually parsed.

aesteve-rh:
Regarding the format, we can have a custom Formatter to have different formats for different log levels.

That would allow having INFO messages to print only the messages (no format nor timestamp), thus achieving the same format as before. Other levels will keep the format in this version, with: `LEVEL    | msg`.

That may be a good idea also to have some messages that I have kept with print (as in the `show-default-config` file) logged, and visible from the log file.

mz-pdm:
Yes, OK, let's see what the others think.

aesteve-rh:
Didn't get attention in a while and is not relevant anymore. Let's keep things as they are.

PR #308, opened 2022-09-05
Remove legacy automation

aesteve-rh:
Rename automation folder to ci and update all
references to automation scripts.

Remove unused ci (old automation) scripts.

Fixes: https://github.com/oVirt/vdsm/issues/43


nirs (changes requested):
Why rename tons of files in automation/ and delete them in the next commit? It would be better to delete the unused file first, then rename the few used files.

doc/development.md
```diff
@@ -26,7 +26,7 @@ Clone your fork:
 
 Install additional packages for Fedora, CentOS, and RHEL:
 
-    sudo dnf install -y `cat ci/check-patch.packages`
+    ci/install-pkg.sh
```
nirs, line 29: This is not used in the ci - it is used only for preparing development environment manually, so why does it belong in the ci directory?

aesteve-rh: right, moved to `contrib`, it should pertain there better.

mz-pdm (approved):

doc/development.md
```diff
@@ -26,7 +26,7 @@ Clone your fork:
 
 Install additional packages for Fedora, CentOS, and RHEL:
 
-    sudo dnf install -y `cat ci/check-patch.packages`
+    contrib/install-pkg.sh
```
nirs, line 29: contrib/install-pkg.sh is not the best name, and the script does have any documentation so someone looking in this script will have no idea what is the purpose of it.

nirs (approved):

doc/development.md
aesteve-rh: Good idea: https://github.com/oVirt/vdsm/issues/309

I can tackle this in a followup PR so that we unblock the SPDX branch.

Issue #55, opened 2022-01-30
Remove vdsm.common.compat module


nirs:
This module was added to handle differences between python versions:
- simplejson on RHEL 6.x
- subprocess32 on python 2
- missing glob.escape on python 2
- using inspect.getargspec on python 2
- handling missing modules on python 3

All these hacks are not needed for a while, so the module can be removed.

Dependnecies:
- vdsm.storage.compat depending on vdsm.common.compat.MissingModule.

Depends: #56

PR #311, opened 2022-09-08
Remove common.compat module

aesteve-rh:
This module was added to handle differences between python versions:

- simplejson on RHEL 6.x
- subprocess32 on python 2
- using inspect.getargspec on python 2
- handling missing modules on python 3

All these hacks are not needed for a while, so the module can be removed.

Fixes: #55


mz-pdm (approved):
Thanks for the cleanup!

tests/storage/backends_test.py
```diff
@@ -30,10 +30,11 @@
 from vdsm.storage import localFsSD
 from vdsm.storage import nfsSD
 
-from . marks import requires_sanlock
+sanlock = pytest.importorskip(
+    modname='sanlock',
+    reason="sanlock is not available")
```
nirs, line 35: Not really needed, sanlock is always available and required. Missing sanlock is
leftover from the porting to python 3.

nirs (approved):
We can simplify sanlock import skip, but lets first merge this awesome cleanup.

PR #316, opened 2022-09-16
merge_test: increase reduce vol size

aesteve-rh:
Increase the size of the volumes created in the
reduce tests in the merge_test module, so that
volume sizes are more realistic.

Consequently, chunked leaf volumes do not reduce
anymore in the test. Chunked leaf volumes add an
extra chunk to the required size as the condition
to call reduce.

The test relied in the allocation
that Vdsm does when the volume is created
in `calculate_volume_alloc_size` if the create
call does not specify an initial size.
In that case, Vdsm allocates an entire chunk,
whereas the requested capacity was smaller (test
requested 2 GiB volumes, Vdsm allocated 2.5 GiB).

As long as the required size is smaller
than the difference between capacity and
chunk size (0.5 GiB), reduce will be called
during finalizeMerge.

In a real setup, tests show that leaf volumes
are not reduced during finalizeMerge except if
they have been extended previously. Thus, increasing 
the size of the volumes and changing the expectation
in the test matches real scenario.

### DIFF
--- a/tests/storage/merge_test.py
+++ b/tests/storage/merge_test.py
@@ -368,7 +368,7 @@ class TestFinalizeMerge:
             format='raw',
             prealloc=sc.SPARSE_VOL,
             chain_len=2):
-        size = 2 * GiB
+        size = 10 * GiB
         base_fmt = sc.name2type(format)
         with fake_env(sd_type) as env:
             with MonkeyPatch().context() as mp:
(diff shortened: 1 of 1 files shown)


tests/storage/merge_test.py
```diff
@@ -539,10 +539,13 @@ def test_reduce_chunked_leaf(self):
...
-            assert fake_base_vol.__calls__ == [
-                ('reduce', (base_vol.optimal_size(as_leaf=True),), {}),
-            ]
+            # Chunked leaf volumes add one chunk of free space as condition
+            # to reduce, so the VM will not pause quickly when it is started.
+            # Consequently, cold merges will not call reduce.
+            calls = getattr(fake_base_vol, "__calls__", {})
+            # Verify that 'calls' is empty which means that 'reduce' wasn't
+            # called
+            assert len(calls) == 0
```
nirs, line 548: A simpler way would be:

    assert fake_base_vol.__calls__ == []
    
And if the list is set only on the first call:

    assert getattr(fake_base_vol, “__calls__”, []) == []

aesteve-rh: The list is only set on first call, yes.

I opted for this solution for consistency with the check in the next test. Also, `__calls__` (if set) would be a dict, so the current solution does not re-type the member. But I'm ok with your proposed change.

PR #332, opened 2022-09-30
tests: Test builtin lvmlocal.conf

nirs:
Recent change adding spdx headers broke OST. Add simple test ensuring that the builtin file can be parsed.

### DIFF
--- a/tests/tool_configurators_lvm_test.py
+++ b/tests/tool_configurators_lvm_test.py
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Red Hat, Inc.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import os
+
+from vdsm.tool import confmeta
+
+TOPDIR = os.path.dirname(os.path.dirname(__file__))
+
+
+def test_lvmlocal_conf():
+    lvmlocal_conf = os.path.join(
+        TOPDIR, "static/usr/share/vdsm/lvmlocal.conf")
+    md = confmeta.read_metadata(lvmlocal_conf)
+    assert md.revision >= 7
+    assert md.private is False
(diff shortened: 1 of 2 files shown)


tests/tool_configurators_lvm_test.py
```diff
@@ -0,0 +1,16 @@
...
+import os
+
+from vdsm.tool import confmeta
+
+TOPDIR = os.path.dirname(os.path.dirname(__file__))
+
+
+def test_lvmlocal_conf():
+    lvmlocal_conf = os.path.join(
+        TOPDIR, "static", "usr", "share", "vdsm", "lvmlocal.conf")
```
aesteve-rh, line 13: This works and is good, but I would consider having the relative part of the path in a single string to make it a bit easier to read:
```python
    lvmlocal_conf = os.path.join(TOPDIR, "static/usr/share/vdsm/lvmlocal.conf")
```
Maybe even pre-assigned to a local or global variable.

But there is not much difference, so feel free to ignore it.

aesteve-rh (approved):
Great! We discovered a verification gap, and you covered it right away with a simple test.

The tool would need some more coverage in general, including other configurators. This module is a nice placeholder for that, too.

nirs:
> > Recent change adding spdx headers broke OST.
> 
> Really, or did you get confused by my PR #331? OST eventually passed on master for me although there had been previous repeated failure related LVM in [oVirt/ovirt-system-tests#270](https://github.com/oVirt/ovirt-system-tests/pull/270) .
> 
> Regardless, adding this test is good of course.

Yes, we added empty line before lvmconf header and this breaks the code parsing 
metadata from the header. This breaks "vdsm-tool is-configured" since the file
will never have the expected metadata.

aesteve-rh (approved):

mz-pdm (approved):

mz-pdm:
> @mz-pdm please ack and merge, I cannot merge now without another review.

@michalskrivanek , requiring two maintainers to merge a PR is impractical. Would you change the setting in Branch protection rules, please?

PR #340, opened 2022-10-10
Fix missing informations on geo replication tab of GlusterFS volumes.

yasalos:
Since the upgrade to oVirt 4.5 and GlusterFS 10, information about snapshots and geo replication are no more visible on the engine Webui. After digging VDSM logs, I found that it's caused by the change of some XML elements name introduced in a new Gluster release (https://github.com/gluster/glusterfs/pull/1568) and by the addition of the timezone in output of status command.

This commit fix the parsing process of the output returned by gluster command, so that data sent to ovirt engine are as expected.

Bug-URL: https://bugzilla.redhat.com/show_bug.cgi?id=2104278

### DIFF
--- a/lib/vdsm/gluster/cli.py
+++ b/lib/vdsm/gluster/cli.py
@@ -1125,7 +1125,7 @@ def _parseGeoRepStatus(tree):
                     bricks: [{host: 'local node',
                               hostUuid: 'uuid of brick host',
                               brickName: 'brick in the local volume',
-                              remoteHost: 'slave',
+                              remoteHost: 'secondary',
                               status: 'status'
                               remoteUserName: 'root'
                               timeZone: 'nodes time zone'
@@ -1142,6 +1142,8 @@ def _parseGeoRepStatus(tree):
                ]....
     }
     """
+    gluster_old_format = tree.find(
+        'geoRep/volume/sessions/session/session_secondary') is None
     status = {}
     for volume in tree.findall('geoRep/volume'):
         sessions = []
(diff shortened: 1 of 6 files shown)


mz-pdm (changes requested):
Thank you for the fix!  In order not to break hosts running with older GlusterFS versions, it should either support both the old and new formats or a dependency on the new GlusterFS version should be added to the .spec.in file (which would probably break el8 environments, so supporting both the formats seems to be preferable).

yasalos:
OK, i'll update the code so that both version are supported.

yasalos:
I updated the code to make sure that both (old and new) format of XML output are supported.

lib/vdsm/gluster/cli.py
```diff
@@ -1149,17 +1149,29 @@ def _parseGeoRepStatus(tree):
         for session in volume.findall('sessions/session'):
             pairs = []
             sessionDetail = {}
-            sessionDetail['sessionKey'] = session.find('session_slave').text
+            if session.find('session_secondary') is None:
```
mz-pdm, line 1152: Maybe it would be a good idea to look up the format only once, outside the nested loops, and store it into a variable used here and below to select the proper format.

mz-pdm (commented):
Looks good to me, just one suggestion for a possible improvement inside the code.

A technical note: Please squash the two commits -- we don't squash commits when merging a PR, commits are kept separate because they may implement different pieces.  For this reason, we use force (replacement) commits rather than amendment commits when modifying a PR during the review process.

lib/vdsm/gluster/cli.py
yasalos: I tought so at the begining, but i'm still learning Python language. In fact, this is my first real projet using Python and I dont know API and functions very well yet. However I learnt about XML parsing and updated the code to identify the version before entering the loop.

mz-pdm: Cool :-), one more thing to fix and then the patch should be ready.

tests/glusterGeoRepStatusOld.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
mz-pdm, line 1: Please add glusterGeoRepStatusOld.xml.license file (similar to the *.license files already present) to make the linter happy.

mz-pdm (changes requested):
Please rebase on master and make sure the PR contains only a single commit (no merge commit).

lib/vdsm/gluster/cli.py
```diff
@@ -1142,24 +1142,42 @@ def _parseGeoRepStatus(tree):
                ]....
     }
     """
+    if tree.find('geoRep/volume/sessions/session/session_secondary') is None:
+        # Gluster version < 10
+        gluster_format = 9
+    else:
+        # Gluster version >= 10
+        gluster_format = 10
```
aesteve-rh, line 1150: Take this comment as a suggestion.

This works, but it is a boolean in the end, so it could be:
```python
gluster_old_format = tree.find('geoRep/volume/sessions/session/session_secondary') is None
```
If we need to support more formats in the future, we could use global variables for them. But boolean suffices for now. And then below we just need:
```python
if gluster_old_format:
     ...
else:
     ...
```
`elifs` check the condition again, which is not necessary for now.

tests/gluster_cli_test.py
```diff
@@ -1287,6 +1287,14 @@ def test_parseGeoRepStatus(self):
         status = gcli._parseGeoRepStatus(tree)
         self.assertEqual(status, glusterTestData.GLUSTER_GEOREP_STATUS)
 
+    def test_parseGeoRepStatusOld(self):
+        with open("glusterGeoRepStatusOld.xml") as f:
+            out = f.read()
+        tree = etree.fromstring(out)
+        gcli._TIME_ZONE = 'IST'
+        status = gcli._parseGeoRepStatus(tree)
+        self.assertEqual(status, glusterTestData.GLUSTER_GEOREP_STATUS_OLD)
```
aesteve-rh, line 1296: This is the same as the `test_parseGeoRepStatus`, except for the filename and the expected data, which could be parametrized and remove the extra test.

However, the test code is simple enough, so we can leave it as is.

aesteve-rh (dismissed):
Looks good to me, after having addressed the comments from @mz-pdm.

A couple of minor comments from my side.

tests/glusterGeoRepStatusOld.xml
yasalos: Done.

lib/vdsm/gluster/cli.py
yasalos: Done.

mz-pdm:
Some more errors from the linter (see the CI failures -- too long lines).

yasalos:
> Some more errors from the linter (see the CI failures -- too long lines).

Fixed

mz-pdm:
Great, CI and I are happy now. Let's see whether OST is happy too, as well as Albert about the formatting :-).

aesteve-rh (approved):
lgtm!

mz-pdm (approved):

mz-pdm:
Although not reported here for whatever reason, OST actually run and succeeded.

mz-pdm:
Congratulations to completing your first contribution!

yasalos:
> Congratulations to completing your first contribution!

It was not just my first contribution to oVirt, but to any project. I want to thank, both Albert and you for leading the way and helping me during this process. Now, I have to go and read the new contributors guidelines section in development.md file ;-)

PR #341, opened 2022-10-12
Remove SPDX headers from sensitive files

mz-pdm:
The recently added SPDX headers cause some problems with downstream builds:

- SPDX headers in sudoers files require security review.

- SPDX headers in autounattended files report linting failure because the files don’t start with an XML header at the very beginning.

Let’s remove SPDX headers from the given places to avoid false alerts. Let’s also disable ‘reuse’ check in ‘make lint’ to avoid linter failures due to missing SPDX headers.

This is 4.5.3-only change to avoid problems with downstream builds.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -125,7 +125,7 @@ check-unit:
 	$(MAKE) -C tests/ check NOSE_EVAL_ATTR="type=='unit'"
 
 .PHONY: lint
-lint: gitignore reuse execcmd black flake8 pylint
+lint: gitignore execcmd black flake8 pylint
 
 .PHONY: venv
 venv:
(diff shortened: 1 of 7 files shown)


mwperina (approved):
+1

aesteve-rh:
Rather than disabling the reuse linter, we should've used explicit licenses (i.e., create files with `.license` extension, which allows keeping files without SPDX header, and still makes the linter happy).

```
$ contrib/add-spdx-header.sh --explicit-license \
    static/etc/sudoers.d/50_vdsm.in \
    static/usr/share/vdsm/autounattend/Autounattend.xml.in \
    vdsm_hooks/extra_ipv4_addrs/sudoers \
    vdsm_hooks/localdisk/sudoers.vdsm_hook_localdisk \
    vdsm_hooks/openstacknet/sudoers.in \
    vdsm_hooks/vhostmd/sudoers.vdsm_hook_vhostmd
```

mz-pdm:
This fix/merge was quick because we needed it to proceed with the build.  It might be worth to backport the XML fix from master and add the explicit licenses as you suggest for the sudoers files.  I'll post a PR for that.

PR #343, opened 2022-10-12
reuse: Put SPDX after XML declaration

mz-pdm:
A valid XML should start with the declaration and only then comments are allowed.


mwperina (approved):
+1

aesteve-rh:
Thanks for the fix! I will make an issue in the reuse forums with this one, as the reuse tool autofills it wrongly.

PR #346, opened 2022-10-25
lvmdevices: create devices directory

aesteve-rh:
Current implementation assumes /etc/lvm/devices folder exists, which may not be true with a fresh installation (only in an upgrade).

We need to make sure that we create the directory
before creating the empty devices file, with
exist_ok flag set to True. This will ensure
that it works in both scenarios.

Fixes: b8692e93f38abaed973c528fe457a767cc1245ea


aesteve-rh:
@didib fyi

nirs (approved):

Issue #352, opened 2022-11-16
Live merge sometimes fails, "No space left on device" appear in the log


ahadas:
Description of problem: sometime snapshot removal fails, leaving VM with disk(s) in illegal state; subsequent tries to remove snapshot fail again

Version-Release number of selected component (if applicable): ovirt-engine-4.5.2.4-1.el8.noarch, oVirt Node 4.5.2 (vdsm-4.50.2.2-1.el8)

How reproducible: not always; issue happens now and then when backup system (storware vprotect, interacting with oVirt via API) tries to delete snapshots

Actual results: snapshot removals fail, and subsequent tries fail again 

Expected results: snapshot is correctly removed

Additional info: issue doesn't happen every time, VMs are always (almost) the same (5 or 6 out of about 40), with no apparent pattern (VM either linux or windows, big or small disks, all hypervisors, ...)

Original bug: https://bugzilla.redhat.com/2122525

Tasks:
- [x] Add bitmaps.prune_bitmaps() (#354)
- [x] Prune bitmaps before cold merge (#354)
- [ ] Prune bitmaps before live merge (#357)

PR #354, opened 2022-11-17
 cold merge: prune stale bitmaps in base volume

aesteve-rh:
Prune the stale base volume bitmaps during
the prepare step on a merge operation.

These stale bitmaps can cause the merge
operation to fail due to 'No space left on device'.
In this case, qemu does not end with error, so
the failure goes unnoticed.

As there is not a reliable way to measure
the size of these stale bitmaps, and they are
invalid and can never be used for incremental
backup, it is better to prune them to avoid
the error.

Related: https://github.com/oVirt/vdsm/issues/352

### DIFF
--- a/lib/vdsm/storage/bitmaps.py
+++ b/lib/vdsm/storage/bitmaps.py
@@ -125,6 +125,27 @@ def remove_bitmap(vol_path, bitmap):
             vol_path=vol_path)
 
 
+def prune_bitmaps(base_path, top_path):
+    """
+    Prune all the stale bitmaps from the base volume path.
+    A bitmap is considered stale if it appears only in the base volume,
+    and is missing or invalid in the top volume.
+
+    Args:
+        base_path (str): Path to the base volume
+        top_path (str): Path to the top volume
+    """
+    base_bitmaps = _query_bitmaps(base_path)
+    valid_top_bitmaps = _query_bitmaps(top_path, filter=_valid)
+
+    stale_bitmaps = [
+        name for name in base_bitmaps if name not in valid_top_bitmaps]
+    if stale_bitmaps:
+        log.warning("Prune stale bitmaps %s from %r", stale_bitmaps, base_path)
+        for bitmap in stale_bitmaps:
+            remove_bitmap(base_path, bitmap)
+
+
 def clear_bitmaps(vol_path):
     """
     Remove all the bitmaps from the given volume path
--- a/lib/vdsm/storage/merge.py
+++ b/lib/vdsm/storage/merge.py
@@ -22,6 +22,7 @@ import logging
 from vdsm.common import properties
 from vdsm.common.units import MiB
 
+from vdsm.storage import bitmaps
 from vdsm.storage import constants as sc
 from vdsm.storage import exception as se
 from vdsm.storage import guarded
@@ -158,6 +159,15 @@ def prepare(subchain):
     log.info("Preparing subchain %s for merge", subchain)
     with guarded.context(subchain.locks):
         with subchain.prepare():
+            # There may be stale bitmaps in base if they were left
+            # after failed delete bitmap operation. These bitmaps may
+            # cause the bitmaps merge operation to fail with ENOSPC.
+            # As there is not a reliable way to calculate the size of
+            # these bitmaps, and they are invalid and can never be used
+            # for incremental backup, we prune them before the commit.
+            bitmaps.prune_bitmaps(subchain.base_vol.getVolumePath(),
+                                  subchain.top_vol.getVolumePath())
+
             _update_base_capacity(subchain.base_vol,
                                   subchain.top_vol)
             _extend_base_allocation(subchain.base_vol,
--- a/tests/storage/bitmaps_test.py
+++ b/tests/storage/bitmaps_test.py
@@ -7,6 +7,8 @@ import pytest
 
 from collections import namedtuple
 
+from . import qemuio
+
 from vdsm.common import cmdutils
 from vdsm.common import exception
 from vdsm.common.units import MiB
@@ -232,6 +234,78 @@ def test_remove_bitmap_failed(monkeypatch, tmp_mount, vol_chain):
 def test_remove_non_existing_bitmap_succeed(tmp_mount, vol_chain):
     # try to remove a non-existing bitmap from top_vol
     bitmaps.remove_bitmap(vol_chain.top_vol, 'bitmap')
+
+
+def test_prune_stale_bitmaps(tmp_mount, vol_chain):
+    # Add valid bitmap to volumes
+    bitmap = 'valid-bitmap'
+    qemuimg.bitmap_add(vol_chain.base_vol, bitmap).run()
+    qemuimg.bitmap_add(vol_chain.top_vol, bitmap).run()
+
+    # Add stale bitmaps to base volume
+    for i in range(5):
+        qemuimg.bitmap_add(vol_chain.top_vol, f"stale-bitmap-{i}").run()
+
+    bitmaps.prune_bitmaps(vol_chain.base_vol, vol_chain.top_vol)
+
+    info = qemuimg.info(vol_chain.base_vol)
+    assert info['format-specific']['data']['bitmaps'] == [
+        {
+            "flags": ['auto'],
+            "name": bitmap,
+            "granularity": 65536
+        },
+    ]
+
+
+def test_prune_disabled_bitmaps(tmp_mount, vol_chain):
+    # Add valid bitmap to volumes
+    bitmap = 'valid-bitmap'
+    qemuimg.bitmap_add(vol_chain.base_vol, bitmap).run()
+    qemuimg.bitmap_add(vol_chain.top_vol, bitmap).run()
+
+    # Add disabled bitmaps to top volume
+    for i in range(5):
+        bitmap_disabled = f"disabled-bitmap-{i}"
+        qemuimg.bitmap_add(vol_chain.base_vol, bitmap_disabled).run()
+        qemuimg.bitmap_add(
+            vol_chain.top_vol, bitmap_disabled, enable=False).run()
+
+    bitmaps.prune_bitmaps(vol_chain.base_vol, vol_chain.top_vol)
+
+    info = qemuimg.info(vol_chain.base_vol)
+    assert info['format-specific']['data']['bitmaps'] == [
+        {
+            "flags": ['auto'],
+            "name": bitmap,
+            "granularity": 65536
+        },
+    ]
+
+
+def test_prune_in_use_bitmaps(tmp_mount, vol_chain):
+    # Add inconsistent "in-use" bitmaps to volumes
+    for i in range(5):
+        bitmap_in_use = f"in-use-bitmap-{i}"
+        qemuimg.bitmap_add(vol_chain.base_vol, bitmap_in_use).run()
+        qemuimg.bitmap_add(vol_chain.top_vol, bitmap_in_use).run()
+    qemuio.abort(vol_chain.top_vol)
+
+    # Add valid bitmap to volumes
+    bitmap = 'valid-bitmap'
+    qemuimg.bitmap_add(vol_chain.base_vol, bitmap).run()
+    qemuimg.bitmap_add(vol_chain.top_vol, bitmap).run()
+
+    bitmaps.prune_bitmaps(vol_chain.base_vol, vol_chain.top_vol)
+
+    info = qemuimg.info(vol_chain.base_vol)
+    assert info['format-specific']['data']['bitmaps'] == [
+        {
+            "flags": ['auto'],
+            "name": bitmap,
+            "granularity": 65536
+        },
+    ]
 
 
 def test_clear_bitmaps(tmp_mount, vol_chain):
--- a/tests/storage/blocksd_test.py
+++ b/tests/storage/blocksd_test.py
@@ -13,6 +13,7 @@ import string
 
 import pytest
 
+from vdsm import jobs
 from vdsm import utils
 from vdsm.config import config
 from vdsm.common.units import MiB, GiB
@@ -20,22 +21,28 @@ from vdsm.storage import blockSD
 from vdsm.storage import clusterlock
 from vdsm.storage import constants as sc
 from vdsm.storage import exception as se
+from vdsm.storage import guarded
 from vdsm.storage import image
 from vdsm.storage import lvm
 from vdsm.storage import qemuimg
+from vdsm.storage import merge
 from vdsm.storage import sanlock_direct
 from vdsm.storage import sd
 from vdsm.storage import sp
 from vdsm.storage.sdc import sdCache
+from vdsm.storage.sdm.api import merge as api_merge
 from vdsm.storage.spbackends import StoragePoolDiskBackend
 
 from . import qemuio
 from . marks import requires_root
 
+from fakelib import FakeNotifier
+from fakelib import FakeScheduler
 from storage.storagefakelib import fake_spm
 from storage.storagefakelib import fake_vg
 from storage.storagefakelib import FakeDomainMonitor
 from storage.storagefakelib import FakeTaskManager
+from storage.storagefakelib import fake_guarded_context
 
 TESTDIR = os.path.dirname(__file__)
 CHUNK_SIZE_MB = config.getint("irs", "volume_utilization_chunk_mb")
@@ -1242,6 +1249,125 @@ def test_create_volume_with_new_bitmap(
             "granularity": 65536
         },
     ]
+
+
+@requires_root
+@pytest.mark.root
+@pytest.mark.parametrize("stale_bitmaps", [
+    6,   # required size < lv size
+    11,  # required size = lv size
+    15,  # required size > lv size
+])
+@pytest.mark.xfail(reason='Stale bitmaps in base volume not considered')
+def test_merge_with_bitmap(
+        tmp_storage, tmp_repo, fake_access, fake_rescan, tmp_db, fake_task,
+        fake_sanlock, monkeypatch, stale_bitmaps, caplog):
+    sd_id = str(uuid.uuid4())
+
+    dev = tmp_storage.create_device(10 * GiB)
+    lvm.createVG(sd_id, [dev], blockSD.STORAGE_UNREADY_DOMAIN_TAG, 128)
+    vg = lvm.getVG(sd_id)
+
+    dom = blockSD.BlockStorageDomain.create(
+        sdUUID=sd_id,
+        domainName="domain",
+        domClass=sd.DATA_DOMAIN,
+        vgUUID=vg.uuid,
+        version=4,
+        storageType=sd.ISCSI_DOMAIN)
+
+    sdCache.knownSDs[sd_id] = blockSD.findDomain
+    sdCache.manuallyAddDomain(dom)
+
+    dom.refresh()
+    dom.attach(tmp_repo.pool_id)
+
+    img_id = str(uuid.uuid4())
+    base_vol_uuid = str(uuid.uuid4())
+    top_vol_uuid = str(uuid.uuid4())
+
+    # Create base volume.
+    dom.createVolume(
+        imgUUID=img_id,
+        capacity=128 * MiB,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType='DATA',
+        volUUID=base_vol_uuid,
+        desc="Test volume",
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID)
+
+    base_vol = dom.produceVolume(img_id, base_vol_uuid)
+
+    # Create top volume with bitmaps.
+    dom.createVolume(
+        imgUUID=img_id,
+        capacity=128 * MiB,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType='DATA',
+        volUUID=top_vol_uuid,
+        desc="Test top volume",
+        srcImgUUID=base_vol.imgUUID,
+        srcVolUUID=base_vol.volUUID)
+
+    top_vol = dom.produceVolume(img_id, top_vol_uuid)
+
+    # Create leaf volume.
+    dom.createVolume(
+        imgUUID=img_id,
+        capacity=128 * MiB,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType='DATA',
+        volUUID=str(uuid.uuid4()),
+        desc="Test top volume",
+        srcImgUUID=top_vol.imgUUID,
+        srcVolUUID=top_vol.volUUID)
+
+    host_id = 0
+    subchain_info = dict(sd_id=base_vol.sdUUID,
+                         img_id=base_vol.imgUUID,
+                         base_id=base_vol.volUUID,
+                         top_id=top_vol.volUUID,
+                         base_generation=0)
+    subchain = merge.SubchainInfo(subchain_info, host_id)
+
+    # Nearly fill the volume with data.
+    top_vol.prepare()
+    qemuio.write_pattern(
+        top_vol.getVolumePath(),
+        qemuimg.FORMAT.QCOW2,
+        offset=0,
+        len=126 * MiB)
+    # Add good bitmap to top.
+    qemuimg.bitmap_add(base_vol.getVolumePath(), "good-bitmap").run()
+    top_vol.teardown(sd_id, top_vol_uuid)
+
+    # Add bitmaps to the base volume after creating top volume.
+    base_vol.prepare()
+    for i in range(stale_bitmaps):
+        qemuimg.bitmap_add(base_vol.getVolumePath(), f"stale-bitmap{i}").run()
+    base_vol.teardown(sd_id, base_vol_uuid)
+
+    with monkeypatch.context() as mc:
+        # TODO: check why this context fake is required. Otherwise, we need to
+        # acquire the host id, which causes other tests to fail.
+        mc.setattr(guarded, 'context', fake_guarded_context())
+        # Prepare for merge, should expand the base volume.
+        merge.prepare(subchain)
+        # Run the actual merge.
+        jobs.start(FakeScheduler(), FakeNotifier())
+        caplog.clear()
+        job = api_merge.Job(str(uuid.uuid4()), subchain, merge_bitmaps=True)
+        job.run()
+        merge.finalize(subchain)
+
+    # Merge ended correctly.
+    for record in caplog.records:
+        assert record.levelname != "WARNING"
+    assert job.status == jobs.STATUS.DONE
 
 
 @requires_root
--- a/tests/storage/blocksd_test.py
+++ b/tests/storage/blocksd_test.py
@@ -1258,7 +1258,6 @@ def test_create_volume_with_new_bitmap(
     11,  # required size = lv size
     15,  # required size > lv size
 ])
-@pytest.mark.xfail(reason='Stale bitmaps in base volume not considered')
 def test_merge_with_bitmap(
         tmp_storage, tmp_repo, fake_access, fake_rescan, tmp_db, fake_task,
         fake_sanlock, monkeypatch, stale_bitmaps, caplog):


nirs:
> Measure also the base volume when prepare for a merge. We need to consider base volume bitmaps size, since measuring only top may result in untracked bitmaps even with backing enabled. 

This is not very clear, maybe explain that measuring only top does not consider the size
required for the bitmaps in the base image, which may result in failing to finish the commit
or to copy bitmaps from the top volume.

> This can happen if the bitmaps were added to base after the top volume was created.

This cannot happen in the current system. I think the reason is leftover bitmaps that
were not deleted when a checkpoint was deleted - known issue in engine.

> Considering both base and top volume bitmaps may result in allocating more than needed, but will avoid errors when commiting the image due to 'No space left on device'.

Or when copying bitmaps from top to base after the commit finish.

The expected cases are:

1. Bitmap is in both top and base - measuring only top is enough
2. Bitmap is only in top - measuring only top is good enough

The case we see in the related bug is bitmap that existing in base but not in top.
These bitmaps are broken and can never be used for backup, but they still take space
in the image and can cause live or cold merge to fail.

tests/storage/blocksd_test.py
```diff
@@ -1244,6 +1251,117 @@ def test_create_volume_with_new_bitmap(
...
+        imgUUID=img_id,
+        capacity=1 * GiB,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType='DATA',
+        volUUID=top_vol_uuid,
+        desc="Test top volume",
+        srcImgUUID=base_vol.imgUUID,
+        srcVolUUID=base_vol.volUUID,
+        add_bitmaps=True)
```
nirs, line 1309: Why use add_bitmaps? this creates invalid situation since the leaf has no bitmaps.
In a normal chain, bitmaps are always in the child volumes. For example, bitmap
can be missing from base but it must be in top and leaf.

tests/storage/blocksd_test.py
```diff
@@ -1244,6 +1251,117 @@ def test_create_volume_with_new_bitmap(
...
+        imgUUID=img_id,
+        capacity=1 * GiB,
+        volFormat=sc.COW_FORMAT,
+        preallocate=sc.SPARSE_VOL,
+        diskType='DATA',
+        volUUID=base_vol_uuid,
+        desc="Test volume",
+        srcImgUUID=sc.BLANK_UUID,
+        srcVolUUID=sc.BLANK_UUID,
+        bitmap="new-bitmap")
```
nirs, line 1294: We can create bitmap in base and have it in both top and leaf, but we can also have
bitmap in top and leaf. Lets keep this test simple as possible, having one bitmap in
top and leaf.

aesteve-rh:
Understood. I will update the comments and try to explain it better.

```
I think the reason is leftover bitmaps that
were not deleted when a checkpoint was deleted - known issue in engine.
```
ouch, yes, that could lead to this issue.

```
These bitmaps are broken and can never be used for backup, but they still take space
in the image and can cause live or cold merge to fail.
```
More so in the cold merge, as the live merge adds an entire extra chunk to avoid pausing if the guest writes data, which will be enough to "hide" the extra size of the untracked bitmaps.

It is not ideal that we potentially over extend unnecessarily the volume (as in the common case we potentially double the space required for bitmaps), but is better than having many reports due to this buggy scenario occurring.

tests/storage/blocksd_test.py
```diff
@@ -1244,6 +1251,117 @@ def test_create_volume_with_new_bitmap(
...
+    subchain_info = dict(sd_id=base_vol.sdUUID,
+                         img_id=base_vol.imgUUID,
+                         base_id=base_vol.volUUID,
+                         top_id=top_vol.volUUID,
+                         base_generation=0)
+    subchain = merge.SubchainInfo(subchain_info, host_id)
+
+    # Nearly fill the volume with data.
+    top_vol.prepare()
+    data_size = GiB - 5 * MiB
```
nirs, line 1335: The initial size of the base volume is 2.5g. We write here 1g, and we write 100 2k
bitmpas that takes 200K plus some space for the bitmap directory. How this can fail
with ENOSPC?

lib/vdsm/storage/merge.py
```diff
@@ -192,18 +192,29 @@ def _extend_base_allocation(base_vol, top_vol):
...
-    measure = qemuimg.measure(
+    top_measure = qemuimg.measure(
         top_vol.getVolumePath(),
         format=qemuimg.FORMAT.QCOW2,
         output_format=qemuimg.FORMAT.QCOW2,
         is_block=True,
         base=base_vol.getVolumePath())
-    log.debug("Measure result: %s", measure)
+    # There may be untracked bitmaps in base if they were added after the
+    # creating the top volume. Even if this is a rare case, we need to measure
```
nirs, line 202: There may be untracked bitmaps in base if they were left after failed
delete bitmap operation.

tests/storage/merge_test.py
```diff
@@ -236,17 +236,23 @@ def test_block_cow(self, monkeypatch, base, top, expected):
...
+                bitmaps = 0
                 # Make sure we are called with the right arguments.
-                assert image == top_vol.getVolumePath()
+                if image == top_vol.getVolumePath():
+                    assert base == base_vol.getVolumePath()
+                    # Top measure returns bitmaps.
+                    bitmaps = 1 * MiB
+                else:
+                    assert image == base_vol.getVolumePath()
+                    assert base is None
```
nirs, line 247: This is very fragile, it will better if the test would set a fake measure function that
returns different value for every volume. One way is to set a function that return values from a mapping from volume id to size of bitmaps.

tests/storage/merge_test.py
```diff
@@ -236,17 +236,23 @@ def test_block_cow(self, monkeypatch, base, top, expected):
...
                 assert format == qemuimg.FORMAT.QCOW2
                 assert output_format == qemuimg.FORMAT.QCOW2
                 assert backing
                 assert is_block
-                assert base == base_vol.getVolumePath()
                 assert not unsafe
 
                 # Return fake response.
-                return {"required": 2048 * MiB, "bitmaps": 1 * MiB}
+                return {"required": 2048 * MiB, "bitmaps": bitmaps}
```
nirs, line 255: This could be:

    return {"required": 2048 * MiB, "bitmaps": fake_bitmap_size[image]}

lib/vdsm/virt/livemerge.py
```diff
@@ -431,18 +431,28 @@ def _base_needs_extend(self, drive, job, base_info):
...
+            base_info["format"])
 
         # Consider the required size and possible bitmaps. Since some of the
         # bitmaps already exist in base, this allocates more space than needed,
         # but bitmaps are small so this is good enough.
-        required_size = measure["required"] + measure.get("bitmaps", 0)
-
+        required_size = (top_measure["required"]
+                         + top_measure.get("bitmaps", 0)
+                         + base_measure.get("bitmaps", 0))
```
nirs, line 455: This looks like the code we have in cold merge. Can we have a generic helper measuring
top and base paths, and use it from both cold and live merge?

tests/virt/livemerge_test.py
```diff
@@ -878,6 +878,35 @@ def test_internal_merge():
...
+    # Add base bitmaps not tracked in top measure.
+    # Merge should account for these bitmaps and extend.
+    vm.cif.irs.measure_info[(base_id, None)] = {
+        "bitmaps": 5 * MiB,
+    }
+    vm.merge(**merge_params)
+
+    # Merge persists job in EXTEND state.
+    persisted_job = parse_jobs(vm)[job_id]
+    assert persisted_job["state"] == Job.EXTEND
```
nirs, line 907: It will be better to separate the test from the actual code so we can verify that the 
test really fail without the code change. Otherwise can have a broken test that does nothing.

tests/storage/blocksd_test.py
```diff
@@ -1244,6 +1251,117 @@ def test_create_volume_with_new_bitmap(
...
+    dom.attach(tmp_repo.pool_id)
+
+    img_id = str(uuid.uuid4())
+    base_vol_uuid = str(uuid.uuid4())
+    top_vol_uuid = str(uuid.uuid4())
+
+    # Create base volume.
+    dom.createVolume(
+        imgUUID=img_id,
+        capacity=1 * GiB,
```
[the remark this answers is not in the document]
nirs: With the current test the chunk size does not mater since we fixed creation
of small volumes (size < chunk), but if we want to test big bitmaps we need
to use large volumes, and in this case using smaller chunk size will make
testing easier.

tests/storage/blocksd_test.py
```diff
@@ -13,29 +13,36 @@
...
+from fakelib import FakeNotifier
+from fakelib import FakeScheduler
 from storage.storagefakelib import fake_spm
 from storage.storagefakelib import fake_vg
 from storage.storagefakelib import FakeDomainMonitor
 from storage.storagefakelib import FakeTaskManager
+from storage.storagefakelib import fake_guarded_context
 
 TESTDIR = os.path.dirname(__file__)
 CHUNK_SIZE_MB = config.getint("irs", "volume_utilization_chunk_mb")
```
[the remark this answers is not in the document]
nirs: OK, this works since we fixed creation of small volumes.

tests/storage/blocksd_test.py
```diff
@@ -1244,6 +1251,117 @@ def test_create_volume_with_new_bitmap(
...
+def test_merge_with_bitmap(
+        tmp_storage, tmp_repo, fake_access, fake_rescan, tmp_db, fake_task,
+        fake_sanlock, monkeypatch):
+    sd_id = str(uuid.uuid4())
+
+    dev = tmp_storage.create_device(10 * GiB)
+    lvm.createVG(sd_id, [dev], blockSD.STORAGE_UNREADY_DOMAIN_TAG, 128)
+    vg = lvm.getVG(sd_id)
+
+    dom = blockSD.BlockStorageDomain.create(
```
nirs, line 1265: This storage domain remain active after the test - fixed in #355.

nirs:
Testing actual merge code is very complicated and requires too much mocking and faking and
is very hard to work with. I tried to reproduce the failure with the tests but could not
reproduce it.

I wrote a simple reproducer using qemu-img commands that reproduces the issue, but it proves that the computation of require size using top and base bitmaps is not safe, so this is not the right fix.

- Running the reproducer with 5 stale bitmaps the required size is less than the lv size, the merge succeeds, all bitmaps are valid, merging the bitmaps from top succeeds, and the good bitmap is valid.

- Running the reproducer with 6 stale bitmaps the required size is less than the lv size, the merge succeeds, merging the bitmaps from top succeeds, but the good bitmap is invalid.
```
Commit top into base
    (100.00/100%)
Image committed.
Add good bitmap to base
Merge good bitmap from top to base
qcow2_free_clusters failed: No space left on device
qemu-img: Lost persistent bitmaps during inactivation of node '#block159': Failed to write bitmap 'good-bitmap' to file: No space left on device
qemu-img: Failed to flush the refcount block cache: No space left on device
Info base after merge
{
  "virtual-size": 1073741824,
  "filename": "/dev/mapper/storage-base.qcow2",
  "cluster-size": 65536,
  "format": "qcow2",
  "actual-size": 0,
  "format-specific": {
    "type": "qcow2",
    "data": {
      "compat": "1.1",
      "compression-type": "zlib",
      "lazy-refcounts": false,
      "bitmaps": [
        ...
        {
          "flags": [
            "in-use",
            "auto"
          ],
          "name": "good-bitmap",
          "granularity": 65536
        }
      ],
      "refcount-bits": 16,
      "corrupt": false,
      "extended-l2": false
    }
  },
  "dirty-flag": false
}
```

- Running the reproducer with 11 stale bitmaps, the required size is exactly the lv size and the merge succeeds, merging the good bitmap success, and the good bitmap from top is valid. 
```
Commit top into base
    (100.00/100%)
qcow2_free_clusters failed: No space left on device
qemu-img: Lost persistent bitmaps during inactivation of node '#block308': Failed to write bitmap 'stale-bitmap-002' to file: No space left on device
qemu-img: Failed to flush the refcount block cache: No space left on device
Image committed.
Add good bitmap to base
Merge good bitmap from top to base
Info base after merge
{
  "virtual-size": 1073741824,
  "filename": "/dev/mapper/storage-base.qcow2",
  "cluster-size": 65536,
  "format": "qcow2",
  "actual-size": 0,
  "format-specific": {
    "type": "qcow2",
    "data": {
      "compat": "1.1",
      "compression-type": "zlib",
      "lazy-refcounts": false,
      "bitmaps": [
        ...
        {
          "flags": [
            "auto"
          ],
          "name": "good-bitmap",
          "granularity": 65536
        }
      ],
```

- Running the reproducer with 13 bitmaps, the required size is more than the lv size, but the merge succeeds. The good bitmap is invalid.
```
Commit top into base
    (100.00/100%)
qcow2_free_clusters failed: No space left on device
qemu-img: Lost persistent bitmaps during inactivation of node '#block326': Failed to write bitmap 'stale-bitmap-000' to file: No space left on device
qemu-img: Failed to flush the refcount block cache: No space left on device
Image committed.
Add good bitmap to base
Merge good bitmap from top to base
Info base after merge
{
  "virtual-size": 1073741824,
  "filename": "/dev/mapper/storage-base.qcow2",
  "cluster-size": 65536,
  "format": "qcow2",
  "actual-size": 0,
  "format-specific": {
    "type": "qcow2",
    "data": {
      "compat": "1.1",
      "compression-type": "zlib",
      "lazy-refcounts": false,
      "bitmaps": [
        ...
        {
          "flags": [
            "auto"
          ],
          "name": "good-bitmap",
          "granularity": 65536
        }
      ],
```

- Running the reproducer with 14 bitmaps the merge fails with ENOSPC.

This shows that we need more size than the size than top and base bitmaps and top required,
but we don't have a good way to compute this value.

Since the issue is stale bitmaps that we cannot use for anything, we can can fix the issue
by deleting all bitmaps in base that do not appear in top before the merge before the merge.

See how we find bitmaps in bitmaps._query_bitmaps(). We can add

```
bitmaps.prune_bitmaps(top, base)
```

that delete invalid bitmaps from base, and will be used before we measure base in cold
and live merge, and maybe later in a new repair_bitmaps vdsm command.

Regardless of the vdsm fix this reproducer reveals issues in `qemu-img commit` and
`qemu-img bitmap --merge`. The commands must fail if they cannot write bitmaps
to storage and they leave them with the "in-use" bit. This is a major failure for
`qemu-img bitamap --merge` since its purpose is to manipulate bitmaps. You can use
the reproducer to report qemu-img bug.

## Reproducer

```
# Reproduce ENOSPC error when merging into base image with lot of bitmaps.

import json
import subprocess

IMG_SIZE = 1 << 30
LV_SIZE = 128 << 20

# Testing shows that we can merge successfully with 13 bitmaps, and require
# size calculation is more strict, allowing up to 11 bitamps.
STALE_BITMAPS = 11

def run(*cmd):
    subprocess.run(cmd, check=True)

def output(cmd):
    cp = subprocess.run(cmd, stdout=subprocess.PIPE, check=True)
    return json.loads(cp.stdout)

def info(img):
    return output(["qemu-img", "info", "--output", "json", img])

def measure(img):
    return output(["qemu-img", "measure", "-f", "qcow2", "-O", "qcow2", "--output", "json", img])

def check(img):
    cmd = ["qemu-img", "check", "-f", "qcow2", "--output", "json", img]
    cp = subprocess.run(cmd, stdout=subprocess.PIPE)
    if cp.returncode not in (0, 3):
        raise RuntimeError(f"Check failed")

    return json.loads(cp.stdout)

def indent(info):
    return json.dumps(info, indent=2)

# Configuration - you need to create these lvs and chown them to the user
# running this script.
#   sudo lvcreate --name base.qcow2 --size 128m storage
#   sudo chown $USER:$USER /dev/mappe/stoarge-base.qcow2
#   sudo lvcreate --name top.qcow2 --size 128m storage
#   sudo chown $USER:$USER /dev/mappe/stoarge-top.qcow2
base = "/dev/mapper/storage-base.qcow2"
top = "/dev/mapper/storage-top.qcow2"

# Start with clean lvs by discarding current data.
run("blkdiscard", base)
run("blkdiscard", top)

print("Creating base")
run("qemu-img", "create", "-f", "qcow2", base, str(IMG_SIZE))

# Simulate stale bitmaps - missing in top.
for i in range(STALE_BITMAPS):
    bitmap = f"stale-bitmap-{i:03d}"
    print(f"Creating stale bitmap {bitmap}")
    run("qemu-img", "bitmap", "--add", base, bitmap)

print("Info base before merge")
base_info = info(base)
print(indent(base_info))

print("Check base before merge")
base_check = check(base)
print(indent(base_check))

print("Measure base before merge")
base_measure = measure(base)
print(indent(base_measure))

print("Creating top")
run("qemu-img", "create", "-f", "qcow2", "-b", base, "-F", "qcow2", top)

print("Adding good bitmap to top")
run("qemu-img", "bitmap", "--add", top, "good-bitmap")

print("Writing data to top")
cmd = f"write -P {ord('B')} 0 126m"
run("qemu-io", "-f", "qcow2", "-c", cmd, top)

print("Info top before merge")
top_info = info(top)
print(indent(top_info))

print("Check top before merge")
top_check = check(top)
print(indent(top_check))

print("Measure top before merge")
top_measure = measure(top)
print(indent(top_measure))

print("Commit top into base")
run("qemu-img", "commit", "-f", "qcow2", "-t", "none", "-b", base, "-d", "-p", top)

print("Add good bitmap to base")
run("qemu-img", "bitmap", "--add", base, "good-bitmap")

print("Merge good bitmap from top to base")
run("qemu-img", "bitmap", "--merge", "good-bitmap", "-F", "qcow2", "-b", top,  base, "good-bitmap")

print("Info base after merge")
print(indent(info(base)))

print("Check base after merge")
print(indent(check(base)))

print("Measure base after merge")
print(indent(measure(base)))

print("Checking that required size for merge is correct")
required_size = top_measure["required"] + top_measure["bitmaps"] + base_measure["bitmaps"]
print(f"required_size: {required_size} lv_size: {LV_SIZE}")
assert required_size <= LV_SIZE
```

nirs:
POC of a fix, using:
- Good bitmaps that are in both base and top
- Good bitmap that is only in top (e.g. bitmap created in backup snapshot)
- Stale bitmaps only in base

After the merge, base should have only the good bitmaps. If we add a merge test
we can simulate this case in the test.

```
# Reproduce ENOSPC error when merging into base image with lot of bitmaps.

import json
import subprocess

IMG_SIZE = 1 << 30
LV_SIZE = 128 << 20

STALE_BITMAPS = 32

def run(*cmd):
    subprocess.run(cmd, check=True)

def output(cmd):
    cp = subprocess.run(cmd, stdout=subprocess.PIPE, check=True)
    return json.loads(cp.stdout)

def info(img):
    return output(["qemu-img", "info", "--output", "json", img])

def measure(img):
    return output(["qemu-img", "measure", "-f", "qcow2", "-O", "qcow2", "--output", "json", img])

def check(img):
    cmd = ["qemu-img", "check", "-f", "qcow2", "--output", "json", img]
    cp = subprocess.run(cmd, stdout=subprocess.PIPE)
    if cp.returncode not in (0, 3):
        raise RuntimeError(f"Check failed")

    return json.loads(cp.stdout)

def bitmaps(img):
    img_info = info(img)
    return {b["name"] for b in img_info["format-specific"]["data"].get("bitmaps", [])}

def indent(info):
    return json.dumps(info, indent=2)

# Configuration - you need to create these lvs and chown them to the user
# running this script.
#   sudo lvcreate --name base.qcow2 --size 128m storage
#   sudo chown $USER:$USER /dev/mappe/stoarge-base.qcow2
#   sudo lvcreate --name top.qcow2 --size 128m storage
#   sudo chown $USER:$USER /dev/mappe/stoarge-top.qcow2
base = "/dev/mapper/storage-base.qcow2"
top = "/dev/mapper/storage-top.qcow2"

# Start with clean lvs by discarding current data.
run("blkdiscard", base)
run("blkdiscard", top)

print("Creating base")
run("qemu-img", "create", "-f", "qcow2", base, str(IMG_SIZE))

# Simulate stale bitmaps - missing in top.
for i in range(STALE_BITMAPS):
    bitmap = f"stale-bitmap-{i:03d}"
    print(f"Creating stale bitmap {bitmap}")
    run("qemu-img", "bitmap", "--add", base, bitmap)

print("Creating top")
run("qemu-img", "create", "-f", "qcow2", "-b", base, "-F", "qcow2", top)

# Add good bitmaps that will be in both base and top.
for i in range(2):
    bitmap = f"good-bitmap-{i:03d}"
    print(f"Adding good bitmap {bitmap} to base and top")
    run("qemu-img", "bitmap", "--add", base, bitmap)
    run("qemu-img", "bitmap", "--add", top, bitmap)

# Add bitmpas that exists only in top - will be added to based after the merge.
print(f"Adding top only bitmap to top")
run("qemu-img", "bitmap", "--add", top, "top-only")

print("Writing data to top")
cmd = f"write -P {ord('B')} 0 126m"
run("qemu-io", "-f", "qcow2", "-c", cmd, top)

print("Info base before merge")
base_info = info(base)
print(indent(base_info))

print("Info top before merge")
top_info = info(top)
print(indent(top_info))

print("Measure top before merge")
top_measure = measure(top)
print(indent(top_measure))

for b in sorted(bitmaps(base) - bitmaps(top)):
    print(f"Removing stale bitmap {b} from base")
    run("qemu-img", "bitmap", "--remove", base, b)

print("Commit top into base")
run("qemu-img", "commit", "-f", "qcow2", "-t", "none", "-b", base, "-d", "-p", top)

# Merging bitmaps only in top to base.
for b in sorted(bitmaps(top) - bitmaps(base)):
    print("Add top only bitmap to base")
    run("qemu-img", "bitmap", "--add", base, b)
    print("Merge top only bitmap from top to base")
    run("qemu-img", "bitmap", "--merge", b, "-F", "qcow2", "-b", top,  base, b)

print("Info base after merge")
print(indent(info(base)))

print("Check base after merge")
print(indent(check(base)))

print("Measure base after merge")
print(indent(measure(base)))

print("Checking that required size for merge is correct")
required_size = top_measure["required"] + top_measure["bitmaps"]
print(f"required_size: {required_size} lv_size: {LV_SIZE}")
assert required_size <= LV_SIZE
```

tests/storage/blocksd_test.py
```diff
@@ -1244,6 +1251,117 @@ def test_create_volume_with_new_bitmap(
...
+    top_vol.teardown(sd_id, top_vol_uuid)
+
+    # Add bitmaps to the base volume after creating top volume.
+    base_vol.prepare()
+    for i in range(100):
+        qemuimg.bitmap_add(base_vol.getVolumePath(), f"bitmap{i}").run()
+    base_vol.teardown(sd_id, base_vol_uuid)
+
+    with monkeypatch.context() as mc:
+        mc.setattr(guarded, 'context', fake_guarded_context())
```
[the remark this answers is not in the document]
aesteve-rh: Yes we need it. If we don't, then the `prepare()` operation timeouts without acquiring the lock.

So we need to call `dom.acquireHostId`, but then, even if we release it, the `test_spm_lifecycle` was failing somewhere in the sanlock code.

I saw this mock in the `merge_test` module, and since the guarded context is not important for this test, I imported it here too.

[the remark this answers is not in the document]
nirs: Maybe add a TODO to check why this is needed.

lib/vdsm/storage/sdm/api/merge.py
```diff
@@ -42,6 +42,14 @@ def _run(self):
             with self.subchain.prepare(), self.subchain.volume_operation():
                 top_vol_path = self.subchain.top_vol.getVolumePath()
                 base_vol_path = self.subchain.base_vol.getVolumePath()
+
+                # There may be stale bitmaps in base if they were left
+                # after failed delete bitmap operation. These bitfields may
+                # cause the bitfields merge operation to fail with ENOSPC.
+                # As there is not a reliable way to calculate the size of
+                # these bitfields, we prune them before the commit.
```
nirs, line 50: and they are invalid and can never be use for incremental backup.

lib/vdsm/storage/sdm/api/merge.py
```diff
@@ -42,6 +42,14 @@ def _run(self):
...
+                bitmaps.prune_bitmaps(base_vol_path, top_vol_path)
```
nirs, line 51: It could be better to prune bitmaps when we prepare for merge, since we measure and extend base volume at that point. But we know that measure does not consider the bitmaps in base, so pruning before or after measure is the same. If qemu-img measure will start to consider bitmaps in base, we may over allocate the base volume before merge. If we want to avoid this issue, we can move pruning to the prepare step (vdsm.storage.merge.prepare).

lib/vdsm/virt/livemerge.py
```diff
@@ -364,12 +364,17 @@ def _start_commit(self, drive, job):
...
+            base_path = drive.volume_path(job.base)
+            top_path = drive.volume_path(job.top)
         except VolumeNotFound as e:
             self._untrack_job(job.id)
             raise exception.MergeFailed(
                 str(e), top=job.top, base=job.base, chain=actual_chain,
                 job=job.id)
 
+        # Prune stale bitmaps from base volume.
+        self._dom._vm.prune_bitmaps(base_path, top_path)
```
nirs, line 376: Few issues:
- This cannot be sued for active layer merge, since bitmaps created while the vms is running were not written to disk yet, and bitmaps that were on disk are all with the in-use flag.
- Like cold merge, it is more future proof to prune bitmaps before we measure the volume, not right before commit.
- We should pass domain, image and volume uuids to storage apis, not paths.

lib/vdsm/virt/livemerge.py
```diff
@@ -364,12 +364,17 @@ def _start_commit(self, drive, job):
         try:
             base_target = drive.volume_target(job.base, actual_chain)
             top_target = drive.volume_target(job.top, actual_chain)
+            base_path = drive.volume_path(job.base)
+            top_path = drive.volume_path(job.top)
```
nirs, line 368: We don't work with paths here, can be removed.

lib/vdsm/virt/vmdevices/storage.py
```diff
@@ -686,6 +686,25 @@ def iotune(self, value):
         vmtune.validate_io_tune_params(iotune)
         self._iotune = iotune
 
+    def volume_path(self, vol_id):
```
nirs, line 689: Not needed since we should pass uuids to storage. The storage layer knows how to locate the volume path.

nirs (changes requested):

lib/vdsm/storage/sdm/api/merge.py
```diff
@@ -42,6 +42,14 @@ def _run(self):
             with self.subchain.prepare(), self.subchain.volume_operation():
                 top_vol_path = self.subchain.top_vol.getVolumePath()
                 base_vol_path = self.subchain.base_vol.getVolumePath()
+
+                # There may be stale bitmaps in base if they were left
+                # after failed delete bitmap operation. These bitfields may
```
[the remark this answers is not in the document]
aesteve-rh: Blast from the past when I used to work with bitmaps, I guess was my mind going auto while writing...

lib/vdsm/virt/livemerge.py
aesteve-rh: Yes yes, this commit needs to be reworked.

lib/vdsm/virt/livemerge.py
aesteve-rh: I'll split this commit into another PR as I rework it.

lib/vdsm/storage/bitmaps.py
```diff
@@ -125,6 +125,25 @@ def remove_bitmap(vol_path, bitmap):
             vol_path=vol_path)
 
 
+def prune_bitmaps(base_path, top_path):
+    """
+    Prune all the stale bitmaps from the base volume path.
+    A bitmap is considered stale if it appears only in the base volume,
+    and is missing in the top volume.
```
nirs, line 132: is missing -> is missing or invalid

lib/vdsm/storage/bitmaps.py
```diff
@@ -125,6 +125,25 @@ def remove_bitmap(vol_path, bitmap):
...
+        base_path (str): Path to the base volume
+        top_path (str): Path to the top volume
+    """
+    base_bitmaps = _query_bitmaps(base_path)
+    valid_top_bitmaps = _query_bitmaps(top_path, filter=_valid)
+
+    for name in base_bitmaps:
+        if name not in valid_top_bitmaps:
+            log.warning("Prune stale bitmap %s from %r", name, base_path)
+            remove_bitmap(base_path, name)
```
nirs, line 144: If we have large amount of stale bitmaps, it may be better to have one log for all, for example:

    stale_bitmaps = [name for name in base_bitmaps if name not in valid_top_bitmaps]
    if stale_bitmaps:
        log.warning("Pruning stale bitmaps %s from %r", stale_bitmaps, base_path)
        for bitmap in stale_bitamps:
            delete it...

tests/storage/bitmaps_test.py
```diff
@@ -234,6 +236,78 @@ def test_remove_non_existing_bitmap_succeed(tmp_mount, vol_chain):
...
+    ]
+
+
+def test_prune_disabled_bitmaps(tmp_mount, vol_chain):
+    # Add valid bitmap to volumes
+    bitmap = 'valid-bitmap'
+    qemuimg.bitmap_add(vol_chain.base_vol, bitmap).run()
+    qemuimg.bitmap_add(vol_chain.top_vol, bitmap).run()
+
+    # Add disabled bitmaps to volumes
```
nirs, line 267: Add disabled bitmap to top

tests/storage/bitmaps_test.py
```diff
@@ -234,6 +236,78 @@ def test_remove_non_existing_bitmap_succeed(tmp_mount, vol_chain):
...
+    ]
+
+
+def test_prune_in_use_bitmaps(tmp_mount, vol_chain):
+    # Add inconsistent "in-use" bitmaps to volumes
+    for i in range(5):
+        bitmap_in_use = f"in-use-bitmap-{i}"
+        qemuimg.bitmap_add(vol_chain.base_vol, bitmap_in_use).run()
+        qemuimg.bitmap_add(vol_chain.top_vol, bitmap_in_use).run()
+        qemuio.abort(vol_chain.top_vol)
```
nirs, line 292: We can run qemuio.abort() once after adding all the bitmaps. It will break all bitmaps in the volume.

nirs (commented):
Looks good, suggested minor tweaks.

tests/storage/bitmaps_test.py
aesteve-rh: I knew since I tried before, but still indented this wrongly... Fixing.

nirs (approved):

PR #355, opened 2022-11-18
Fix tmpstorage teardown

nirs:
Since we moved to lvm devices, tmpstorage teardown is broken, and does not deactivate and remove vgs during teardown. After running a test creating a temporary loop device and creating a storage domain, the storage domain lvs are left active:

    $ sudo dmsetup table
    bae261e2--3bfb--473d--86df--a63f92375df7-ids: 0 262144 linear 7:0 526336
    bae261e2--3bfb--473d--86df--a63f92375df7-inbox: 0 262144 linear 7:0 788480
    bae261e2--3bfb--473d--86df--a63f92375df7-leases: 0 4194304 linear 7:0 1312768
    bae261e2--3bfb--473d--86df--a63f92375df7-master: 0 2097152 linear 7:0 7604224
    bae261e2--3bfb--473d--86df--a63f92375df7-metadata: 0 262144 linear 7:0 264192
    bae261e2--3bfb--473d--86df--a63f92375df7-outbox: 0 262144 linear 7:0 1050624
    bae261e2--3bfb--473d--86df--a63f92375df7-xleases: 0 2097152 linear 7:0 5507072
    cs-root: 0 94453760 linear 252:2 2048
    cs-swap: 0 8298496 linear 252:2 94455808

and the loop device is not detached:

    $ losetup -a
    /dev/loop0: []: (/var/tmp/vdsm/test_merge_with_bitmap0/backing-file-000)

and the backing-file data is not deleted when deleting the backing file. This quickly leads to filling the disk on the host running the tests.

The issue is using a lvm filter in tmpstorage; when using lvm devices, the lvm filter is ignored, we don't find any vg, so we don't clean up the lvm resources.

Fix by dropping the lvm config, and using --devices with all the devices added to tmpstorage.

Tested by running lvm and blocksd tests on centos stream 8 host:

    sudo ~/.venv/vdsm/bin/tox -e storage-root tests/storage/{lvm,blocksd}_test.py -- -v -m root

### DIFF
--- a/tests/storage/tmpstorage.py
+++ b/tests/storage/tmpstorage.py
@@ -7,7 +7,6 @@ from __future__ import division
 import itertools
 import logging
 import os
-import re
 
 import six
 
@@ -21,17 +20,6 @@ log = logging.getLogger("test")
 
 
 class TemporaryStorage(object):
-
-    CONF = """
-        global {
-         use_lvmetad=0
-        }
-        devices {
-         write_cache_state=0
-         filter=%s
-        }
-    """
-    CONF = re.sub(r"\s+", " ", CONF).strip()
 
     def __init__(self, tmpdir):
         self._tmpdir = tmpdir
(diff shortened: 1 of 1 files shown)


aesteve-rh (approved):
Thanks!

I did notice it in my runs, but I was focused on the bug and didn't investigate this issue further. I was periodically rebooting the VM when it run out of space while testing...

mz-pdm (approved):

PR #357, opened 2022-11-30
livemerge: prune bitmaps before measuring

aesteve-rh:
Prune stale bitmaps before a live merge to avoid
failing with ENOSPC.

Fixes: #352

### DIFF
--- a/lib/vdsm/storage/hsm.py
+++ b/lib/vdsm/storage/hsm.py
@@ -34,6 +34,7 @@ from vdsm.common.threadlocal import vars
 from vdsm.common.time import monotonic_time
 from vdsm.common.units import MiB, GiB
 from vdsm.config import config
+from vdsm.storage import bitmaps
 from vdsm.storage import blockSD
 from vdsm.storage import clusterlock
 from vdsm.storage import constants as sc
@@ -42,6 +43,7 @@ from vdsm.storage import dispatcher
 from vdsm.storage import exception as se
 from vdsm.storage import fileUtils
 from vdsm.storage import glusterSD
+from vdsm.storage import guarded
 from vdsm.storage import image
 from vdsm.storage import imagetickets
 from vdsm.storage import iscsi
@@ -2856,6 +2858,30 @@ class HSM(object):
             unsafe=True)
 
         return dict(result=result)
+
+    @public
+    def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
+        """
+        Prune stale bitmaps from the base volume.
+        Warning: Internal use only (i.e., from virt code).
+
+        Arguments:
+            sdUUID (str): The UUID of the storage domain that owns the volume.
+            imgUUID (str): The UUID of the image contained on the volume.
+            volUUID (str): The UUID of the top volume.
+            baseUUID (str): The UUID of the volume that will get its bitmaps
+                pruned if missing or invalid in top.
+        """
+        img_ns = rm.getNamespace(sc.IMAGE_NAMESPACE, sdUUID)
+        locks = [rm.Lock(sc.STORAGE, sdUUID, rm.SHARED),
+                 rm.Lock(img_ns, imgUUID, rm.EXCLUSIVE)]
+        with guarded.context(locks):
+            dom = sdCache.produce_manifest(sdUUID)
+            base_vol = dom.produceVolume(imgUUID, baseUUID)
+            top_vol = dom.produceVolume(imgUUID, volUUID)
+            bitmaps.prune_bitmaps(
+                base_path=base_vol.getVolumePath(),
+                top_path=top_vol.getVolumePath())
 
     @public
     def appropriateDevice(self, guid, thiefId, deviceType):
--- a/lib/vdsm/virt/livemerge.py
+++ b/lib/vdsm/virt/livemerge.py
@@ -287,6 +287,16 @@ class DriveMerger:
             except JobExistsError as e:
                 raise exception.MergeFailed(str(e), job=job.id)
 
+            # Prune stale bitmaps if needed once before measuring the image.
+            # Currently, qemu-img ignores bitmaps in base, but future
+            # version may not.
+            if self._base_needs_prune_bitmaps(job, base_info):
+                self._dom._vm.prune_bitmaps(
+                    drive.domainID,
+                    drive.imageID,
+                    job.top,
+                    job.base)
+
             needs_extend, new_size = self._base_needs_extend(
                 drive, job, base_info)
 
@@ -401,6 +411,19 @@ class DriveMerger:
         except libvirt.libvirtError as e:
             self._untrack_job(job.id)
             raise exception.MergeFailed(str(e), job=job.id)
+
+    def _base_needs_prune_bitmaps(self, job, base_info):
+        if base_info['format'] != 'COW':
+            log.debug("Base volume does not support bitmaps, "
+                      "job=%r base=%r", job.id, job.base)
+            return False
+
+        if job.active_commit:
+            log.debug("Top volume is the active volume, "
+                      "job=%r top=%r", job.id, job.top)
+            return False
+
+        return True
 
     def _base_needs_extend(self, drive, job, base_info):
         """
--- a/lib/vdsm/virt/vm.py
+++ b/lib/vdsm/virt/vm.py
@@ -6106,6 +6106,22 @@ class Vm(object):
 
     # Accessing storage
 
+    def prune_bitmaps(self, domainID, imageID, topID, baseID):
+        """
+        Prune stale bitmaps from the base volume.
+        """
+        res = self.cif.irs.prune_bitmaps(
+            sdUUID=domainID,
+            imgUUID=imageID,
+            volUUID=topID,
+            baseUUID=baseID)
+
+        if res['status']['code'] != 0:
+            message = res['status']['message']
+            raise errors.StorageUnavailableError(
+                f"Unable to prune bitmaps in domain {domainID} "
+                f"image {imageID} top {topID} base {baseID}: {message}")
+
     def measure(self, domainID, imageID, topID, dest_format, backing=True,
                 baseID=None):
         """
--- a/tests/virt/vmfakelib.py
+++ b/tests/virt/vmfakelib.py
@@ -41,6 +41,7 @@ class IRS(object):
         self.extend_requests = []
         self.sd_types = {}
         self.measure_info = {}
+        self.errors = {}
 
     def getDeviceVisibility(self, guid):
         pass
@@ -53,6 +54,9 @@ class IRS(object):
 
     @recorded
     def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
+        error = self.errors.get("prune_bitmaps")
+        if error:
+            return response.error_raw(error.code, error.msg)
         return response.success(result=None)
 
     def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
--- a/tests/virt/vmfakelib.py
+++ b/tests/virt/vmfakelib.py
@@ -51,6 +51,7 @@ class IRS(object):
     def inappropriateDevices(self, ident):
         pass
 
+    @recorded
     def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
         return response.success(result=None)
 
--- a/tests/virt/vmfakelib.py
+++ b/tests/virt/vmfakelib.py
@@ -51,6 +51,9 @@ class IRS(object):
     def inappropriateDevices(self, ident):
         pass
 
+    def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
+        return response.success(result=None)
+
     def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
                 baseUUID=None):
         # Return fake measure set up by the test. Not setting up anyting will
--- a/tests/virt/livemerge_test.py
+++ b/tests/virt/livemerge_test.py
@@ -17,6 +17,7 @@ from vdsm.common import xmlutils
 from vdsm.common.config import config
 from vdsm.common.units import GiB, MiB
 
+from vdsm.virt import errors
 from vdsm.virt import metadata
 from vdsm.virt import migration
 from vdsm.virt import thinp
@@ -887,6 +888,23 @@ def test_internal_merge():
     drive = vm.getDiskDevices()[0]
     assert drive.volumeChain == expected_volumes_chain
     assert vm.volume_monitor.enabled
+
+
+def test_internal_merge_prune_error():
+    config = Config('internal-merge')
+    merge_params = config.values["merge_params"]
+
+    vm = RunningVM(config)
+
+    assert vm.query_jobs() == {}
+
+    # Simulate prune_bitmaps error.
+    error_msg = "Failed to remove bitmap"
+    vm.cif.irs.errors["prune_bitmaps"] = exception.RemoveBitmapError(error_msg)
+
+    with pytest.raises(errors.StorageUnavailableError) as e:
+        vm.merge(**merge_params)
+    assert error_msg in str(e)
 
 
 def test_extend_timeout_recover(fake_time):
--- a/tests/virt/livemerge_test.py
+++ b/tests/virt/livemerge_test.py
@@ -856,9 +856,20 @@ def test_internal_merge():
     # Check current chain matches expected chain - current chain is unordered.
     assert set(vdsm_chain) == set(new_chain)
 
-    # Check imageSyncVolumeChain was called with correct arguments.
-    assert len(vm.cif.irs.__calls__) == 1
+    assert len(vm.cif.irs.__calls__) == 2
+    # First call to prune_bitmaps, as it is supported and may be required.
+    # This is an internal merge with cow volumes.
     meth, arg, kwarg = vm.cif.irs.__calls__[0]
+    assert meth == "prune_bitmaps"
+    assert kwarg == {
+        "sdUUID": sd_id,
+        "imgUUID": img_id,
+        "volUUID": top_id,
+        "baseUUID": base_id
+    }
+
+    # Check imageSyncVolumeChain was called with correct arguments.
+    meth, arg, kwarg = vm.cif.irs.__calls__[1]
     assert meth == "imageSyncVolumeChain"
     assert arg[:3] == (sd_id, img_id, drive.volumeID)
     assert set(arg[3]) == set(new_chain)


aesteve-rh:
Currently I am trying to verify the fix with a stress test. But I post the PR already to refine it early, so that we can integrate as soon as I manage to reproduce the issue (if so).

lib/vdsm/storage/hsm.py
```diff
@@ -2857,6 +2858,25 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
 
         return dict(result=result)
 
+    @public
+    def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
+        """
+        Prune stale bitmaps from the base volume.
```
nirs, line 2864: Please document that this is an internal that should be used only from virt code.

lib/vdsm/storage/hsm.py
```diff
@@ -2857,6 +2858,25 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
...
+
+        Arguments:
+            sdUUID (str): The UUID of the storage domain that owns the volume.
+            imgUUID (str): The UUID of the image contained on the volume.
+            volUUID (str): The UUID of the top volume.
+            baseUUID (str): The UUID of the volume that will get its bitmaps
+                pruned if missing or invalid in top.
+        """
+        base_vol = self._produce_volume(sdUUID, imgUUID, baseUUID)
+        top_vol = self._produce_volume(sdUUID, imgUUID, volUUID)
```
nirs, line 2874: Using _produce_volume() is nice, but it only a shared lock on the storage domain.

We should to take also:
- shared lock on the image? (check how storage jobs lock similar operations)
- shared lock on the top volume
- exclusive lock on the base volume (modified using this code)

Since this should be called only from virt code, and the vm owns the volumes,
no other storage code should access them (engine should ensure this) but it
is safer to take locks so bad engine request will fail at vdsm instead of
corrupting the volume.

You can add locking in a different commit to simplify the work.

lib/vdsm/virt/livemerge.py
```diff
@@ -287,6 +287,13 @@ def merge(self, driveSpec, base, top, bandwidth, job_id):
             except JobExistsError as e:
                 raise exception.MergeFailed(str(e), job=job.id)
 
+            # Prune stale bitmaps from base volume before measuring.
```
nirs, line 290: This should be called once when starting a merge - good.

But cannot be used when:
- top is the leaf volume - active layer merge, since top can have bitmaps which did not appear in the image yet. New bitamps created when the vm is running will be written to the image only when the vm stop or when we create a snapshot.
- base is not using qcow2 format, prune will fail

So I think we need something like:

    if self._base_needs_prune_bitmaps(drive, job, base_info):
        self._dom._vm.prune_bitmaps(...)

lib/vdsm/virt/vm.py
```diff
@@ -6106,6 +6106,16 @@ def onDeviceRemoved(self, device_alias):
 
     # Accessing storage
 
+    def prune_bitmaps(self, domainID, imageID, topID, baseID):
+        """
+        Prune stale bitmaps from the base volume.
+        """
+        self.cif.irs.prune_bitmaps(
```
nirs, line 6113: You are not checking errors - this will not raise on errors!

Check how error are handled in measure() bellow.

tests/virt/vmfakelib.py
```diff
@@ -51,6 +51,9 @@ def appropriateDevice(self, guid, vmid):
     def inappropriateDevices(self, ident):
         pass
 
+    def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
+        pass
```
nirs, line 55: This should use `@mayfail` so we can simulate errors and test merge behavior on failures.

It should also record the fact that this was called and the called arguments, we have
a `@recorded` decorator for this, see how it is used. This will allow verifying that we
prone bitmaps when starting merge for internal merge, and not pruning bitmaps for
active layer merge or internal merge to raw base.

Check how livemerge_test.py verify the flows. We need to add some more asserts in the
existing tests.

nirs (changes requested):
To simplify this work, you can split the commits:
- Add internal api to hsm
- Use the new api in live merge
- Update the tests to verify that we prune bitmaps only when we should

Current pr should be good enough to test the happy case for internal merge - you should be able to reproduce this by adding bitmaps manually to running vm base volume.

lib/vdsm/virt/vm.py
aesteve-rh: It will, if it fails it will raise an `exception.RemoveBitmapError` from the bitmaps module [1].

[1] https://github.com/oVirt/vdsm/blob/master/lib/vdsm/storage/bitmaps.py#L122

nirs: The exception is not raised in the caller. HSM is wrapped by Dispatcher, which warps public methods to return status dict instead of raising.

aesteve-rh: Ah! I see it now, I though `res` in measure was the return of the hsm measure (dict from the parsed json), but now I see the dispatcher in the decorator. Got it.

tests/virt/vmfakelib.py
aesteve-rh: I saw how `@maybefail` works, is a handy wrapper!

But in this case I think is better fit to use the `response.error_raw` if we want to simulate errors, something like what we do in `getVolumeSize` below, to have similar code to this one:
```python
    @recorded
    def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
        if "prune_bitmaps" in self.errors:
            error = self.errors["prune_bitmaps"]
            return response.error_raw(error.code, error.msg)
        return response.success(result=None)
```

nirs: VM.prune_bitmaps() should not return status code! It should raise on errors. Its purpose is to covert the (horrible) self.irs.cif.prune_bitmaps() call returning status codes to proper call raising on errors.

aesteve-rh: It does raise on errors. This `prune_bitmaps` is in the fake IRS class, not in the VM.

So it returns a status code that is processed by `VM.prune_bitmaps()`, and then raise the error.

The fact that both have the same name may cause some confusion.

nirs: The fake IRS should be wrapped with a Dispatcher so it behaves like the real cif.irs object.

But if the current code is not using a dispatcher and try to behave like a wrapped hsm
and return status codes, you should do the same.

aesteve-rh:
I tried to reproduce the issue, both manually and with ovirt-stress, but I couldn't get to see the "No space left on device" in the libvirtd logs on a full setup with two hosts and a running VM. Unfortunately, I can not spend more time with it.

So I would finish the review of this PR, and integrate it. Hopefully on qe they have better luck reproducing.

nirs:
> I tried to reproduce the issue, both manually and with ovirt-stress, but I couldn't get to see the "No space left on device" in the libvirtd logs on a full setup with two hosts and a running VM.

How did you try to reproduce?

When we start live merge we measure the top volume to get the required size of base. This
measure the data and bitmaps in base.
https://github.com/oVirt/vdsm/blob/6f67a73dac7c02009c0125a59e002590f9cc2beb/lib/vdsm/virt/livemerge.py#L405

If base size is smaller than the measured data, we don't extend base.
https://github.com/oVirt/vdsm/blob/6f67a73dac7c02009c0125a59e002590f9cc2beb/lib/vdsm/virt/livemerge.py#L457

So we need to find how much data we must have in top so the measured size is exactly
base size. Then add eough stale bitmaps to base (we needed 6 bitmaps in the tests to
reproduce this).

You can do the measurements with the reproducer script, using 2.5g images (instead of 128m).
When you find the right size, do:

1. Add second disk to running vm using qcow2 format
   - This disk will start as 2.5g empty disk
2. Create a snapshot (snap1)
   - This will create empty 2.5g snapshot
3. Fill this snapshot with the right amount of data so merge should succeed
   - Can be done inside the vm by writing random data directly to the second disk
     ```
     dd if=/dev/urandom bs=1M count=xxx of=/dev/vdb oflag=direct conv=fsync
     ```
     writing zeros will be faster but when committing the data qemu detects
     zeros and convert them to zero clusters, which will prevent reproduction
     of this issue.
4. The snapshot will be extend to 5g during the write - this is fine and does
   not affect the test
5. Create another snapshot (snap2)
   - now we can merge the middle layer into the base
6. Add enough stale bitmaps using `qemu-img bitmap add` to the base volume
7. In engine UI, delete snapshot snap1
   - this merges contents of the middle layer (snap2) into the base layer (snap1)
   - it should fail because of the stale bitmaps

I think we need to reproduce this first before we modify the code. If you don't have
time, maybe someone from QE has time to reproduce.

aesteve-rh:
It is exactly how I tried.

First I tried with 3GiB disk, but was not fully provisioned. Vdsm's created image was 2.5GiB, but virtual size was 3GiB, so I changed it to 2.5GiB (2560MiB). Didn't get it either, and I even tried with 128MiB as is what we used for cold merges. No luck.

Tried to write directly in the VM (to be fair I did only try with `/dev/zero`, so today I retried with `/dev/urandom`), or with `qemu-io` in the host directly in the top volume:
```shell
qemu-io -f qcow2 {top_path} -c "write 0 2000M"
qemu-io -f qcow2 {top_path} -c "write 2000M 558M"
```

I did try also with more snapshots, removing the first or the second snapshot (these tests helped understand the process better, actually), automatically and manually.

None of these really worked. At least in what my goal was: I didn't get the qemu error in `journactl` output in any case.

I will document the steps and hand over to QE, I really can not put much more time into trying. I wanted to move on completely today, but still dedicated the morning to try with the `/dev/urandom` proposal, with no luck either.

My feeling with the patch is that it should work, same as we fixed the cold merges: basically runs the same operation as with cold merges, and the test proves that it runs when required. If OST runs pass (it does not degrade), I'm good with it.

nirs:
> It is exactly how I tried.
> 
> First I tried with 3GiB disk, but was not fully provisioned. Vdsm's created image was 2.5GiB, but virtual size was 3GiB, so I changed it to 2.5GiB (2560MiB). Didn't get it either, and I even tried with 128MiB as is what we used for cold merges. No luck.
> 
> Tried to write directly in the VM (to be fair I did only try with `/dev/zero`, 

This cannot work, qemu detects the zeros during the merge and does not write any
data, so merge cannot fail.

... 
> None of these really worked. At least in what my goal was: I didn't get the qemu error in `journactl` output in any case.

Getting the error is not reliable way to detect the issue. The issue is detected by
checking the merge logs. When the merge complete after an error, the volume is not
removed from the chain, and engine will report the the snapshot is illegal.

When this happens, trying to delete the snapshot again is expected to fail forever.

lib/vdsm/storage/hsm.py
```diff
@@ -2871,6 +2871,9 @@ def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
             baseUUID (str): The UUID of the volume that will get its bitmaps
                 pruned if missing or invalid in top.
         """
+        vars.task.getSharedLock(STORAGE, imgUUID)
+        vars.task.getSharedLock(STORAGE, volUUID)
+        vars.task.getExclusiveLock(STORAGE, baseUUID)
```
nirs, line 2876: This not how storage locks are taken, see how we lock the image and the volume in the sp.py.

nirs: So more specific advice - we cannot take volume locks, since locking the image take a lock on all volumes:

I think it works like this:
1. sd.py _registerResourceNamespaces register imageResourcesNamespace with imageResourceFactory
2. imageResourceFactory takes a lock on all volume of the image when creating a resource
3. So when you take a lock on an image, you lock all the volume in this image with the same lock type

Since we want an exclusive lock on the base volume, we must take exclusive lock on the image, which will take also exclusive lock on the top volume.

So we need something like:

    img_ns = rm.getNamespace(sc.IMAGE_NAMESPACE, sdUUID)
    with rm.acquireResource(img_ns, imgUUID, rm.EXCLUSIVE):
        purne bitmaps here...

BUT: this must be AFTER we take the storage domain lock, otherwise we can have a deadlock. All locks must be taken in the same order:

1. sc.STORAGE
2. sc.IMAGE_NAMESPACE
3. sc.VOLUME_NAMESPACE
4. sc.LVM_ACTIVATION_NAMESPACE

This is why the constants are named: "00_storage", "01_img", "02_vol", "03_lvm".

Code taking multiple locks must sort the locks to ensure the order.

lib/vdsm/storage/hsm.py
```diff
@@ -2871,6 +2871,9 @@ def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
...
         base_vol = self._produce_volume(sdUUID, imgUUID, baseUUID)
         top_vol = self._produce_volume(sdUUID, imgUUID, volUUID)
```
nirs, line 2878: This does not work, since we _produce_volume takes a shared lock on the volume, and this code tries to take the same lock twice:

    def _produce_volume(self, sdUUID, imgUUID, volUUID):
        vars.task.getSharedLock(STORAGE, sdUUID)
        ...

So code is very clean but incorrect. What we want to do is to have list of locks, and lock of of them at the same time and in the right order using guarded.context().

    img_ns = rm.getNamespace(sc.IMAGE_NAMESPACE, self.sd_id)
    locks = [rm.Lock(sc.STORAGE, sdUUID, rm.SHARED),
             rm.Lock(img_ns, imgUUID, rm.EXCLUSIVE)]
    with guarded.context(locks):
        prune bitmaps here...

You need to produce the volume manually instead of using the _produce_volume() helper, it does too much and cannot be used when using multiple volumes. Maybe it can be fixed to support this usage but lets keep this change simple and avoid unneeded changes that can break unrelated flows and need much more testing.

nirs (changes requested):
Partial review about locking. I will review the rest later this week.

aesteve-rh:
I have finally been able to ***consistently*** reproduce the issue with 2560MiB disks.

# Reproducing steps

1. Create snap1
2. Write 2555MiB from the running VM:
```
dd if=/dev/urandom bs=1M count=2555 of=/dev/vda oflag=direct conv=fsync
```
3. Create snap2
4. Add 70 bitmaps to base image from the Host:
```
for i in $(seq 70); do
    qemu-img bitmap --add /dev/<vg_name>/<lv_name> stale-bitmap-$i
done
```
Results in:
Top image measure:
```
required size: 2679767040
fully allocated size: 2685009920
bitmaps size: 0
```
Base image measure:
```
required size: 655360
fully allocated size: 2685009920
bitmaps size: 9240576
```
5. Remove snap1 from engine UI to trigger the internal merge.

# Error case (without the fix in this PR)

Vdsm logs:
```
2022-12-05 18:21:39,245+0100 ERROR (libvirt/events) [virt.vm] (vmId='bb1ff201-a159-4436-a23f-8dad4cced1e9') Block job 89667ef9-0b0f-4b63-8ed9-9a720e0d4845 type COMMIT for drive vdb has failed (vm:5955)
```
Resulting base measure:
```
required size: 2675048448
fully allocated size: 2685009920
bitmaps size: 9240576
```
Engine UI:
image

If we skip adding the bitmaps the merge succeeds, so it is an error caused by the stale bitmaps.

# Success case (with the fix in this PR):
Vdsm logs.

After `START merge` and before measuring the relevant disk, we see the prune_bitmaps logs:
```
2022-12-07 09:49:47,864+0100 WARN  (jsonrpc/2) [storage.bitmaps] Prune stale bitmaps ['stale-bitmap-1', 'stale-bitmap-2', 'stale-bitmap-3', 'stale-bitmap-4', 'stale-bitmap-5', 'stale-bitmap-6', 'stale-bitmap-7', 'stale-bitmap-8', 'stale-bitmap-9', 'stale-bitmap-10', 'stale-bitmap-11', 'stale-bitmap-12', 'stale-bitmap-13', 'stale-bitmap-14', 'stale-bitmap-15', 'stale-bitmap-16', 'stale-bitmap-17', 'stale-bitmap-18', 'stale-bitmap-19', 'stale-bitmap-20', 'stale-bitmap-21', 'stale-bitmap-22', 'stale-bitmap-23', 'stale-bitmap-24', 'stale-bitmap-25', 'stale-bitmap-26', 'stale-bitmap-27', 'stale-bitmap-28', 'stale-bitmap-29', 'stale-bitmap-30', 'stale-bitmap-31', 'stale-bitmap-32', 'stale-bitmap-33', 'stale-bitmap-34', 'stale-bitmap-35', 'stale-bitmap-36', 'stale-bitmap-37', 'stale-bitmap-38', 'stale-bitmap-39', 'stale-bitmap-40', 'stale-bitmap-41', 'stale-bitmap-42', 'stale-bitmap-43', 'stale-bitmap-44', 'stale-bitmap-45', 'stale-bitmap-46', 'stale-bitmap-47', 'stale-bitmap-48', 'stale-bitmap-49', 'stale-bitmap-50', 'stale-bitmap-51', 'stale-bitmap-52', 'stale-bitmap-53', 'stale-bitmap-54', 'stale-bitmap-55', 'stale-bitmap-56', 'stale-bitmap-57', 'stale-bitmap-58', 'stale-bitmap-59', 'stale-bitmap-60', 'stale-bitmap-61', 'stale-bitmap-62', 'stale-bitmap-63', 'stale-bitmap-64', 'stale-bitmap-65', 'stale-bitmap-66', 'stale-bitmap-67', 'stale-bitmap-68', 'stale-bitmap-69', 'stale-bitmap-70'] from '/rhev/data-center/mnt/blockSD/a2ad7a23-96c5-4c23-9007-c0d1892ea07d/images/d4e763f5-67d8-4a73-9c3b-79da330032e5/0033c0b7-69f3-43fa-b36c-8c0a518aea84' (bitmaps:144)
2022-12-07 09:49:47,878+0100 INFO  (jsonrpc/2) [storage.bitmaps] Remove bitmap stale-bitmap-1 from '/rhev/data-center/mnt/blockSD/a2ad7a23-96c5-4c23-9007-c0d1892ea07d/images/d4e763f5-67d8-4a73-9c3b-79da330032e5/0033c0b7-69f3-43fa-b36c-8c0a518aea84' (bitmaps:117)
...
```
Vdsm measures expected data:
```
FINISH measure return={'result': {'bitmaps': 0, 'required': 2679767040, 'fully-allocated': 2685009920}} from=::ffff:192.168.122.96,38934, flow_id=13305ae3-0084-4eed-9f5a-ec31a43424ff, task_id=b77b670e-79ae-4cad-8d64-71b6259b83b8 (api:37)
```
And measure finishes correctly, no ERRORs in the logs.

Other disks also have a start `prune_bitmaps`, without the bitmaps removal WARN logs:
```
INFO  (jsonrpc/3) [vdsm.api] START prune_bitmaps(sdUUID='a2ad7a23-96c5-4c23-9007-c0d1892ea07d', imgUUID='c7b5c2f6-87ca-4ba6-a498-e12025f0ab12', volUUID='285874a4-0fc6-4bcf-9afc-696f40dd12bc', baseUUID='7f2f1e6e-d32d-4883-941e-836b77c4dde3') from=::ffff:192.168.122.96,38934, flow_id=13305ae3-0084-4eed-9f5a-ec31a43424ff, task_id=55136c39-aacb-4c02-9d8f-17d93a763b5f (api:31)
...
INFO  (jsonrpc/3) [vdsm.api] FINISH prune_bitmaps return=None from=::ffff:192.168.122.96,38934, flow_id=13305ae3-0084-4eed-9f5a-ec31a43424ff, task_id=55136c39-aacb-4c02-9d8f-17d93a763b5f (api:37)
```

Resulting base measure (bitmaps removed):
```
required size: 2679767040
fully allocated size: 2685009920
bitmaps size: 0
```

Engine logs:
```
INFO  [org.ovirt.engine.core.dal.dbbroker.auditloghandling.AuditLogDirector] (EE-ManagedScheduledExecutorService-engineScheduledThreadPool-Thread-32) [13305ae3-0084-4eed-9f5a-ec31a43424ff] EVENT_ID: USER_REMOVE_SNAPSHOT_FINISHED_SUCCESS(356), Snapshot 'snap 1' deletion for VM 'internal-livemerge-vm' has been completed.
```

Also disappeared in the engine UI, and the task finished successfully.

lib/vdsm/storage/hsm.py
```diff
@@ -34,6 +34,7 @@
 from vdsm.common.time import monotonic_time
```
nirs, line 34: For the commit message:
> - exclusive lock on the base image

Remove `base`, image is the disk including base and top.

lib/vdsm/storage/hsm.py
```diff
@@ -2857,6 +2859,33 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
...
+                pruned if missing or invalid in top.
+        """
+        img_ns = rm.getNamespace(sc.IMAGE_NAMESPACE, sdUUID)
+        locks = [rm.Lock(sc.STORAGE, sdUUID, rm.SHARED),
+                 rm.Lock(img_ns, imgUUID, rm.EXCLUSIVE)]
+        with guarded.context(locks):
+            dom = sdCache.produce_manifest(sdUUID=sdUUID)
+            try:
+                base_vol = dom.produceVolume(imgUUID=imgUUID, volUUID=baseUUID)
+                top_vol = dom.produceVolume(imgUUID=imgUUID, volUUID=volUUID)
```
nirs, line 2882: In produce_manifest() and produceVolume() there no need to use key=value in
such call, both arguments are required, and the key=value style adds no value
because the name of the argument is self documentating.

lib/vdsm/storage/hsm.py
```diff
@@ -2857,6 +2859,33 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
...
+            except se.VolumeDoesNotExist as e:
+                raise exception.expected(e)
```
nirs, line 2884: This is good for public API, meaning that this is not vdsm error but an error of the
caller. But in this internal API it is better to raise to make debugging easier, so
we can remove the try-except block.

lib/vdsm/virt/livemerge.py
```diff
@@ -287,6 +287,14 @@ def merge(self, driveSpec, base, top, bandwidth, job_id):
             except JobExistsError as e:
                 raise exception.MergeFailed(str(e), job=job.id)
 
+            if self._base_needs_prune_bitmaps(drive, job, base_info):
+                # Prune stale bitmaps from base volume before measuring.
```
nirs, line 291: The comment does not add any value, it describe what the code does, which is pretty 
clear.

The only interesting info is "before measuring". This comment can come before the if
and say something like:

    # Prune stale bitmaps if needed once before measuring the image. Current
    # qemu-img ignores bitmaps in base but future version may not.
    if self._base_needs_prune_bitmaps(drive, job, base_info):
        ...

lib/vdsm/virt/livemerge.py
```diff
@@ -402,6 +410,21 @@ def _start_commit(self, drive, job):
             self._untrack_job(job.id)
             raise exception.MergeFailed(str(e), job=job.id)
 
+    def _base_needs_prune_bitmaps(self, drive, job, base_info):
+        if base_info['format'] != 'COW':
+            log.debug("Base volume does not support bitmap operations "
+                      "job=%s drive=%s volume=%s",
+                      job.id, job.drive, job.base)
```
nirs, line 417: Base volume does not support bitmaps. I'm not sure that this log is needed.

lib/vdsm/virt/livemerge.py
```diff
@@ -402,6 +410,21 @@ def _start_commit(self, drive, job):
...
+            return False
+
+        if job.top == drive.volumeID:
```
nirs, line 420: This should use:

    if job.active_commit:

to match other checks like _base_needs_extend().

lib/vdsm/virt/livemerge.py
```diff
@@ -402,6 +410,21 @@ def _start_commit(self, drive, job):
...
+            log.debug("Top volume is the active volume "
+                      "job=%s drive=%s volume=%s",
```
nirs, line 422: It would be nice to unify the way we log job and base. You copied the style use in _base_needs_extend() but in other code we log "job_id=%r", "base=%r" or "(job: %s)".

I think "job=%r base=%r" here will be better and later someone will unify the logs.

tests/virt/vmfakelib.py
```diff
@@ -53,6 +54,9 @@ def inappropriateDevices(self, ident):
 
     @recorded
     def prune_bitmaps(self, sdUUID, imgUUID, volUUID, baseUUID):
+        if "prune_bitmaps" in self.errors:
+            error = self.errors["prune_bitmaps"]
+            return response.error_raw(error.code, error.msg)
```
nirs, line 59: We can simplify this to:

    error = self.errors.get("prune_bitmaps")
    if error:
        raise ...

This is shorter and does only one dict lookup.

nirs (commented):
Looks good, see comments for minor logs, comments, and style improvements.

lib/vdsm/virt/livemerge.py
aesteve-rh: Ah, I see, it stores the same condition in the `Job` class. Good. I'll use it.

lib/vdsm/virt/livemerge.py
aesteve-rh: How about this for both debug logs (regarding text & style)?

```python
    def _base_needs_prune_bitmaps(self, job, base_info):
        if base_info['format'] != 'COW':
            log.debug("Base volume does not support bitmaps, "
                      "job=%r base=%r", job.id, job.base)
            return False

        if job.active_commit:
            log.debug("Top volume is the active volume, "
                      "job=%r top=%s", job.id, job.top)
            return False
```

lib/vdsm/storage/hsm.py
```diff
@@ -2857,6 +2859,30 @@ def measure(self, sdUUID, imgUUID, volUUID, dest_format, backing=True,
...
+            imgUUID (str): The UUID of the image contained on the volume.
+            volUUID (str): The UUID of the top volume.
+            baseUUID (str): The UUID of the volume that will get its bitmaps
+                pruned if missing or invalid in top.
+        """
+        img_ns = rm.getNamespace(sc.IMAGE_NAMESPACE, sdUUID)
+        locks = [rm.Lock(sc.STORAGE, sdUUID, rm.SHARED),
+                 rm.Lock(img_ns, imgUUID, rm.EXCLUSIVE)]
+        with guarded.context(locks):
+            dom = sdCache.produce_manifest(sdUUID=sdUUID)
```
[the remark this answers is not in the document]
aesteve-rh: oops :)

nirs (approved):

PR #363, opened 2022-12-19
docker: create venv for tests

aesteve-rh:
In vdsm-test containers, create a
new venv [1] at /venv instead of installing
the required packages directly in the
root environment. Installing packages in
the root environment is strongly discouraged,
as it can run into conflicts with the system
package manager.

[1] - https://pip.pypa.io/warnings/venv

### DIFF
--- a/docker/Dockerfile.alma-9
+++ b/docker/Dockerfile.alma-9
@@ -74,7 +74,8 @@ RUN debuginfo-install -y python3 \
 # Add development packages
 COPY requirements.txt requirements.txt
 RUN python3 -m pip install --upgrade pip \
-    && python3 -m pip install --requirement requirements.txt
+    && python3 -m venv /venv \
+    && /venv/bin/python3 -m pip install --requirement requirements.txt
 
 # Add lvm configuration.
 COPY lvmlocal.conf /etc/lvm/
--- a/docker/Dockerfile.centos-8
+++ b/docker/Dockerfile.centos-8
@@ -73,7 +73,8 @@ RUN debuginfo-install -y python3 \
 # Add development packages
 COPY requirements.txt requirements.txt
 RUN python3 -m pip install --upgrade pip \
-    && python3 -m pip install --requirement requirements.txt
+    && python3 -m venv /venv \
+    && /venv/bin/python3 -m pip install --requirement requirements.txt
 
 # Add lvm configuration.
 COPY lvmlocal.conf /etc/lvm/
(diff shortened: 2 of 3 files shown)


aesteve-rh:
Example of a conflict with the system package manager: https://github.com/oVirt/vdsm/actions/runs/3731464902/jobs/6329726583

docker/Dockerfile.alma-9
```diff
@@ -74,7 +74,8 @@ RUN debuginfo-install -y python3 \
 # Add development packages
 COPY requirements.txt requirements.txt
 RUN python3 -m pip install --upgrade pip \
-    && python3 -m pip install --requirement requirements.txt
+    && python3 -m venv /venv \
```
tinez, line 77: Will this env be functional without `--system-site-packages`? I think we need that to access the deps that we install as RPMs.

aesteve-rh: We do not need to do `--system-site-packages` in the local test venv. This shouldn't be different, no?

I tested locally with:
```
podman run \
    --env-host --privileged --rm -it \
    --volume `pwd`:/vdsm:Z \
    localhost/vdsm-test:centos-8 \
        bash -c "cd /vdsm && source /venv/bin/activate && ./ci/tests.sh"
```
And the results were mostly good. A few tests failed, but mostly due to environment issues (as the loop devices, or network stuff, or systemctl tests), that would fail also with current containers in local environment:
```
...
common/systemctl_test.py FFFF
...
lib/rpc/http_image_request_handler_test.py ..............127.0.0.1 - - [20/Dec/2022 08:31:58] "my-image-data" 400 
...
network/integration/netinfo_test.py ...s......E
network/integration/netlink_test.py ............                                                                                                                                    
network/integration/tc_test.py ........FFx
...

ERROR:   tests: commands failed
ERROR:   lib: commands failed
ERROR:   network: commands failed
  virt: commands succeeded
  gluster: commands succeeded
  hooks: commands succeeded
```
But no complaint about missing packages.

Same run with `quay.io/ovirt/vdsm-test-centos-8:latest` (and without venv activation):
```
....
ERROR:   tests: commands failed
ERROR:   lib: commands failed
ERROR:   network: commands failed
  virt: commands succeeded
  gluster: commands succeeded
  hooks: commands succeeded
```
Same results.

aesteve-rh: I think it is probably `tox` finding the system packages.

tinez: Right, this is possible.

tinez (approved):

mz-pdm (approved):
Makes sense and if it works then OK.

PR #366, opened 2022-12-20
ci: use new vdsm-test container

aesteve-rh:
Update the cis workflow to use the new vdsm-test
container at quay.io/ovirt/vdsm-test:<distro>.

Modify the test scripts to activate the venv
before running the tests, otherwise the tests
will have missing dependencies that are only
installed in the venv in the new containers.

### DIFF
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -3,6 +3,9 @@
 # SPDX-FileCopyrightText: Red Hat, Inc.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+# Activate the tests venv (for containers only)
+[ -d /venv ] && source /venv/bin/activate
+
 ./autogen.sh --system
 make
 make lint
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -3,8 +3,15 @@
 # SPDX-FileCopyrightText: Red Hat, Inc.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-# Activate the tests venv (for containers only)
-[ -d /venv ] && source /venv/bin/activate
+# Only when running in a container
+[ -d /venv ] && {
+    # Workaround to avoid this warning:
+    #   fatal: detected dubious ownership in repository at '/dir'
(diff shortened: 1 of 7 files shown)


ci/lint.sh
```diff
@@ -3,6 +3,9 @@
 # SPDX-FileCopyrightText: Red Hat, Inc.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+# Activate the tests venv
+source /venv/bin/activate
```
tinez, line 7: Where is that venv created?

aesteve-rh: Here:
https://github.com/oVirt/vdsm/blob/master/docker/Dockerfile.centos-9#L77

But it is actually passing with the modification from this PR (which should be merged first):
https://github.com/oVirt/vdsm/pull/367

Includes the system packages in the venv.

tinez: Thanks, I missed. Still, I think these shell scripts are (or at least were) useful outside of the container envs. How about we use `[ -d /venv ] && source /venv/bin/activate` with a comment that this is for virtual envs in container images?

aesteve-rh: Fair point. I'll add that.

These scripts are in the `ci` folder and they finally rely on the `make` target, so outside the CIs they do not necessarily need to be used. But if they "could" be used before, I think we should keep that in the new approach.

tinez (dismissed):

ci/lint.sh
```diff
@@ -3,6 +3,13 @@
 # SPDX-FileCopyrightText: Red Hat, Inc.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
+# Workaround to avoid this warning:
+#   fatal: detected dubious ownership in repository at '/dir'
+git config --global --add safe.directory "$(pwd)"
```
tinez, line 8: This is another thing I think we should do only inside container env, so I propose:
```bash
[ -d /venv ] && {
    git config ...    
    source /venv/bin/activate
}
```

tinez (approved):

PR #369, opened 2023-01-17
Fix flapping disk info due to timing differences

dupondje:
In commit 45903d01e142047093bf844628b5d90df12b6ffb there was code added to refresh the fs info earlier when a hotplug disk was added. Now this caused some side effect that when only that command was triggered, the info dict never contained 'disk.count', which caused us to change the diskMapping to the parsed data from fsinfo.

Some minutes later when the disk command needed to be refreshed, it changed the diskMapping again, because (sometimes) the disks call returns more disks in comparision to the fsinfo call (for ex unmounted disks).

Again a bit later, the fsinfo had to be refreshed, and the diskMapping was modified again.

This occurs continiously, causing the device hash to change every time, and causing an excessive amount of dumpxml's also due to that.

Fixed this by not setting the diskMapping when we have the _QEMU_DISKS_COMMAND in our caps. As this means somewhere else in time this command will refresh the disk info.
Also refresh the disk info and not only fs info after a disk hotplug.

Signed-off-by: Jean-Louis Dupond <jean-louis@dupond.be>
Change-Id: I27a30e020a5221237d9d4ae7835a1cc4e69c0c59

### DIFF
--- a/lib/vdsm/virt/qemuguestagent.py
+++ b/lib/vdsm/virt/qemuguestagent.py
@@ -460,7 +460,8 @@ class QemuGuestAgentPoller(object):
                 if _QEMU_COMMANDS[command] not in caps['commands']:
                     continue
                 after_hotplug = \
-                    command == VIR_DOMAIN_GUEST_INFO_FILESYSTEM and \
+                    (command == VIR_DOMAIN_GUEST_INFO_FILESYSTEM or
+                     command == VIR_DOMAIN_GUEST_INFO_DISKS) and \
                     vm_obj.last_disk_hotplug() is not None and \
                     (now - vm_obj.last_disk_hotplug() >=
                         _HOTPLUG_CHECK_PERIOD) and \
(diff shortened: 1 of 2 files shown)


dupondje:
@nyoxi don't know if you want to check this also :)

mz-pdm (commented):
@nyoxi, any objections to merging this?  can anything get wrong?

mz-pdm (commented):
/ost

nyoxi (dismissed):

nyoxi:
Looks good to me.

dupondje:
Fixed the test and lint :)

mz-pdm:
@dupondje, please fix the problems reported by the CI.

mz-pdm:
Ah, you were faster, great. :-)

mz-pdm (approved):

mz-pdm:
Thanks for the fix!

PR #377, opened 2023-02-14
network: tests - use make to build nmstate

erav:
Building nmstate from source for the tests now requires using `make rpm` instead of pip install.

Change-Id: I46bf7fa6acf29da41ed313252aa54a64c6b24ab1


dupondje:
Missing a bunch of stuff, like build-deps for the make rpm stage and doesn't install the build rpm's.
Will be superseded once needed.

PR #378, opened 2023-02-17
When calling the command 'vdsm-client Host getAllVmStats', sometimes value of cpuUser is coming negative.

saksham-oracle:
When calling the command 'vdsm-client Host getAllVmStats', sometimes value of cpuUser was coming negative.

Fix:

The cpuUser value was coming negative due to less accuracy of user_time and system_time values (upto 2 decimal) as compared to cpu_time values (upto 9 decimal) returned by libvirt. Fixed the issue by adding the check in vmstats.py for negative cpuUser values and setting them to 0.

Signed-off-by: Saksham Srivastava saksham.sa.srivastava@oracle.com


lib/vdsm/virt/vmstats.py
```diff
@@ -130,6 +130,7 @@ def cpu(stats, first_sample, last_sample, interval):
                 ((last_sample['cpu.time'] - first_sample['cpu.time']) -
                  cpu_sys),
                 interval)
+            stats['cpuUser'] = 0.00 if stats['cpuUser'] < 0 else stats['cpuUser']
```
mz-pdm, line 133: Thank you for the fix.  I'd suggest to polish it a bit:

- Wouldn't it be better to handle it in `_usage_percentage`?  It would also handle a similar case for `stats['cpuSys']` if it can ever happen.
- It would be good to add an explanatory source code comment.
- I think it's better to check whether the value is close to zero or clearly negative and log a warning in the latter case, to not hide cases where the value would be clearly wrong.
- Using `0.0` rather than `0.00` would be consistent with the surrounding code.

mz-pdm (changes requested):

saksham-oracle:
Can someone review this? I have made the suggested changes.

saksham-oracle:
Made changes for the failed checks too:

```
lib/vdsm/virt/vmstats.py:134:80: E501 line too long (81 > 79 characters)
            stats['cpuUser'] = 0.00 if stats['cpuUser'] < 0 else stats['cpuUser']
```

lib/vdsm/virt/vmstats.py
mz-pdm: Could you please respond to the points above? You're free to disagree with them but we need to come to some conclusion in order to proceed.
As for the source code comment, it should explain, rather than the obvious thing of avoiding the negative value, why the value can become negative (something in the sense of the fix description you provided in this PR).
Also, when changing commits inside a Vdsm PR, we use force push rather than adding additional commits, so please squash the commits.

saksham-oracle: Sorry, I had pasted the answers earlier too but they were in pending state.

saksham-oracle: Also made the suggested changes in the comment and squashed the two commits.

mz-pdm: Thank you for clarification and improving the comment, the example was helpful.
One last thing to clarify is the "0.00" value: Since it's a float and not a string there, I fail to understand how the notation can influence the output. For example, in the Python interpreter prompt:

```
>>> x = 0.00
>>> x
0.0
```

What do I miss?

saksham-oracle: Yes, thanks. I checked again, variable type is float. Made the change to "0.0".

saksham-oracle:
Removed the trailing spaces that caused the test failure last time.

mz-pdm:
Great, looks good now and CI passes. The last thing to do is to add a proper header line to the commit message. See other commits in Vdsm repository how the first lines of commit messages look like, something like "virt: Fix negative cpuUsed in getAllVmStats" could fit here.

saksham-oracle:
Thanks, made the change.

mz-pdm:
Ah, I meant the header line of the commit message, not of the PR.

saksham-oracle:
Made the change to commit message.

mz-pdm (approved):

mz-pdm:
Hm, CI is not in a very good mood today: shaky tests, OST failing, ... @michalskrivanek what can we do about it?

michalskrivanek:
el8stream doesn't work due to ansible 2.14 currently. try el9stream

michalskrivanek (approved):
re-approving after rebase

PR #380, opened 2023-02-21
volume: allow extend cow preallocated

aesteve-rh:
Backport for https://github.com/oVirt/vdsm/pull/379

Currently we assume in the code that cow
volumes do not need to be extended in any case.
However, all preallocated volumes, both
cow and raw, should be extended when requested.
Otherwise, cow preallocated volumes have bigger
capacity than their truesize. Thus, when the size
of the disk reaches the truesize, the VM will
pause, which is an undesired effect.

To avoid this, we relax this assumption so that
only cow sparse volumes are skipped on extend calls.

Bug-Url: https://bugzilla.redhat.com/2170689

### DIFF
--- a/lib/vdsm/storage/blockVolume.py
+++ b/lib/vdsm/storage/blockVolume.py
@@ -718,7 +718,7 @@ class BlockVolume(volume.Volume):
     def getMetaSlot(self):
         return self._manifest.getMetaSlot()
 
-    def _extendSizeRaw(self, new_capacity):
+    def _extendSize(self, new_capacity):
         # Since this method relies on lvm.extendLV (lvextend) when the
         # requested size is equal or smaller than the current size, the
         # request is siliently ignored.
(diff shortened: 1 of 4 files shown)


nirs:
@aesteve-rh please fix the pr and commit message to describe only the
issue in the bug - preallocated volume are not preallocated during extend.

You can add a comment like:

    (commit message was updated to to reflect the actual fix better)

ahadas (approved):

ahadas:
> @aesteve-rh please fix the pr and commit message to describe only the issue in the bug - preallocated volume are not preallocated during extend.
> 
> You can add a comment like:
> 
> ```
> (commit message was updated to to reflect the actual fix better) 
> ```

+1
my approval was on the code changes

nirs (approved):

ahadas (approved):

PR #386, opened 2023-03-21
ci: Mark the current git directory as safe

mz-pdm:
This is necessary to prevent

  fatal: detected dubious ownership in repository at '/__w/vdsm/vdsm'

messages in the CentOS Stream 9 container and, more importantly, making wrong rpm release version there when building rpm’s.  Without this, the determined release version on CI is always ‘1’, which results in not installing the required set of rpm’s in OST and an OST failure.


michalskrivanek:
"--define="vdsm_release 202303211349.gitebda4d6d9"

yep, works better

PR #387, opened 2023-03-24
managedvolume: add support for local device

WanzenBug:
"local" connections are used by storage systems that directly set up devices on vdsm nodes, such as linstor[1]

[1]: https://docs.openstack.org/cinder/latest/configuration/block-storage/drivers/linstor-driver.html

### DIFF
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -49,6 +49,9 @@ SUPPORTED_DRIVERS = (
 
     # Tested by Muli Ben-Yehuda <info (at) lightbitslabs.com>
     "lightos",
+
+    # Tested by Moritz Wanzenböck <technik+ovirt (at) linbit.com>
+    "local",
 )
 
 log = logging.getLogger("storage.managedvolume")


aesteve-rh:
Thanks for the contribution!

Could you please briefly describe how did you test the update?

WanzenBug:
> Could you please briefly describe how did you test the update?

I used a local oVirt setup (Version 4.5.4), consisting of 4 nodes. I enabled the Managed Block Storage preview, and completed the necessary set up for using [linstor](https://docs.openstack.org/cinder/latest/configuration/block-storage/drivers/linstor-driver.html), which runs a storage agent on every node. The driver can use either `iSCSI` transport, or it creates a DRBD device directly on the node, for which it then uses the `local` connection driver.

Then, after making some changes (the one you see here to vdsm, and some changes to the linstor driver):
* I created a volume
* I attached it to a VM
* I verified (using `virsh dumpxml`) that the volume attached is actually using the expected device
* Migrated the VM to another host
* Stopped and started the VM
* Created a snapshot

aesteve-rh (approved):

aesteve-rh:
Thanks again for contributing :)

PR #393, opened 2023-06-12
Add list_bitmaps command

dupondje:
Add list_bitmaps command to list bitmaps in the qcow2 image.

### DIFF
--- a/lib/vdsm/API.py
+++ b/lib/vdsm/API.py
@@ -1757,6 +1757,9 @@ class SDM(APIBase):
     def clear_bitmaps(self, job_id, vol_info):
         return self._irs.sdm_clear_bitmaps(job_id, vol_info)
 
+    def list_bitmaps(self, job_id, vol_info):
+        return self._irs.sdm_list_bitmaps(job_id, vol_info)
+
 
 class Lease(APIBase):
     ctorArgs = []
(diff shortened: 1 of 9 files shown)


dupondje:
This is still WIP, but having some questions here:

- There is no way to have a SDM command/job return something else than just the status?
- Or we want to extend SDM jobs to also be able to return a response?
- It might even be better/cleaner to extend getQemuImageInfo to return also bitmaps?
- Or just create the list_bitmaps command like getQemuImageInfo (without using SDM)

bennyz:
> This is still WIP, but having some questions here:
> 
>     * There is no way to have a SDM command/job return something else than just the status?
> 
>     * Or we want to extend SDM jobs to also be able to return a response?

This is intended for somewhat longer async operations, so not really

>     * It might even be better/cleaner to extend getQemuImageInfo to return also bitmaps?
> 
>     * Or just create the list_bitmaps command like getQemuImageInfo (without using SDM)

This sounds like a better approach because this is a sync operation to perform a validation, I'll let vdsm maintainers decide

aesteve-rh:
Sorry for the delay, but was out for a few days. And I wanted to check the code before giving an answer.

I agree with @bennyz. Both extend `getQemuImageInfo` or adding a new command are fit solutions here. Also, they will not interfere with other flows.

Thanks for the update!

dupondje:
Let's merge https://github.com/oVirt/vdsm/pull/394 then. I fix the ovirt-engine code also then so it read's the args and uses them for validation.

PR #402, opened 2023-10-26
virt: v2v: parse progress from new virt-v2v

nyoxi:
Progress generated by virt-v2v has changed and is more graphical. This breaks the parser and causes the import to abort even when there is no real problem and virt-v2 completes successfully. Update the parsing logic to handle both old and new progress format.


nyoxi:
Confirmed by a user on mailing list that the progress is now updated properly. From vdsm.log:

```
2023-10-26 16:04:15,862-0500 INFO  (v2v/4a6237d9) [root] Job '4a6237d9-05a0-4d5a-a7b1-ff01764a761f' copy disk 1 progress 80/100 (v2v:900)
[...]
2023-10-26 16:04:15,896-0500 INFO  (v2v/4a6237d9) [root] Job '4a6237d9-05a0-4d5a-a7b1-ff01764a761f' copy disk 1 progress 90/100 (v2v:900)
[...]
2023-10-26 16:04:17,437-0500 INFO  (v2v/4a6237d9) [root] Job '4a6237d9-05a0-4d5a-a7b1-ff01764a761f' copy disk 1 progress 100/100 (v2v:900)
2023-10-26 16:04:17,608-0500 INFO  (v2v/4a6237d9) [root] Job '4a6237d9-05a0-4d5a-a7b1-ff01764a761f' finished import successfully (v2v:875)
```

michalskrivanek:
LGTM

lib/vdsm/v2v.py
```diff
@@ -970,8 +970,9 @@ def _parse_progress(self, chunk):
         m = self.DISK_PROGRESS_RE.match(chunk)
         if m is None:
             return None
+        value = [x for x in m.groups() if x is not None][0]
```
mz-pdm, line 973: Is there any risk of this failing because the regexp wouldn't match? Previously, it was part of the following try block, now maybe it could fail with IndexError rather than OutputParserError?

mz-pdm (commented):
Well, it would be best if virt-v2v had an option for managing the output, but I agree this patch is the simplest solution now. One question about the patch inside.

lib/vdsm/v2v.py
nyoxi: No, that's a misunderstanding. If the regex does not match it is  caught by the `if m is None` statement above and not the `try` block below. This is still true after the change. Also, the way the regex is defined there is no change that the regex would match but all the groups were None -- exactly one is always not None.

mz-pdm: Ah, right, sorry.

mz-pdm (approved):

PR #405, opened 2024-02-01
Fix default route with libnl3.8+

dupondje:
Since libnl3.8, he nl_addr_parse does not return 'none' anymore for 'default', but it returns a 0.0.0.0/0 addr.
https://github.com/thom311/libnl/commit/0c0aee829c1d2d468d89eda75d38e7490e4de575

But since we depend on that value in vdsm, it never reported a default gateway anymore, and caused networks to be out of sync.

So next to 'none' we also check for '0.0.0.0/0' or '::/0'.


aesteve-rh (changes requested):

lib/vdsm/network/netinfo/routes.py
```diff
@@ -53,7 +53,11 @@ def is_default_route(gateway, routes):
             and route['family'] == 'inet'
             and route['scope'] == 'global'
             and route['gateway'] == gateway
-            and route['destination'] == 'none'
+            and (
+                route['destination'] == 'none'
+                or route['destination'] == '0.0.0.0/0'
+                or route['destination'] == '::/0'
```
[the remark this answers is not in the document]
dupondje: Fixed! :)

lib/vdsm/network/netinfo/routes.py
```diff
@@ -83,7 +83,7 @@ def get_gateway(
     gateways = [
         r
         for r in routes
-        if r['destination'] == 'none'
+        if r['destination'] in ('none', '0.0.0.0/0', '::/0')
```
aesteve-rh, line 86: nit: Using sets specifically (i.e., with curly braces) seems to be more performant. Not sure how often this code is executed, so feel free to ignore the comment, but for the sake of experimentation, I did try:

```python
from timeit import timeit

dest = 'some'
print(timeit(lambda: dest in ['none', '0.0.0.0/0', '::/0'], number=10000000))
print(timeit(lambda: dest in {'none', '0.0.0.0/0', '::/0'}, number=10000000))
print(timeit(lambda: dest in ('none', '0.0.0.0/0', '::/0'), number=10000000))
print(timeit(lambda: dest == 'none' or dest == '0.0.0.0/0' or dest == '::/0', number=10000000))
```
Result:
```
8.609254561990383
5.435367603000486
7.046174183997209
9.278653385001235
```

dupondje: Afaik only on network sync and update of capabilities.
I'm happy to change it, but seems like we use ( ) everywhere in the code, and nowhere { }

aesteve-rh (approved):

lib/vdsm/network/netinfo/routes.py
almusil: I'm more for consistency than this optimization that most likely doesn't have any impact overall.

almusil (approved):
Thanks!

PR #408, opened 2024-02-14
vdsm: get gluster volume info from any gluster peer

josgutie:
The function _get_gluster_volinfo query the glusterfs volume info the the storage server, this is translated to the gluster client adding the parameter --remote-host which limits the query to one server, so we are converting the storage server as a single point of failure, if it is not available, it can led to cluster outtage.
The proposed changed let the cluster cli to use any available gluster peer.

### DIFF
--- a/lib/vdsm/taskset.py
+++ b/lib/vdsm/taskset.py
@@ -91,10 +91,13 @@ def cpulist_parse(cpu_range):
     or the output of the 'taskset' and 'lscpu' tools.
     """
     cpus = []
+    excluded_cpus = []
     for item in cpu_range.split(','):
         if '-' in item:
             begin, end = item.split('-', 1)
             cpus.extend(range(int(begin), int(end) + 1))
+        elif item.startswith("^"):
+            excluded_cpus.append(int(item[1:]))
         else:
             cpus.append(int(item))
-    return frozenset(cpus)
+    return frozenset(cpus) - frozenset(excluded_cpus)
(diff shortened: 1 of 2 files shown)


josgutie (commented):
Let gluster cli to query gluster volume information from any available peer.

lib/vdsm/storage/storageServer.py
```diff
@@ -384,8 +384,7 @@ def _get_backup_servers_option(self):
     def _get_gluster_volinfo(self):
         try:
             superVdsmProxy = supervdsm.getProxy()
-            volinfo = superVdsmProxy.glusterVolumeInfo(self._volname,
-                                                       self._volfileserver)
```
nirs, line 388: I think this was added years ago because not specifying the server could be a problem. Not sure that the gluster cluster can cope with a case when one of the server is down.

Please check git history to understand when and why self._volfileserver was added.

Also how did you test this change? Please describe what was tested - how you simulated the issue and tested that this change solves the issue, and how you tested that this change does not cause regressions in other flows.

nirs: I found this test:
https://github.com/oVirt/vdsm/blob/master/tests/storage/storageserver_test.py#L268

Using `backup-vol-fileserver`. Maybe the issue you experience can be solved with this configuration option?

Shwetha-Acharya: @mohit84 do you have any suggestions?

PR #409, opened 2024-02-28
Fix the cpu pinning parsing logic to parse the format appropriately and exclude cpu from pinning list

shubhaOracle:
Issue: The CPU pinning tool tip says that to exclude a cpu from pinning, it needs to provided in the following format.
0#10-13_1#25-27,^26

With this format CPU 26 will be excluded from pinning. However, vdsm code fails to parse the above string resulting in VM start failure.

Fix: Fix the cpu pinning parsing logic to parse the format appropriately and exclude cpu from pinning list.

Signed-off-by: ShubhaOracle <Shubha.kulkarni@oracle.com>

### DIFF
--- a/lib/vdsm/taskset.py
+++ b/lib/vdsm/taskset.py
@@ -91,10 +91,13 @@ def cpulist_parse(cpu_range):
     or the output of the 'taskset' and 'lscpu' tools.
     """
     cpus = []
+    excluded_cpus = []
     for item in cpu_range.split(','):
         if '-' in item:
             begin, end = item.split('-', 1)
             cpus.extend(range(int(begin), int(end) + 1))
+        elif item.startswith("^"):
+            excluded_cpus.append(int(item[1:]))
         else:
             cpus.append(int(item))
-    return frozenset(cpus)
+    return frozenset(cpus) - frozenset(excluded_cpus)
--- a/tests/virt/domaindescriptor_test.py
+++ b/tests/virt/domaindescriptor_test.py
@@ -112,6 +112,7 @@ PINNED_CPUS = """
     <cputune>
         <vcpupin vcpu="0" cpuset="1,2,5-7" />
         <vcpupin vcpu="1" cpuset="1,6,10" />
+        <vcpupin vcpu="2" cpuset="1-4,^3,6" />
     </cputune>
 </domain>
 """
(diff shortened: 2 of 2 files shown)


lib/vdsm/taskset.py
```diff
@@ -95,6 +95,9 @@ def cpulist_parse(cpu_range):
         if '-' in item:
             begin, end = item.split('-', 1)
             cpus.extend(range(int(begin), int(end) + 1))
+        elif '^' in item:
```
tinez, line 98: The format for exclusion is `^NUMBER`, so maybe it makes more sense to use `item.startswith("^")` here. Then we can try converting the number with `int(item[1:])`.

Another thing is that almost every example you will find on those strings will have the CPUs listed in a sorted manner, but I don't think there's a strict requirement for that (e.g. `taskset -cp 2,0 MYPID` will work fine). That means, that if someone uses a string like `^26,25-27`, the algorithm won't work. It would be better to keep the exluded CPUs in a separate list and return `frozenset(cpus) - frozenset(excluded_cpus)`.

tinez: We have tests for this function in `domaindescriptor_test.py`, please add some test cases to cover the new logic.

shubhaOracle: Updated the code and the test.

aesteve-rh:
Please squash both commits into a single one. In general, changes addressing reviews shall be made into the relevant commit, not adding new commits, to keep the git history clean.

Thanks for your contribution!

shubhaOracle:
Updated the branch. Can we squash commits when merging the changes?

shubhaOracle:
I tried to follow the instructions to rebase and squash. While the rebase part seems to have worked, the squash commit can only work with adjacent commit. Pl see the screenshot. The commits need to squash are 37fe02548 and 8f12263bc. Pl suggest alternative.
image

tinez:
I've made a nice tutorial for you, enjoy :) https://asciinema.org/a/9QlEBnXyUlEBIJ2q1putVOoZC

shubhaOracle:
@tinez Thank you so much for the tutorial!!!  I have made the changes now.

tinez (approved):

tinez:
yw :)

tinez (changes requested):
We still need to address the failing assertion.

shubhaOracle:
Made fix  to address the failing assertion.

tinez (approved):

mz-pdm:
The patch itself looks good to me, but I'd suggest:

- Improving the commit message. It's important to have good and descriptive commit messages because they are the primary source of information about the changes applied when examining them in future. In this case, I would change the commit title to something like `virt: Add support for excluding CPUs in CPU list parsing`, which fits into a single line easily. And would add the reasoning, basically this PR description, to the commit message body.
- Using negative ranges, e.g. `1-10,^4-6`, is not supported, which is fine but is worth to mention in the commit message.
- It may be worth to mention the exclusion syntax in `cpulist_parse` docstring.

shubhaOracle:
Updated the commit message.

mz-pdm (approved):

PR #410, opened 2024-03-01
fix_Flake8_Rules_E713

gorsing:
While the difference between X not in Y and not X in Y may be negligible from a performance standpoint, using a more direct form can in some cases make it easier for the code optimizer in the Python interpreter.


mz-pdm (approved):

tinez (approved):

gorsing:
Thanks for your time @tinez @mz-pdm

mz-pdm:
Thanks for your patch. :)

PR #412, opened 2024-03-20
vdsm: get gluster volume info from any gluster peer

josgutie:
The function _get_gluster_volinfo query the glusterfs volume info the the storage server, this is translated to the gluster client adding the parameter --remote-host which limits the query to one server, so we are converting the storage server as a single point of failure, if it is not available, it can led to cluster outtage. The proposed changed let the cluster cli to use any available gluster peer.

### DIFF
--- a/lib/vdsm/storage/storageServer.py
+++ b/lib/vdsm/storage/storageServer.py
@@ -387,9 +387,21 @@ class GlusterFSConnection(MountConnection):
             volinfo = superVdsmProxy.glusterVolumeInfo(self._volname,
                                                        self._volfileserver)
             return volinfo[self._volname]
-        except ge.GlusterCmdExecFailedException as e:
-            log.warning("Failed to get volume info: %s", e)
-            return {}
+        except ge.GlusterException as e:
+            # The remote host may be down.
+            # If we are running on a hyperconverged system the gluster client
+            # can use one of the other gluster servers.
+            log.info("Failed to get volume info from remote server %s: %s",
+                     self._volfileserver, e)
+            log.debug("Trying to get volume info from backup servers: %s",
+                      self._options)
+            try:
+                volinfo = superVdsmProxy.glusterVolumeInfo(self._volname)
+                return volinfo[self._volname]
+            except ge.GlusterException as e:
+                log.warning(
+                    "Failed to get volume info from backup servers: %s", e)
+        return {}
 
 
 class NFSConnection(Connection):
--- a/tests/storage/storageserver_test.py
+++ b/tests/storage/storageserver_test.py
@@ -250,6 +250,40 @@ class TestGlusterFSConnection:
         gluster = GlusterFSConnection(id="id", spec="192.168.122.1:/music")
         assert gluster.options == ""
 
+    @pytest.mark.parametrize("userMountOptions", [
+        '',
+        'backup-volfile-servers=192.168.122.1:192.168.122.2',
+    ])
+    def test_glusterfs_retry_withotut_volfile_server(self, monkeypatch,
+                                                     userMountOptions):
+        """
+        The test will fail if called with volfileserver, simulating the case
+        when a remote server is down. If the caller try again without
+        volfileserver the call will succeed, simulating the case when running
+        on a hyperconverged system when the local glsuter glsuter client can
+        use one of the connected servers.
+        """
+        monkeypatch.setattr(storageServer, 'supervdsm', FakeSupervdsm())
+        monkeypatch.setattr(gluster_cli, 'exists', lambda: True)
+
+        def glusterVolumeInfo(volname=None, volfileServer=None):
+            assert volname == "music"
+            if volfileServer is not None:
+                raise ge.GlusterException()
+            return {'music': {'brickCount': '2',
+                              'bricks': ['192.168.122.1:/tmp/music',
+                                         '192.168.122.2:/tmp/music']
+                              }
+                    }
+        storageServer.supervdsm.glusterVolumeInfo = glusterVolumeInfo
+        gluster = GlusterFSConnection(id="id", spec="192.168.122.3:/music",
+                                      options=userMountOptions)
+        expected_volinfo = {'brickCount': '2',
+                            'bricks': ['192.168.122.1:/tmp/music',
+                                       '192.168.122.2:/tmp/music']
+                            }
+        assert gluster.volinfo == expected_volinfo
+
 
 class TestGlusterFSNotAccessibleConnection:
 


josgutie:
> Why did you send a new PR when we have #408?
> 
> You need to answer my questions in #408. Please close this PR so we can continue the review on the existing PR.

Because the PR was bad formed, and I guess that it's more clean this PR.

Here is your question in the other PR, 

I think this was added years ago because not specifying the server could be a problem. Not sure that the gluster cluster can cope with a case when one of the server is down.

Please check git history to understand when and why self._volfileserver was added.

Also how did you test this change? Please describe what was tested - how you simulated the issue and tested that this change solves the issue, and how you tested that this change does not cause regressions in other flows.

Regarding the tests, I will describe in a new comment if you don't mind.

nirs:
@josgutie ok lets continue here, but please do not close PRs and submit new one like this. If there was an issue with the previous PR it can be easily fixed by pushing a new version.

josgutie:
- Environment:
- RHHI:
  - 1.8
  - ovirt-engine:
    - 4.5.3.10
  - RHVH:
    -  4.5.3.10
    - vdsm:
      -   4.50.3.9-1
    - vdsm-gluster:
      - 4.50.3.9-1
    - glusterfs:
      - 6.0-63
- Gluster Storage domains configuration:
###
$ /usr/share/ovirt-engine/dbscripts/engine-psql.sh -c "SELECT connection, mount_options FROM storage_server_connections;"
          connection           |                          mount_options                           
-------------------------------+------------------------------------------------------------------
 rhhi01.storage01.lab:/data    | backup-volfile-servers=rhhi01.storage01.lab:rhhi02.storage01.lab
 rhhi03.storage01.lab:/engine  | backup-volfile-servers=rhhi01.storage01.lab:rhhi02.storage01.lab
 rhhi03.storage01.lab:/vmstore | backup-volfile-servers=rhhi01.storage01.lab:rhhi02.storage01.lab

$ grep -E "storage|mnt_options" /etc/ovirt-hosted-engine/hosted-engine.conf 
storage=rhhi03.storage01.lab:/engine
mnt_options=backup-volfile-servers=rhhi01.storage01.lab:rhhi02.storage01.lab

$ df -h | grep glusterSD
rhhi03.storage01.lab:/engine                     100G   14G   87G  14% /rhev/data-center/mnt/glusterSD/rhhi03.storage01.lab:_engine
rhhi03.storage01.lab:/vmstore                    150G  2.6G  148G   2% /rhev/data-center/mnt/glusterSD/rhhi03.storage01.lab:_vmstore
rhhi01.storage01.lab:/data                        50G  908M   50G   2% /rhev/data-center/mnt/glusterSD/rhhi01.storage01.lab:_data

###

josgutie:
Issue:
  - Put on maintenance mode `data` storage domain.
  - Shutdown rhhi01 host.
  - `data` storage domain can't be activated.
 - Here are the tests completed and their result after applying the patch.
   - Scenario of the issue:
     - SD `data`can be activated.
   - Environment power off:
     -  Host rhhi01 is powered of, rest are active.
     - Shutdown managed by the ansible role shutdown_env.
     - Result:
       -  Shutdown without any issues.
      - Environment power up
        - Only two hosts started, rhhi02 and rhhi03.
        - Result:
          - All Storage Domains are active
          - Cluster and Data Center active.

 @nirs suggested to test a new deployment, it will take some time, but I guess that I could check the scenario.

nirs:
> @nirs suggested testing a new deployment, it will take some time, but I guess that I could check the scenario.

Test looks good, and sure we want to test the common operations like creating a new storage domain, putting storage domain to maintenance and activating it.

But I think you change is not complete - it assumes that `backup-volfile-servers=` is configured with all the nodes. In this case it seems that gluster does use any of the nodes and does not need the volfileserver argument (otherwise it would not work). But if I remember correctly, `backup-volfile-servers=` is optional on engine side, so your change must handle the case when it is not configured on engine side.

So the logic seems to be:

    if backup_volfile_servers:
        ommit the volfileserver argument
    else:
        use the volfileserver argument

Also not using `backup-volfile-servers` seems very fragile, so we can consider warning about it when connecting to the storage.

josgutie:
@nirs the gluster command doesn't have any parameter to point to the backup servers, you can review the following client documentation [Gluster Command Line Interface](https://docs.gluster.org/en/main/CLI-Reference/cli-main/). In fact it's hard to find the definition of the --remote-host gluster client parameter.

Regarding the suggested change, the problem is if the volfile server is down, the call will fail because --remote-host limits the call to the volfile server and omit calling to the backup servers.

We could deal with connection issue in that part of the code, but from my point of view, it's a gluster cli responsibility not vdsm.

I will create and destroy a gluster storage domain.

josgutie:
New tests completed:

- GlusterFS based Storage Domain creation:
  - Create a new Storage Domain from a gluster volume.
   - New domain parameters:
    - Name: data2
    - Storage Type: GlusterFS
    - Host: rhhi01.libvirt.lab
    - Path: rhhi01.libvirt.lab:/data2
    - Mount Options: backup-volfile-servers=rhhi02.storage01.lab:rhhi03.storage01.lab
    - Result: Storage Domain data2 created.
- GlusterFS based Storage Domain maintenance:
  - Domain name: data2
  - Result: Correct
- GlusterFS based Storage Domain activation:
  - Domain name: data2
  - Result: Correct
- GlusterFS based Storage Domain detach:
 - Domain name: data2 
 - Result: Correct
- GlusterFS based Storage Domain attach:
 - Domain name: data2 
 - Result: Correct
- GlusterFS based Storage Domain remove:
 - Domain name: data2 
 - Result: Correct, gluster volume is empty.
 
The only thing left to test is to perform a full RHHI 1.8 deployment.

aesteve-rh:
@josgutie Thanks, your test battery looks correct. @nirs wdyt? do you still have concerns regarding the update?

@josgutie Could you check the module tests that are failing in the pipelines? Probably expectations changed after the update.

aesteve-rh (approved):
LGTM :)

aesteve-rh:
@josgutie linter failed, mind the longer lines. Limit is 79 characters.

I was also wondering if you could add a small test that throws an exceptions and triggers the retry with the backup servers. So that at least we add some coverage for the new code.

tests/storage/storageserver_test.py
```diff
@@ -250,6 +250,32 @@ def test_glusterfs_cli_missing(self, monkeypatch):
         gluster = GlusterFSConnection(id="id", spec="192.168.122.1:/music")
         assert gluster.options == ""
 
+    @pytest.mark.parametrize("userMountOptions", [
+        '',
+        'backup-volfile-servers=192.168.122.1:192.168.122.2',
+    ])
+    def test_glusterfs_backup_serv_connection(self, monkeypatch):
```
aesteve-rh, line 257: The test is parametrized, but then you do not really use the parameter.

aesteve-rh (approved):

michalskrivanek:
el8stream OST is not working, el9stream should pass

lib/vdsm/storage/storageServer.py
```diff
@@ -382,14 +382,24 @@ def _get_backup_servers_option(self):
...
+        except ge.GlusterException as e:
             log.warning("Failed to get volume info: %s", e)
-            return {}
+            try:
+                if "backup-volfile-servers" in self._options:
+                    log.warning(
+                        "Trying to get volume info through backup servers: %s",
+                        self._options)
+                    volinfo = superVdsmProxy.glusterVolumeInfo(self._volname)
+                    return volinfo[self._volname]
```
nirs, line 398: Do we always have "backup-volfile-servers" in options? If we don't this fix will not help in the cases we don't. If we are running on a hyperconverge system we have a local gluster server regardless of having "backup-volfile-servers" options.

Regardless, there is no reason to put the if inside the try block.

Warning before trying "backup-volfile-servers" is not needed, this can be info level log.

I would always retry, and add a comment explaining why we retry. Something like:

```python
    try:
        volinfo = superVdsmProxy.glusterVolumeInfo(self._volname, self._volfileserver)
        return volinfo[self._volname] 
    except ge.GlusterException as e:
        # The remote host may be down. If we are running on a hyperconverged system
        # the gluster client can use one of the other gluster servers.
        log.info("Failed to get volume info from remote server: %s",
                 self._volfile_server)
        try:
            volinfo = superVdsmProxy.glusterVolumeInfo(self._volname)
            return volinfo[self._volname] 
        except ge.GlusterException as e
            log.warning("Failed to get volume info using local server: %s", e)

    return {}
```

tests/storage/storageserver_test.py
```diff
@@ -250,6 +250,29 @@ def test_glusterfs_cli_missing(self, monkeypatch):
...
+                    }
+        storageServer.supervdsm.glusterVolumeInfo = glusterVolumeInfo
+        user_options = "backup-volfile-servers=192.168.122.1:192.168.122.2"
+        gluster = GlusterFSConnection(id="id", spec="192.168.122.3:/music",
+                                      options=user_options)
+        expected_volinfo = {'brickCount': '2',
+                            'bricks': ['192.168.122.1:/tmp/music',
+                                       '192.168.122.2:/tmp/music']
+                            }
+        assert gluster.volinfo == expected_volinfo
```
nirs, line 274: The test looks right for the current code, but need to change if the code will always retry.

nirs (commented):
This change will not work in all cases, is little bit messy and need better documentation. See comments on to make it more clear.

josgutie:
Hi Nir, PR is updated, my doubt in on the test coverage, but it should work in both cases, where backup server are defined and not.

lib/vdsm/storage/storageServer.py
```diff
@@ -387,9 +387,21 @@ def _get_gluster_volinfo(self):
...
-            return {}
+        except ge.GlusterException as e:
+            # The remote host may be down.
+            # If we are running on a hyperconverged system the gluster client
+            # can use one of the other gluster servers.
+            log.info("Failed to get volume info from remote server %s: %s",
+                     self._volfileserver, e)
+            try:
+                log.debug("Trying to get volume info from backup servers: %s",
+                          self._options)
```
nirs, line 398: Logging the mount options in debug level makes sense, but we don't need to log inside the try block. We try to keep try blocks small as possible to avoid incorrect error handling.

tests/storage/storageserver_test.py
```diff
@@ -250,6 +250,33 @@ def test_glusterfs_cli_missing(self, monkeypatch):
...
+    ])
+    def test_glusterfs_backup_serv_connection(self, monkeypatch,
+                                              userMountOptions):
+        monkeypatch.setattr(storageServer, 'supervdsm', FakeSupervdsm())
+        monkeypatch.setattr(gluster_cli, 'exists', lambda: True)
+
+        def glusterVolumeInfo(volname=None, volfileServer=None):
+            assert volname == "music"
+            if volfileServer is not None:
+                raise ge.GlusterException()
```
nirs, line 265: This is nice way to fail the test on the first try, but it is hidden in the middle of the test, so it is hard to understand the purpose of the test.

I think we should rename the test and add docstring like other tests explaning the purpose of the test.

```python
def test_glusterfs_retry_withotut_volfile_server(...)
    """
    The test will fail if called with volfileserver, simulating the case when
    a remote server is down. If the caller try again without volfileserver the
    call will succeed, simulating the case when running on a hyperconverged
    system when the local glsuter glsuter client can use one of the connected
    servers.
    """
```

nirs (approved):
Thanks @josgutie! Looks very clean and safe change. The test can be renamed and documented.

lib/vdsm/storage/storageServer.py
josgutie: Suggested change implemented.

tests/storage/storageserver_test.py
josgutie: Test function name changed, and purpose description added.

aesteve-rh (approved):

PR #414, opened 2024-05-02
Backport: get gluster volume info from any gluster peer

aesteve-rh:
The function _get_gluster_volinfo query the glusterfs volume info the the storage server, this is translated to the gluster client adding the parameter --remote-host which limits the query to one server, so we are converting the storage server as a single point of failure, if it is not available, it can led to cluster outtage. The proposed changed let the cluster cli to use any available gluster peer.

Bug-Url: https://bugzilla.redhat.com/2264175

Signed-off-by: José Enrique Gutiérrez Mazón <josgutie@redhat.com>

### DIFF
--- a/lib/vdsm/storage/storageServer.py
+++ b/lib/vdsm/storage/storageServer.py
@@ -387,9 +387,21 @@ class GlusterFSConnection(MountConnection):
             volinfo = superVdsmProxy.glusterVolumeInfo(self._volname,
                                                        self._volfileserver)
             return volinfo[self._volname]
-        except ge.GlusterCmdExecFailedException as e:
-            log.warning("Failed to get volume info: %s", e)
-            return {}
+        except ge.GlusterException as e:
+            # The remote host may be down.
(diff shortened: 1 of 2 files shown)


mz-pdm (changes requested):
As for functionality, I trust the reviews from master.

The typos in code comments can be addressed in master but please separate the commit title from the rest by a blank line to prevent some git clients to present the whole commit message as a heading. And break the excessively long lines in the commit message.

aesteve-rh:
> As for functionality, I trust the reviews from master.
> 
> The typos in code comments can be addressed in master but please separate the commit title from the rest by a blank line to prevent some git clients to present the whole commit message as a heading. And break the excessively long lines in the commit message.

I did not notice the issue with the commit message in the original patch... But at least we can fix it in the backport, true. Just pushed the change. Thanks for checking!

nirs (approved):

mz-pdm (approved):
Thanks.

PR #416, opened 2024-05-17
virt: remove bandwidth limit on nic update

dupondje:
There was some logic error in the update_bandwidth_xml function. If you have a running VM with a NIC that has a QoS, it will contain a bandwidth setting in the XML.
Now if you change the NIC to a vNIC Profile without QoS, it will not contain the 'inbound' and 'outbound' specParams, and the old QoS will never get removed (only after stop/start).

So we take the removal of the bandwidth out of the if so the limit gets cleared correctly.


mz-pdm (approved):
Thank you for the patch!

PR #417, opened 2024-05-23
Skip LVM filesystem check on LV reduction

BrooklynDewolf:
Since LVM v2.03.17 (https://github.com/lvmteam/lvm2/commit/f6f2737015746b1b6c7fbd0d297a4596c584749b), reducing a logical volume (LV) requires the LV to be active due to the default 'checksize' option, which requires an active LV. This results in the following error when attempting to reduce a non-active LV:

`err=[\'  The LV must be active to safely reduce (see --fs options.)\']' ----`

To resolve this, we now check if the LVM version is newer than 2.03.17. If so, we bypass the 'checksize' by using the '--fs ignore' option. This approach is viable because oVirt already handles checksize, making lvreduce redundant.

### DIFF
--- a/lib/vdsm/osinfo.py
+++ b/lib/vdsm/osinfo.py
@@ -248,6 +248,7 @@ def package_versions():
             'qemu-kvm': ('qemu-kvm', 'qemu-kvm-rhev', 'qemu-kvm-ev'),
             'spice-server': ('spice-server',),
             'vdsm': ('vdsm',),
+            'lvm2': ('lvm2', 'lvm2-libs'),
         }
 
         if glusterEnabled:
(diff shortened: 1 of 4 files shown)


dupondje:
Reviewed-by: Jean-Louis Dupond <jean-louis@dupond.be>

Quite an important one. As if the lvreduce fails, the snapshot disk becomes ILLEGAL.
This only affects el9 hosts btw, as el8 has older lvm version without this --fs option.

BrooklynDewolf:
Thank you for you feedback. I have added a parameterized test that tests the reduce function with different LVM versions which runs without any problems!

aesteve-rh:
Great, can you please check for `lvm` at `osinfo_test.py` too?

BrooklynDewolf:
I am not quite sure what you mean. I've added an extra assert to the `test_package_versions()` function.

aesteve-rh (approved):

BrooklynDewolf:
No changes, just squashed my commits together.

sandrobonazzola:
Merging based on previous review and approval

PR #431, opened 2025-06-03
aarch64: add default image

TrappM:
Add the 'generic virtual board model' for aarch64 support.

Currently there is no aarch64 board supported in vdsm/_DEFAULT_MACHINES and according to the qemu documentation
https://qemu-project.gitlab.io/qemu/system/arm/virt.html
'virt' seems to be the recommended generic board model for this platform.


TrappM:
Hi,

we are currently looking for aarch64 support in ovirt and this fix is required in the vdsm part of our setup,
because vdsm can't match the cpu architecture to an image.
There might be more changes required, like the feature/flag detection and 'model_*' labeling, but at the moment this is what we've fixed in vdsm and the virt model is sufficient for the qemu setup.

dupondje:
@TrappM : Thanks for the additional info!
Hope to see more patches coming :)

PR #442, opened 2025-10-16
storage: add dynamic managed storage dispatcher

sp-viktori:
Dispatch mechanism that third-party storage vendors can use to redirect managed storage operations.

* Add an optional `adapter` field in the `ManagedVolumeConnection` struct.

* If there is an `adapter` field, direct storage requests to `managedvolume-helper-{adapter}`, otherwise use the default (os-brick) implementation.

Works together with a simplar PR in ovirt-engine that allows redirection of the rest of the managed storage operations.

### DIFF
--- a/doc/managed-volume-adapters.md
+++ b/doc/managed-volume-adapters.md
@@ -1,0 +1,52 @@
+<!--
+SPDX-FileCopyrightText: oVirt Developers
+SPDX-License-Identifier: GPL-2.0-or-later
+-->
+
+# Managed Volume Adapters
+
+By default the managed volumes in Vdsm are manipulated through an external
+helper that interfaces with the `os_brick` package (part of OpenStack).
+
+Similarly on the oVirt Engine side there is another external helper that
+interfaces with CinderLib.
+
+The two together can expose any supported Cinder storage driver to oVirt.
+
+- on the oVirt Engine side the helper deals with creating and managing volumes,
+  snapshots and preparing them to be attached to a VM;
+
+- on the Vdsm side the helper deals only with attaching and detaching volumes
+  that have been created by the oVirt Engine.
+
+The adapter mechanism redirects the execution to vendor-provided helper
+executables. To facilitate this redirection both the oVirt Engine and Vdsm have
+the notion of `adapter`.
+
+This allows storage vendors to integrate their managed storage directly in
+oVirt/Vdsm.
+
+## oVirt Engine
+
+Managed Storage Domains with adapter dispatch have an `adapter` field in their
+`driver_options` map that indicates the helper executable to run instead of
+the default `cinderlib-client.py` helper.
+
+The vendor packaging is expected to install a symlink in the Managed Block
+Storage data directory (defaults to `/usr/share/ovirt-engine/managedblock/`)
+named `{adapter}-adapter`.
+
+## Vdsm
+
+The `adapter` field is passed to Vdsm in the `connection_info` parameter (of
+type `ManagedVolumeConnection`).
+
+If the `adapter` field is not present, Vdsm uses the default OS-Brick helper
+`managedvolume-helper`.
+
+If the `adapter` field is present, Vdsm will instead execute
+`managedvolume-helper-{adapter}` when attaching and detaching volumes.
+
+The default install path for the helper (and where Vdsm would be looking for the
+adapter) is `/usr/libexec/vdsm`. Vendors should add a symlink in that directory
+to their own installation path.
--- a/lib/vdsm/api/vdsm-api.yml
+++ b/lib/vdsm/api/vdsm-api.yml
@@ -1185,6 +1185,11 @@ types:
                 volume type.
             name: data
             type: *StringMap
+        -   defaultvalue: null
+            description: Adapter vendor
+            name: adapter
+            type: string
+            added: '4.5.8'
         type: object
 
     ManagedVolumeAttachement: &ManagedVolumeAttachement
--- a/lib/vdsm/storage/managedvolume.py
+++ b/lib/vdsm/storage/managedvolume.py
@@ -17,7 +17,6 @@ import functools
 import json
 import logging
 import os
-import sys
 
 from contextlib import closing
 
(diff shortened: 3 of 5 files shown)


lib/vdsm/api/vdsm-api.yml
```diff
@@ -1185,6 +1185,11 @@ types:
                 volume type.
             name: data
             type: *StringMap
+        -   defaultvalue: null
+            description: Adapter vendor
+            name: adapter
+            type: string
+            added: '4.5.6'
```
dupondje: 4.5.8 it would be if merged now

doc/managed-volume-adapters.md
```diff
@@ -0,0 +1,47 @@
...
+oVirt/Vdsm.
+
+## oVirt Engine
+
+Managed Storage Domains with adapter dispatch have an `adapter` field in their
+`driver_options` map that indicates the helper executable to run instead of
+the default `cinderlib-client.py` helper.
+
+The vendor packaging is expected to install a symlink in
+`/usr/share/ovirt-engine/cinderlib/` named `{adapter}-adapter`.
```
dupondje: Just in doubt if this is the right directory to put this in.
`/usr/share/ovirt-engine/managedvolume/` perhaps and move cinderlib into it?

lib/vdsm/storage/managedvolume.py
```diff
@@ -94,7 +94,7 @@ def attach_volume(sd_id, vol_id, connection_info):
                   vol_id, connection_info)
 
         try:
-            attachment = run_helper("attach", connection_info)
+            attachment = run_helper("attach", connection_info, connection_info.get("adapter"))
```
dupondje: This is useless? As you pass connection_info, you can also get the adapter in the run_helper function then.

dupondje (changes requested):

lib/vdsm/api/vdsm-api.yml
sp-viktori: thanks for catching this, updated

doc/managed-volume-adapters.md
sp-viktori: Right, I've changed this to reflect the oVirt Engine PR (the updated proposal indeed uses a `managedblock` directory instead of `cinderlib`).

lib/vdsm/storage/managedvolume.py
sp-viktori: In this case, yes, but the `run_helper` function has a union type: in this case it is `connection_info` but when calling for `detach` it is `vol_info`. Good news is that `vol_info` has `connection_info` as a field (which has `adapter`) but it becomes very convoluted to figure it out.

I'll think on simplifying the whole thing while keeping it robust, and push an update soon.

sp-viktori: I've pushed a different approach, and also split it in two:
1. A change in how managedvolume-helper is called that makes the protocol a bit stricter (so there is no ambiguity what is sent as input: it is always volume_info struct).
2. The adapter dispatch which now knows exactly where to look for the adapter field.

doc/managed-volume-adapters.md
```diff
@@ -0,0 +1,48 @@
+<!--
+SPDX-FileCopyrightText: Red Hat, Inc.
```
dupondje: You can change this to 'oVirt Developers'.

doc/managed-volume-adapters.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Vdsm
+
+The `adapter` field is passed to Vdsm in the `connection_info` parameter (of
+type `ManagedVolumeConnection`).
+
+If the `adapter` field is not present, Vdsm uses the default OS-Brick helper
+`managedvolume-helper`.
+
+If the `adapter` field is present, Vdsm will instead execute
```
dupondje, line 47: Might be useful to include path where this adapter needs to be installed?

lib/vdsm/storage/managedvolume.py
```diff
@@ -181,17 +182,27 @@ def volumes_info(vol_ids=()):
...
-            cmd_input = json.dumps(cmd_input).encode("utf-8")
+        adapter = None
+        cmd_input = None
+        if vol_info:
+            cmd_input = json.dumps(vol_info).encode("utf-8")
+            adapter = vol_info.get("connection_info", {}).get("adapter")
+        helper = HELPER
+        if adapter:
+            helper = f"{HELPER}-{adapter}"
+            if not os.path.exists(helper):
```
dupondje: What if the file exists but is non-executable?

lib/vdsm/storage/managedvolume.py
```diff
@@ -181,17 +182,27 @@ def volumes_info(vol_ids=()):
...
+                raise se.ManagedVolumeHelperFailed(
+                    f"Helper for adapter '{adapter}' not found"
+                    f" at '{helper}'")
         # This is the only sane way to run python scripts that work with both
         # python2 and python3 in the tests.
```
dupondje: Perhaps unrelated to this PR, but we dropped python2 support.
So if the execution can be improved, feel free :)

dupondje (changes requested):

lib/vdsm/storage/managedvolume.py
sp-viktori: Oof this is actually related now that you've pointed it out. My assumption (and the documentation) say that this helper can be anything but the way it's been run (for Python2 compatibility) is to hand it over to the Python executable as a script (`$python helper` basically). This means that it must have been a python script with the same version of python as Vdsm.

I'll remove the Python2 workaround and execute it directly, also checking for the executable bit. Need to run tests locally again to make sure it works (and the legacy cinderlib helper also). Will push when ready.

sp-viktori: Alright, did some more test this time and also ran OST locally with the RPMs from my Vdsm branch as well as the RPMs from my oVirt engine branch.

Removing the Python 2 workaround is OK and allows vendors to use non-Python adapters/helpers as well.

lib/vdsm/storage/managedvolume.py
sp-viktori: Added executable check as well, works with symlinks in my tests.

doc/managed-volume-adapters.md
sp-viktori: Done

doc/managed-volume-adapters.md
sp-viktori: Added more info, including default path similar to the engine-side adapter

dupondje (approved):

PR #443, opened 2025-10-16
tests: remove deprecated ssl wrap socket

JasperB-TeamBlue:
## Requested Changes
* ssl wrap socket is deprecated since 3.7 so it should be updated. Due to this function not being in use anymore, remove it instead.

### DIFF
--- a/tests/integration/sslhelper.py
+++ b/tests/integration/sslhelper.py
@@ -13,7 +13,6 @@
 from __future__ import absolute_import
 from __future__ import division
 
-import ssl
 import tempfile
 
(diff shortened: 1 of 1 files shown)


tests/integration/sslhelper.py
```diff
@@ -79,12 +78,3 @@ def key_cert_pair():
 def create_ssl_context(key_file, cert_file):
```
dupondje: I think this one is unused also, can you check and also drop it if so :)

dupondje (approved):

PR #455, opened 2026-01-15
v2v: emit -oo vdsm-...=... plugin options instead of deprecated --vds…

aikei421:
Summary
-------
This patch updates vdsm/v2v.py and related tests so that VDSM passes virt-v2v plugin
options in the modern `-oo key=value` form instead of the deprecated
`--vdsm-key value` pairs.

Motivation
----------
Newer virt-v2v versions expect plugin options in `-oo key=value` form and reject
`--vdsm-*` style arguments (causing invocation failures such as
"unrecognized option '--vdsm-vm-uuid'"). This change restores compatibility
between VDSM-generated commands and current virt-v2v releases.
https://www.mankier.com/1/virt-v2v-output-ovirt

What I changed
--------------
- vdsm/v2v.py
  - Replace occurrences where VDSM previously emitted `--vdsm-<name> <value>`
    with `-oo` / `key=value` pairs.
  - Affected places:
    - LibvirtCommand._command
    - OvaCommand._command
    - XenCommand._command
    - _disk_parameters() — disk image/volume UUIDs are emitted as:
      `-oo vdsm-image-uuid=<id> -oo vdsm-vol-uuid=<id>`

- tests/fake-virt-v2v
  - Extend the fake virt-v2v test binary to accept `-oo key=value` arguments and
    map them to the same internal variables the rest of the fake program uses.
    This keeps tests compatible with both old `--vdsm-*` and new `-oo`
    forms.

- tests (v2v_test.py)
  - Update tests to expect `-oo vdsm-...=...` pairs. Test suite passes locally
    after the change.

Reproduction / Before/After
---------------------------
Before:
- Importing an OVA with current virt-v2v produced:
  `/usr/bin/virt-v2v: unrecognized option '--vdsm-vm-uuid'`
- virt-v2v invocation failed because VDSM passed deprecated option format.

After:
- VDSM now invokes virt-v2v like:
  `virt-v2v -oo vdsm-vm-uuid=<id> -oo vdsm-ovf-output=<path> -oo vdsm-image-uuid=<id> ...`
- OVA import via virt-v2v proceeds and VM boots (verified in test environment).

Testing
-------
- `python3 -m py_compile vdsm/v2v.py` (syntax check)
- `pytest tests/v2v_test.py::v2vTests::testV2VOutput` (unit test added/updated)
- Performed a full OVA import in our environment after removing passt and using slirp;
  import completed and VM booted successfully.

Compatibility
-------------
- This change is backwards-compatible for the runtime: VDSM emits the modern
  `-oo` plugin options that virt-v2v expects. Tests and the fake test helper now
  accept both forms to preserve compatibility for older test cases.

Notes for reviewers
-------------------
- Please double-check any other `--vdsm-*` occurrences in the repo (I updated the
  obvious spots used for virt-v2v invocations; if there are other code paths that
  build `virt-v2v` commands, they should be checked).
- The tests include changes to the fake-virt-v2v helper to support `-oo`
  key=value; ensure this helper change is acceptable for CI.
- I used `-oo` exactly as virt-v2v expects (pairs of `-oo` and `key=value`).
  Order of `-oo` pairs matters in tests that assert exact command sequences;
  tests were updated accordingly.


tests/fake-virt-v2v
```diff
@@ -40,8 +40,41 @@ parser.add_argument('-x', dest='libguestfsTrace',
                     action='store_true',
                     help='Enable tracing of libguestfs API calls.')
 parser.add_argument('vmname', nargs='?')
+parser.add_argument('-oo', action='append', dest='oo', default=[],
```
dupondje, line 43: While this works, I would perhaps adjust the existing parsers?
Cause now they are there without any reason.

dupondje (changes requested):
Looks good. Just a remark on the tests.
And next to that, please also sign your commits with 'Signed-off-by' tag (git commit -s), and add some info in the commit message.

Thanks!

ogasay-git:
Thank you for accommodating this option specification change.
There is one more change I would like you to add, and line 403 should be changed from

```
self._base_command.extend(['--vdsm-compat', qcow2_compat])
```

to

```
self._base_command.extend(['-oo', 'vdsm-compat=' + qcow2_compat])
```

Thank you!

dupondje:
Superseded by #460

PR #459, opened 2026-03-02
Added option to upload disk to Managed Block domains

slavkap:
## Changes introduced with this PR

* Enables uploading disks to Managed Block Storage domains. This change adds the full upload flow for Managed Block Storage using the existing NBD-based transfer path, including connect/attach on the host and cleanup on finish. 

* It was tested with StorPool and Ceph as managed block storage domains

* The changes in this PR highly depend on [ovirt-engine PR](https://github.com/oVirt/ovirt-engine/pull/1125)

### DIFF
--- a/lib/vdsm/storage/nbd.py
+++ b/lib/vdsm/storage/nbd.py
@@ -14,6 +14,7 @@ import json
 import logging
 import os
 import time
+from contextlib import closing
 
 from vdsm.common import cmdutils
 from vdsm.common import constants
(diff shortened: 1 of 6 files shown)


lib/vdsm/storage/nbd.py
```diff
@@ -490,18 +529,35 @@ def json_uri(config):
     return "json:" + json.dumps(image)
 
 
+# Path prefixes allowed for attached Managed Block Storage volumes
+_MANAGED_VOLUME_PATH_PREFIXES = (
+    "/dev/mapper/",
```
ryan-ronnander: Hello, 

Just a heads up, I've been testing this PR with LINSTOR and DRBD to improve MBS functionality in oVirt. One minor change is needed to add support for LINSTOR to also take advantage of the disk upload:

```suggestion
    "/dev/drbd/",
    "/dev/mapper/",
```

ryan-ronnander (commented):
Greetings from LINBIT - another storage provider using oVirt's Managed Block Storage (MBS) feature.

Thanks for submitting this PR, we've tested it with LINSTOR (which uses DRBD) and everything works great after making a 1 line change to allow DRBD volumes.

We have some other MBS fixes/improvements to work through as well. Looking forward to having a much better MBS experience with oVirt/OLVM in the future.

Leaving as a comment, but consider this "approved" from another MBS storage provider.

lib/vdsm/storage/nbd.py
slavkap: Thank you @ryan-ronnander, for testing this! I've added your suggestion in the latest commit

ryan-ronnander:
I've been coming back and retesting latest changes in all the relevant MBS-related PRs, etc. 

Just re-tested with the latest changes on LINSTOR/DRBD. The reworked attach path returns a clean local attachment over os_brick's LocalConnector:

```
START  attach_volume(vol_id='3e5828ed-…', connection_info={'driver_volume_type': 'local', 'data': {'device_path': '/dev/drbd/by-res/ovirt_3e5828ed_…/0', …}})
FINISH attach_volume return={'attachment': {'type': 'local', 'path': '/dev/drbd/by-res/ovirt_3e5828ed_…/0'}, …}
```

200 MiB upload to an MBS domain completed and the bytes matched the source bit-for-bit. Looks good :+1:

lib/vdsm/storage/nbd.py
```diff
@@ -490,18 +529,36 @@ def json_uri(config):
...
 def _verify_path(path):
     """
     Anyone running as vdsm can invoke nbd_start_transient_service() with
     arbitrary path. Verify the path is in the storage repository, or path is a
-    transient disk with a backing file in the storage repository.
+    transient disk with a backing file in the storage repository, or path is an
+    attached Managed Block Storage volume.
     """
     path = os.path.normpath(path)
+    log.info("_verify_path checking path=%r", path)
```
peter-boden: How heavily is this method used? Because info logging for each call could produce quite a lot of noise in the logging.

lib/vdsm/storage/nbd.py
peter-boden: Can we try to keep things vendor agnostic? I might be wrong but, everything that went through `attach_volume` has a symlink in `/run/vdsm/managedvolumes`. 
So if there is a symlink in there that points to the volume then it is ok/valid?
That way we wont have to add prefixes for each vendor?

peter-boden (changes requested):

lib/vdsm/storage/nbd.py
slavkap: thanks @peter-boden, I've changed it to debug()

slavkap:
Thank you @peter-boden, I made the change to check for the symlink. I have tested it with StorPool and Ceph as MBS.

@ryan-ronnander, is it possible when you have time to test the changes with Linstor?

peter-boden:
Alright, lgtm

I've triggered CI, seems some things need fixing.

We'll also wait for Ryan, if he has the time to test :pray:

ryan-ronnander:
@slavkap + @peter-boden - I re-tested the latest with LINSTOR (DRBD-backed) and it all works:

- Disk upload end to end on both thin and thick pools
- `convert_volume` on a qcow2 upload (qcow2→raw), all sha256-verified
- The symlink-based `_verify_path` picks up our `/dev/drbd/` devices with no vendor prefix needed, looks good from here :+1:

PR #467, opened 2026-05-06
lvmdevices: preserve external storage VGs in the devices file

### LINKED ISSUE #466: oVirt host setup evicts pre-existing LVM-backed storage on hyperconverged deployments
**Affected version**

* VDSM version: vdsm-4.50.7-202604291142.gitd93ea129b.el9.x86_64 (master snapshot, hosts on Red Hat Enterprise Linux release 9.7 / "Plow")

Also reproduced on Oracle Linux Virtualization Manager (OLVM) 4.5.5:

* VDSM version: vdsm-4.50.5.1-8.el9.x86_64 (hosts on Oracle Linux 9.7)

**Describe the bug**

Hyperconverged deployments require external storage to be configured on the host *before* oVirt host setup runs. LINSTOR with the LVM (or LVM-thin) driver backs its pools with LVM Volume Groups (e.g. `drbdpool`); operator-deployed VGs serve the same shape. When `vdsm-tool config-lvm-filter` runs as part of host setup (or any equivalent path that drives `/etc/lvm/devices/system.devices`), vdsm rebuilds the file using only the VGs that back the host's mounted filesystems (`lvmfilter.find_lvm_mounts`, typically the OS root VG, or empty if root is on a partition). Any VG that already existed on the host for any other reason is silently evicted: it stays on disk, but LVM stops seeing it.

This happens on **OL/EL 8** (where `use_devicesfile` defaults to `0` and vdsm flips it to `1` while writing the file) and on **OL/EL 9** (where `use_devicesfile=1` is already the default but vdsm still rewrites the file). The triggering mechanism differs by host OS but the operator-visible result is the same: the external VG disappears.

The result is that the hyperconverged storage layer loses access to its own pool the moment vdsm finishes host setup. Operators see `vgs` showing the VGs before host setup and not showing them after, with no warning issued by either oVirt or vdsm.

**Especially problematic for self-hosted engine.** Self-hosted engine deployments require the storage that will host the engine VM disk to exist *before* the installer runs. The recommended hyperconverged pattern (LINSTOR satellite up first, then deploy oVirt on top) currently breaks because the first oVirt host setup pass evicts the LINSTOR-managed VG that was supposed to back the engine disk. Operators have to manually re-run `vgimportdevices` or hand-edit `system.devices` to recover.

**To reproduce**

On a fresh AlmaLinux 9 / Oracle Linux 9 / RHEL 9 host:

1. Configure an external LVM-backed storage layer first. We'll use `drbdpool` as a LINSTOR VG example: `vgcreate drbdpool /dev/vdb` on a spare disk (`vgcreate` creates the PV automatically).
2. Verify `vgs` shows `drbdpool`.
3. Run vdsm host setup (e.g. `vdsm-tool config-lvm-filter`, or attach the host through Engine).
4. After host setup completes, run `vgs` again.

**Expected behavior**

The pre-existing `drbdpool` VG (and any other external LVM-backed VG already on the host) is preserved across vdsm host setup. `vgs` continues to show it. The hyperconverged storage layer continues to operate normally.

**Actual behavior**

The pre-existing external VG is missing from `vgs` output after host setup. `system.devices` lists only the VGs that back the host's mounted filesystems (typically `vg_root` for the OS root if root is on LVM), plus any oVirt-managed SD VGs vdsm attached during this run. The external VG remains on disk but is no longer visible to LVM, breaking the hyperconverged storage layer. No warning is issued by oVirt or vdsm that LVM's view just changed.

**Additional context**

Affects any LVM-backed VG that exists on the host before vdsm runs. LINSTOR's volume group requirement is the case we hit and verified, but the bug is in vdsm's devices-file logic and applies to any pre-existing VG. Most acutely felt on self-hosted-engine deployments, where the storage layer must exist before the engine disk gets created.

We've been using some exotic deployment hooks and strict ordering to get around these issues, but it would be much better to fix this at the source in vdsm. Standalone engine oVirt environments can be installed before LINSTOR and volume group exists, requiring no workarounds.

Ideally the LVM device file is not wiped upon KVM host installation, leaving more flexibility to install LINSTOR in either order (LINSTOR required first for hyperconverged self-hosted engine deployments).

ryan-ronnander:
Two commits that together stop `vdsm-tool config-lvm-filter` (and equivalent paths) from evicting pre-existing external-storage VGs from `/etc/lvm/devices/system.devices`.

1. `lvmdevices: preserve external-storage VGs across config-lvm-filter`. List every VG `lvm` currently sees on the host before writing the new devices file, then continue with the additive `vgimportdevices` loop for the mount-backing VGs received in the `vgs` argument. Existing populated devices files are left untouched. Preserves the PR #324 invariant ("file exists when `use_devicesfile=1`"): a host with no VGs anywhere still gets a minimal header file dropped.

2. `lvmdevices: skip oVirt-tagged VGs when listing visible VGs`. Filters VGs carrying the `RHAT_storage_domain` tag out of commit 1's listing, so a stale oVirt SD VG from a prior install isn't pulled back into `system.devices` (which would let sanlock try to acquire leases on a VG the engine no longer references). Mirrors the skip rule in `lvmfilter.find_lvm_mounts`.

Fixes #466

## Changes introduced with this PR

* `lib/vdsm/storage/lvmdevices.py`: `_create_system_devices` lists currently-visible VGs before writing on the missing-or-empty-header path; `_list_all_visible_vgs` queries `vgs -o vg_name,vg_tags --separator |` and skips entries tagged with `lvmfilter.OVIRT_VG_TAG` (`RHAT_storage_domain`).
* `tests/storage/lvmdevices_test.py` (new): five test scenarios.

## Verification

Reproduced and verified on Oracle Linux 9.7:

1. Fresh host, no oVirt yet.
2. `vgcreate drbdpool /dev/vdb`. `vgs` shows `drbdpool`.
3. Run `vdsm-tool config-lvm-filter`.
4. Run `vgs`.

Before this PR: step 4 shows `drbdpool` missing. After this PR: step 4 shows `drbdpool` preserved with physical devices populated in `/etc/lvm/devices/system.devices`:

```
[root@ovirt-2 ~]# cat /etc/lvm/devices/system.devices
# LVM uses devices listed in this file.
# Created by LVM command vgimportdevices pid 23236 at Sat May  2 22:18:30 2026
# HASH=1873244822
PRODUCT_UUID=79eecfca-941b-4391-bb3f-8214ce28bd44
VERSION=1.1.2
IDTYPE=devname IDNAME=/dev/vda3 DEVNAME=/dev/vda3 PVID=4FEsaqkurHXOxzBbcK1drtYtvG9LAt8K PART=3
IDTYPE=devname IDNAME=/dev/vdb DEVNAME=/dev/vdb PVID=0DuKehmrfiPCclNGJVnmCQlDHJMMCjLf
IDTYPE=devname IDNAME=/dev/vdc DEVNAME=/dev/vdc PVID=sX1SQHFomvy8JrxCmjfcfIfdAI5lv9aj
IDTYPE=devname IDNAME=/dev/vdd DEVNAME=/dev/vdd PVID=W6xB7SAOGPf5zIklbTkmManUJC9HYkpK
```

## Are you the owner of the code you are sending in, or do you have permission of the owner?

Yes.

### DIFF
--- a/lib/vdsm/storage/lvmdevices.py
+++ b/lib/vdsm/storage/lvmdevices.py
@@ -22,6 +22,7 @@ import subprocess
 from vdsm import constants
 from vdsm.common import cmdutils
 from vdsm.common import commands
+from vdsm.storage import constants as sc
 from vdsm.storage import fileUtils
 from vdsm.storage import lvmconf
 from vdsm.storage import lvmfilter
@@ -104,22 +105,93 @@ def _configure_devices_file(enable=True):
 
 def _create_system_devices(vgs):
     """
-    Import devices of provided VGs into LVM devices file.
+    Import the VGs in `vgs` (the VGs that back the host's mounts, from
+    lvmfilter.find_lvm_mounts) plus any other VGs already visible to
+    lvm, so external storage VGs are not evicted when vdsm rebuilds
+    the devices file.
     """
-    if not vgs:
-        # Create empty devices file if no VGs are provided.
-        devices_file = _get_devices_file_path()
-        now = datetime.datetime.now()
-        data = (
-            f"# Created by Vdsm pid {os.getpid()} at "
-            f"{now.strftime('%a %b %d %H:%M:%S %Y')}\n"
-        )
+    devices_file = _get_devices_file_path()
 
-        os.makedirs(os.path.dirname(devices_file), exist_ok=True)
-        fileUtils.atomic_write(devices_file, data.encode("utf-8"))
-    else:
-        for vg in vgs:
-            _run_vgimportdevices(vg)
+    existing_vgs = ()
+    if not _devices_file_has_entries(devices_file):
+        # No device entries yet. List visible VGs so external storage
+        # is preserved on the use_devicesfile=1 transition.
+        existing_vgs = _list_all_visible_vgs()
+        if not existing_vgs and not vgs:
+            # No VGs anywhere. Drop a header so lvm does not scan every
+            # device under use_devicesfile=1.
+            now = datetime.datetime.now()
+            data = (
+                f"# Created by Vdsm pid {os.getpid()} at "
+                f"{now.strftime('%a %b %d %H:%M:%S %Y')}\n"
+            )
+            os.makedirs(os.path.dirname(devices_file), exist_ok=True)
+            fileUtils.atomic_write(devices_file, data.encode("utf-8"))
+            return
+
+    # Import visible and requested VGs, filtering out duplicate VGs.
+    for vg in sorted(set(existing_vgs) | set(vgs)):
+        _run_vgimportdevices(vg)
+
+
+def _devices_file_has_entries(devices_file):
+    """
+    Return True if the devices file records at least one device entry
+    (an IDTYPE= line). A missing or empty file, or one holding only
+    comments and bare lvm metadata (VERSION=, PRODUCT_UUID=) with no
+    device entries, counts as no entries.
+    """
+    try:
+        with open(devices_file) as f:
+            for line in f:
+                if line.strip().startswith("IDTYPE="):
+                    return True
+    except FileNotFoundError:
+        return False
+    return False
+
+
+def _list_all_visible_vgs():
+    """
+    Return the sorted VG names lvm can see, bypassing the devices file
+    so VGs created while use_devicesfile was off are still surfaced.
+    Skips oVirt SD VGs (sc.STORAGE_DOMAIN_TAG), like
+    lvmfilter.find_lvm_mounts.
+    """
+    cmd = [
+        constants.EXT_LVM,
+        'vgs',
+        '--noheadings',
+        '--readonly',
+        '-o',
+        'vg_name,vg_tags',
+        '--separator',
+        '|',
+        '--config',
+        'devices { use_devicesfile = 0 filter = ["a|.*|"] }',
+    ]
+
+    p = commands.start(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    out, err = commands.communicate(p)
+
+    if p.returncode != 0:
+        raise cmdutils.Error(cmd, p.returncode, out, err)
+
+    result = set()
+    for line in out.decode("utf-8").splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        # vgs prints "vg_name|tag1,tag2,..." with our --separator.
+        name, _, tags = line.partition("|")
+        name = name.strip()
+        if not name:
+            continue
+        if sc.STORAGE_DOMAIN_TAG in tags.split(","):
+            log.debug("Skipping oVirt-tagged VG %r", name)
+            continue
+        result.add(name)
+    return sorted(result)
 
 
 def _run_vgimportdevices(vg):
--- a/tests/storage/lvmdevices_test.py
+++ b/tests/storage/lvmdevices_test.py
@@ -1,0 +1,161 @@
+# SPDX-FileCopyrightText: oVirt Developers
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+import os
+from unittest import mock
+
+import pytest
+
+from vdsm.storage import lvmdevices
+
+
+@pytest.fixture
+def fake_devices_file(tmpdir, monkeypatch):
+    """Point _get_devices_file_path() at a tmpdir file."""
+    devices_dir = os.path.join(str(tmpdir), "lvm", "devices")
+    devices_file = os.path.join(devices_dir, "system.devices")
+    monkeypatch.setattr(
+        lvmdevices, "_get_devices_file_path", lambda: devices_file
+    )
+    return devices_file
+
+
+@pytest.fixture
+def stub_imports(monkeypatch):
+    """Capture VGs passed to _run_vgimportdevices."""
+    imported = []
+    monkeypatch.setattr(
+        lvmdevices, "_run_vgimportdevices", lambda vg: imported.append(vg)
+    )
+    return imported
+
+
+def _stub_visible_vgs(monkeypatch, vgs):
+    monkeypatch.setattr(lvmdevices, "_list_all_visible_vgs", lambda: list(vgs))
+
+
+def test_create_no_vgs_anywhere_writes_empty_header(
+    fake_devices_file, stub_imports, monkeypatch
+):
+    # No VGs anywhere: a header file must still exist so lvm does not
+    # scan every device under use_devicesfile=1.
+    _stub_visible_vgs(monkeypatch, [])
+
+    lvmdevices._create_system_devices(set())
+
+    assert stub_imports == []
+    assert os.path.exists(fake_devices_file)
+    with open(fake_devices_file) as f:
+        header = f.read()
+    assert header.startswith("# Created by Vdsm")
+
+
+def test_create_imports_externally_visible_vg_when_file_missing(
+    fake_devices_file, stub_imports, monkeypatch
+):
+    # External VG on disk but not in the missing file must be imported,
+    # not dropped.
+    _stub_visible_vgs(monkeypatch, ["external_vg"])
+
+    lvmdevices._create_system_devices(set())
+
+    assert stub_imports == ["external_vg"]
+
+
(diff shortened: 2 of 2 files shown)


tests/storage/lvmdevices_test.py
```diff
@@ -0,0 +1,125 @@
+# SPDX-FileCopyrightText: Red Hat, Inc.
```
peter-boden: We don't need this copyright anymore 🙈

lib/vdsm/storage/lvmdevices.py
```diff
@@ -105,21 +105,84 @@ def _configure_devices_file(enable=True):
...
+                    f"{now.strftime('%a %b %d %H:%M:%S %Y')}\n")
+            os.makedirs(os.path.dirname(devices_file), exist_ok=True)
+            fileUtils.atomic_write(devices_file, data.encode("utf-8"))
+            return
+
+        for vg in existing_vgs:
             _run_vgimportdevices(vg)
 
+    # vgimportdevices is additive; existing entries are not touched.
+    for vg in vgs:
```
peter-boden: There's probably no harm in it but if a VG is in both existing_vgs and in vgs, you'll run `_run_vgimportdevices` twice for the same VG.
Maybe merge the collections and deduplicate? `all_vgs = set(existing_vgs) | set(vgs)`

peter-boden (changes requested):
In the commit messages, try to refer to a commit SHA instead of referring to a PR. oVirt already moved from gerrit to github once, that gerrit metadata is now useless while commit SHA stay valid forever.

Otherwise, I don't have a lot of remarks, it's all explained nicely, good job!

tests/storage/lvmdevices_test.py
dupondje: We might want to use 'oVirt Developers'

ryan-ronnander:
> In the commit messages, try to refer to a commit SHA instead of referring to a PR. oVirt already moved from gerrit to github once, that gerrit metadata is now useless while commit SHA stay valid forever.
> 
> Otherwise, I don't have a lot of remarks, it's all explained nicely, good job!

Thanks, makes sense on SHA-over-PR refs. I dropped the `PR #324` reference entirely with the fixup commit with some alternate comment framing. I'll clean it out of the commit message as well when I squash the fixup into the original before merge.

lib/vdsm/storage/lvmdevices.py
ryan-ronnander: Good catch! I collapsed the two import loops into one over `sorted(set(existing_vgs) | set(vgs))`, so a VG that shows up in both lists is imported once. Added a test for the in-both case too.

tests/storage/lvmdevices_test.py
ryan-ronnander: Ah yeah, I was just following conventions found in nearby files. Switched this to "oVirt Developers" as suggested. If you'd rather it be a different format, feel free to tweak it or make another suggestion.

peter-boden:
@ryan-ronnander still an issue with one of the tests it seems

dupondje:
> @ryan-ronnander still an issue with one of the tests it seems

@peter-boden : unrelated, re-ran, and works now :)

ryan-ronnander:
> @ryan-ronnander still an issue with one of the tests it seems

@peter-boden - Diving into the test failure,  `TestExecCmdAffinity.testResetAffinityWhenConfigured`, a CPU-affinity test unrelated to the LVM changes here (centos-9 runner only, passes on alma-9,10, etc).  Just some sort of runner CPU topology or cgroup-cpuset difference in the CI run?

edit: Never mind -  just noticed you re-replied.

ryan-ronnander:
@peter-boden @dupondje - Just rebased onto current master, conflict resolved and re-tested on LINSTOR, should be ready to merge now :+1:

tests/storage/lvmdevices_test.py
```diff
@@ -0,0 +1,172 @@
+# SPDX-FileCopyrightText: oVirt Developers
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+from __future__ import absolute_import
+from __future__ import division
```
peter-boden: These 2 lines can be removed as since no longer target python 2

peter-boden (changes requested):
One last tiny remark

tests/storage/lvmdevices_test.py
ryan-ronnander: Good catch, just pushed an amended commit.

lib/vdsm/storage/lvmdevices.py
```diff
@@ -104,22 +105,102 @@ def _configure_devices_file(enable=True):
...
-            f"# Created by Vdsm pid {os.getpid()} at "
-            f"{now.strftime('%a %b %d %H:%M:%S %Y')}\n"
-        )
-
-        os.makedirs(os.path.dirname(devices_file), exist_ok=True)
-        fileUtils.atomic_write(devices_file, data.encode("utf-8"))
-    else:
-        for vg in vgs:
-            _run_vgimportdevices(vg)
+    cmd = [
```
dupondje, line 161: In lib/vdsm/storage/lvmfilter.py, there are similar commands executed.
Here we add --readonly to the vgs command. Do we need the same?

lib/vdsm/storage/lvmdevices.py
```diff
@@ -104,22 +105,102 @@ def _configure_devices_file(enable=True):
...
+        '--noheadings',
+        '-o',
+        'vg_name,vg_tags',
+        '--separator',
+        '|',
+        '--config',
+        'devices { use_devicesfile = 0 filter = ["a|.*|"] }',
+    ]
+
+    p = commands.start(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
dupondje: in the lvmfilter.py we run it with 'out = subprocess.check_output(args)'.
One better than the other?

lib/vdsm/storage/lvmdevices.py
```diff
@@ -104,22 +105,102 @@ def _configure_devices_file(enable=True):
...
+    ownership. Strictly safer than today's empty-overwrite behaviour.
+    """
+    devices_file = _get_devices_file_path()
+    file_was_populated = (
+        os.path.exists(devices_file) and os.path.getsize(devices_file) > 0
+    )
+
+    existing_vgs = ()
+    if not file_was_populated:
+        # File doesn't exist or holds only an empty header. List every
```
dupondje: If the file holds an empty header, the `os.path.getsize(devices_file) > 0` is true, and `file_was_populated` == true.
Which means it never gets into this if then :) Either fix the comment or adapt `file_was_populated` to ignore the header.

dupondje (changes requested):
Some minor comments/idea's.
Can you also rebase with my changes commited yesterday. These fixes the tests on vdsm, so we can then also run the tests here and see everything is green :)

lib/vdsm/storage/lvmdevices.py
ryan-ronnander: Makes sense, added `--readonly` to match `lvmfilter.py`.

lib/vdsm/storage/lvmdevices.py
ryan-ronnander: In the end I kept it consistent with this module: `_run_vgimportdevices` and `_run_check` both use `commands.start`/`communicate`, so `_list_all_visible_vgs` does the same rather than copy lvmfilter.py's older raw `subprocess.check_output`.

lib/vdsm/storage/lvmdevices.py
ryan-ronnander: Good catch, I fixed the detection rather than the comment: `_devices_file_has_entries()` now keys on actual `IDTYPE=` device entries, so a header-only file (or one holding only lvm metadata like `VERSION=` with no devices) is correctly treated as unpopulated.

ryan-ronnander:
Made one last little switch to use `mock` in  `lvmdevices_test.py` with some minor changes, no more incoming changes from me, don't know if we need to re-`/ost`?

peter-boden (approved):

dupondje (approved):

PR #470, opened 2026-05-12
virt: run cleanup after failed vm start

dupondje:
When a VM fails to start, or when an incoming migration fails, we should run a cleanup on the hypervisor.

This avoids having unused LV's active etc on the hypervisor when the VM didn't start/failed to migrate to it.


peter-boden (approved):
LGTM

JasperB-TeamBlue (approved):

PR #476, opened 2026-06-05
v2v: replace deprecated --password-file with --ip

tdemaret:
The --password-file option is no longer valid in virt-v2v. Replace it with --ip in all virt-v2v and kvm2ovirt calls, as well as in test fixtures and the fake virt-v2v script.


dupondje (changes requested):
Seems a valid change. But please:
- Make sure you have 1 proper commit in the PR, not 3 ...
- Please mention the commit (eb508ba) where this was added, and eventually the commit+version of virt-v2v where it was removed.

dupondje:
You can just force push ... :man_shrugging:

PR #478, opened 2026-06-09
lint: Expand linting to other subdirectories of vdsm

JasperB-TeamBlue:
Previously only a small part of the vdsm library was checked by the black linting.
Expand the tox command to cover all python files present.
This ensures that during CI runs the linting now ensures that all files comply with it.


tox.ini
dupondje: We might want to bump the version also?

tox.ini
```diff
@@ -197,7 +197,7 @@ commands =
         --check \
         --diff \
         {posargs} \
-        ./lib/vdsm/network/ \
+        ./lib/vdsm/ \
```
dupondje: Why not whole /lib? Cause now we still miss some files :)

dupondje (changes requested):
Some 2 minor comments already.
And next to that, I just merged something small causing merge conflict.

And finally. Is it an idea to split the changes in 3 (or so) commits:
1. Enable black on all
2. commit containing all changes black did automatically (by just running black with the correct args on the folder).
3. commit containing rest of the changes.

Cause now, so many changes, no visibility between auto changes and manual changes. Which makes it hard to review and spot the manual changes.
If the black-auto-changes are in 1 commit, I think we can just assume it's ok.

tox.ini
JasperB-TeamBlue: Version bumped to latest version available on all 4 containers.

tox.ini
JasperB-TeamBlue: whole lib is now added as well as the test directory.

tests/ssl_test.py
```diff
@@ -90,10 +90,12 @@ def dummy_register_protocol_detector(monkeypatch):
     )
 
 
-@pytest.fixture  # noqa: F811 # TODO: remove after upgrading flake to 3.9.2
+@pytest.fixture
 def listener(
-    dummy_register_protocol_detector, key_cert_pair, request
-):  # noqa: F811, E501
+    dummy_register_protocol_detector,
+    key_cert_pair,  # noqa: F811 TODO: remove after upgrading flake to 3.9.2
```
dupondje: We are already running flake8 > 6.0.0? :)

tox.ini
```diff
@@ -255,7 +255,7 @@ timeout_method = thread
 # The following test must be skipped as this rule goes against the
 # PEP 8 recommended style, which was changed on April 16th, 2016:
 # - W503 line break before binary operator
-ignore = E731, E722, W504, W503
+ignore = E731, E722, W504, W503, E203
```
dupondje, line 263: Perhaps add a comment why E203 was added. What about the others? :)

dupondje (changes requested):

tests/ssl_test.py
JasperB-TeamBlue: This error was masked as a pyflake8 versioning issue but was a fundamental issue in how this test was constructed. Instead of importing the key_cert_pair directly, which caused these pyflake errors, use the generate_key_cert_pair like it is used in the other tests.

tox.ini
JasperB-TeamBlue: Added more information about E203, other ignores can't be removed because E731 and E722 are still present while W504 and W503 are known to clash with each other and with black.

lib/vdsm/tool/check_volume_leases.py
```diff
@@ -187,9 +187,9 @@ def _repair(broken_leases):
                     repaired += 1
                 except sanlock.SanlockException as e:
                     print(
-                        "Failed to repair lease of volume {}/{}. Error {}".format(
-                            vol_lease['image'], vol_lease['volume'], e
-                        )
+                        "Failed to repair lease of volume"
```
dupondje: Missing trailing space.

lib/vdsm/tool/check_volume_leases.py
```diff
@@ -187,9 +187,9 @@ def _repair(broken_leases):
...
+                        f"{vol_lease['image']}/{vol_lease['volume']}."
```
dupondje: Same here

dupondje (changes requested):

dupondje (approved):

PR #479, opened 2026-06-09
ci fixes + copyright update

dupondje:
We want nmstate to wait that everything is applied, even on rollback. Therefor we set verify_change=True on rollback.

Otherwise changes are pushed, but might not be applied yet when the call ends.

### DIFF
--- a/contrib/add-spdx-header.sh
+++ b/contrib/add-spdx-header.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# SPDX-FileCopyrightText: Red Hat, Inc.
+# SPDX-FileCopyrightText: oVirt Developers
 # SPDX-License-Identifier: GPL-2.0-or-later
 #
 # Simple script to add SPDX copyright notice header lines to new files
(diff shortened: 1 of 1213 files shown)


contrib/add-spdx-header.sh
JasperB-TeamBlue: shouldn't this also be updated to oVirt developers?

doc/development.md
```diff
@@ -171,7 +171,7 @@ This will create default `GPL-2.0-or-later` license header
 with `Red Hat, Inc.` as copyright holder.
```
JasperB-TeamBlue: this is contradictory to what is presented below.

doc/development.md
JasperB-TeamBlue: Should also be updated.

JasperB-TeamBlue (changes requested):
Some files still have red hat licensing:
- `vdsm_log/vdsmlog.style` seems to have a copyright still on red hat, seems the search functionality missed that one due to the // in front of the comment instead of the # . 
- Same goes for `tests/virt/data/disk_devices.xml.license` and other files in that folder ending in the `.license` extension having a reference to Red Hat instead of the oVirt Developers.

JasperB-TeamBlue (changes requested):
I found one last file with a header with the old copyright, `pylinterc` has it, seems like i missed it myself. In the main README there is also the line:
```
In addition, as a special exception, Red Hat, Inc. and its affiliates
```
I don't know if this needs refactoring.

dupondje:
> 
> I don't know if this needs refactoring.

I would keep that one for now ...

JasperB-TeamBlue (approved):

PR #481, opened 2026-06-19
Feature/nvme overtcp

tdemaret:
Add NVMe-oF/TCP storage domain support to oVirt. This is a block/shared storage domain type, analogous to iSCSI, using the built-in Linux NVMe 
initiator (nvme-cli). It supports both authenticated (DH-HMAC-CHAP) and non-authenticated NVMe-oF targets.

The feature try to be backend-agnostic for any NVMe/TCP target works (Ceph NVMe-oF Gateway is the reference implementation).

# Change :
- New storage domain type: nvmeof in REST API / Ansible / SDK
- Authentication : optional DH-HMAC-CHAP 
- Multipath : native NVMe multipath detected automatically; falls back to dm-mpath
- Host deployment: NVMe host NQN auto-detected from /etc/nvme/hostnqn

# Limitations :
- No iSCSI bond support for NVMe-oF
- NVMe-oF storage domain cannot be edited or imported via webadmin


lib/vdsm/storage/nvme.py
```diff
@@ -0,0 +1,410 @@
+# SPDX-FileCopyrightText: Red Hat, Inc.
```
dupondje: we switched to 'oVirt Developers'

lib/vdsm/storage/multipath.py
```diff
@@ -589,3 +589,10 @@ def devIsiSCSI(type):
 
 def devIsFCP(type):
     return type in [DEV_FCP, DEV_MIXED]
+
+
+DEV_NVMEOF = "NVMe-oF"
```
dupondje: Put at the top

lib/vdsm/storage/storageServer.py
```diff
@@ -59,7 +63,7 @@
     sd.FCP_DOMAIN: 'fcp',
     sd.POSIXFS_DOMAIN: 'posixfs',
     sd.GLUSTERFS_DOMAIN: 'glusterfs',
-}
+    sd.NVMEOF_DOMAIN: 'nvmeof'}
```
dupondje: keep the } on next line

lib/vdsm/host/caps.py
```diff
@@ -39,7 +39,7 @@
     haClient = None
 
 
-def _parseKeyVal(lines, delim='='):
+def _parseKeyVal(lines, delim="="):
```
dupondje: Single quote to double quote changes should be in another commit if you want to change that.
Otherwise no overview of 'real' changes.

lib/vdsm/storage/multipath.py
```diff
@@ -545,6 +545,71 @@ def pathListIter(filterGuids=()):
...
+                devInfo["paths"].append(pathInfo)
+
+                if session_key:
+                    conn_info = {
+                        "connection": ctrl_info["traddr"],
+                        "port": ctrl_info["trsvcid"],
+                        "nqn": session_key,
+                        "transport": ctrl_info["transport"],
+                    }
+                    if session_key not in knownSessions or                             not knownSessions[session_key]:
```
dupondje: weird spacing :)

dupondje (changes requested):
Just a real quick review.
Will do more in depth next week :)

lib/vdsm/storage/multipath.py
```diff
@@ -514,22 +517,102 @@ def pathListIter(filterGuids=()):
...
+            session_key = ns_info.get("nqn")
+            if session_key and session_key not in knownSessions:
+                knownSessions[session_key] = {}
+
+            for ctrl_info in ns_info.get("controllers", []):
+                pathInfo = {
+                    "physdev": ctrl_info["ctrl"],
+                    "state": "active",
+                    "capacity": str(getDeviceSize(ns_name)),
+                    "lun": 0,
```
peter-boden, line 596: `state` and `lun` are hardcoded. 

Especially for `state` this can give a false-positive. I think we can easily read the state from `/sys/class/nvme/[nvmeX]/state` or something? We will have to do a little translation as the state for NVMe isn't binary (https://github.com/torvalds/linux/blob/master/drivers/nvme/host/sysfs.c#L520-L526).

Optional for `lun`, NVMe doesn't have a LUN concept, so setting it to 0 is a bit weird? Can't we just omit it? I don't think the value is consumed at the engine side?

In a lter phase I would even send the "raw" state to the engine in some way, which might come in handy for events/audit/UI/...

lib/vdsm/storage/nvme.py
```diff
@@ -0,0 +1,514 @@
...
+    if host_nqn:
+        cmd.extend(["-w", host_nqn])
+    if dhchap_key:
+        cmd.extend(["-k", dhchap_key])
+
+    log.info("Connecting to NVMe-oF target %s at %s:%s (transport=%s)",
+             nqn, traddr, trsvcid, transport)
+    try:
+        commands.run(cmd)
+    except cmdutils.Error as e:
```
peter-boden, line 78: `cmdutils` will unwillingly expose the `dhchap_key` in the log file.
The error contains the whole command: https://github.com/oVirt/vdsm/blob/master/lib/vdsm/common/cmdutils.py#L131

You can use `ProtectedPassword` to automatically obfuscate the secret (e.g. auto called when running a command https://github.com/oVirt/vdsm/blob/699fc80ca920eb85b7e33e2af7605e012136f7ec/lib/vdsm/common/commands.py#L159)

lib/vdsm/storage/storageServer.py
```diff
@@ -866,6 +875,103 @@ def __hash__(self):
...
+
+    @classmethod
+    def translate_error(cls, e):
+        if isinstance(e, nvme.NvmeAuthenticationError):
+            return (502, "NVMe-oF authentication error")
+        if isinstance(e, nvme.NvmeConnectionError):
+            return (501, "NVMe-oF connection error")
+        if isinstance(e, nvme.NvmeDisconnectionError):
+            return (503, "NVMe-oF disconnection error")
+        return super().translate_error(e)
```
peter-boden, line 944: These codes are already used elsewhere: https://github.com/oVirt/vdsm/blob/699fc80ca920eb85b7e33e2af7605e012136f7ec/lib/vdsm/storage/exception.py#L1377-L1389

You should probably use proper error codes like e.g. iSCSI does: https://github.com/oVirt/vdsm/blob/699fc80ca920eb85b7e33e2af7605e012136f7ec/lib/vdsm/storage/storageServer.py#L682-L689

**+ you will need to add new error codes to the engine as well**

lib/vdsm/storage/storageServer.py
```diff
@@ -33,10 +34,17 @@
...
 
+NvmeofConnectionParameters = namedtuple("NvmeofConnectionParameters",
+                                        "id, nqn, transport, traddr, "
+                                        "trsvcid, host_nqn, dhchap_key")
+
 GlusterFsConnectionParameters = namedtuple(
     "GlusterFsConnectionParameters", "id, spec, vfsType, options"
 )
 
+PosixFsConnectionParameters = namedtuple("PosixFsConnectionParameters",
```
peter-boden, line 45: duplicate of line 33?

lib/vdsm/storage/nvme.py
```diff
@@ -0,0 +1,514 @@
...
+           "-s", trsvcid]
+    if host_nqn:
+        cmd.extend(["-w", host_nqn])
+    if dhchap_key:
+        cmd.extend(["-k", dhchap_key])
+
+    log.info("Connecting to NVMe-oF target %s at %s:%s (transport=%s)",
+             nqn, traddr, trsvcid, transport)
+    try:
+        commands.run(cmd)
```
peter-boden, line 77: Connect does not pass `--ctrl-loss-tmo`, so every NVMe-oF controller uses the kernel default of 600 seconds ([`NVMF_DEF_CTRL_LOSS_TMO`](https://github.com/torvalds/linux/blob/master/drivers/nvme/host/fabrics.h#L17)).

That does not match how vdsm handles the same problem for device-mapper multipath. [mpathconf.py:124-135](https://github.com/oVirt/vdsm/blob/40678841d5f0a182f0a0480dff92db7f05e0506c/lib/vdsm/storage/mpathconf.py#L124-L135) says the storage path-loss grace period should match sanlock's lease window. The formula is `8 × sanlock:io_timeout / polling_interval`, which is `8 × 10 / 5 = 80s`. DM uses [`no_path_retry=16`](https://github.com/oVirt/vdsm/blob/40678841d5f0a182f0a0480dff92db7f05e0506c/lib/vdsm/storage/mpathconf.py#L79) with [`polling_interval=5`](https://github.com/oVirt/vdsm/blob/40678841d5f0a182f0a0480dff92db7f05e0506c/lib/vdsm/storage/mpathconf.py#L104), which also gives 80s.

The comment explains why: holding I/O longer than the lease window does not help, because sanlock fences the host at 80s anyway. It only leaves the host stuck. vdsm also rejects unlimited retry ([mpathconf.py:111-113](https://github.com/oVirt/vdsm/blob/40678841d5f0a182f0a0480dff92db7f05e0506c/lib/vdsm/storage/mpathconf.py#L111-L113)). On `no_path_retry=queue` it says: "will cause commands run by vdsm (e.g. lvm) to get stuck for unlimited time, causing timeouts in various flows, and may cause a host to become non-responsive." Setting `ctrl_loss_tmo=-1` is the NVMe equivalent of that mode.

I would pass an explicit `--ctrl-loss-tmo` at connect, set to about 80s to match the sanlock window. Not 600s, not -1.

btw: The vendor udev rules (shipped by nvme-cli, e.g. `71-nvmf-netapp.rules`) already set -1 for ANA-capable arrays, where redundant paths make infinite retry safe ([multipath.c:324-343](https://github.com/torvalds/linux/blob/master/drivers/nvme/host/multipath.c#L324-L343) routes I/O only to `NVME_ANA_OPTIMIZED` paths). Those rules override the connect-time value anyway, because `ctrl_loss_tmo` is a writable sysfs attribute ([sysfs.c:607-608](https://github.com/torvalds/linux/blob/master/drivers/nvme/host/sysfs.c#L607-L608)) that udev writes after the device is registered. So the connect-time value only matters for arrays with no vendor rule.

lib/vdsm/storage/nvme.py
```diff
@@ -0,0 +1,514 @@
...
+    """
+    cmd = [_NVME.cmd, "connect",
+           "-n", nqn,
+           "-t", transport,
+           "-a", traddr,
+           "-s", trsvcid]
+    if host_nqn:
+        cmd.extend(["-w", host_nqn])
+    if dhchap_key:
+        cmd.extend(["-k", dhchap_key])
```
peter-boden, line 72: `-w` and `-k` appear to be the wrong flags here. From `nvme connect --help`:
```
  [  --host-traddr=<STR>, -w <STR> ]    --- host traddr (e.g. FC WWN's)
  [  --keep-alive-tmo=<NUM>, -k <NUM> ] --- keep alive timeout period in
                                            seconds
```
This should be:
```
  [  --hostnqn=<STR>, -q <STR> ]        --- user-defined hostnqn
  [  --dhchap-secret=<STR>, -S <STR> ]  --- user-defined dhchap key (if
                                            default not used)
```
Also, wouldn't it be better to use long flag names instead of single letter ones, troughout the whole nvme code? That would prevent errors like this and make the code more readable.

PR #485, opened 2026-07-20
virt: handle vm stuck in powering down state

dupondje:
When we poweroff a vm, and it's doing some actions (for example a blockjob), this can return VIR_ERR_OPERATION_INVALID.

But if in the meantime the VM stopped anyway, then this will not be handled anymore. As from now on destroy on the VM will return VIR_ERR_OPERATION_INVALID (because the vm doesn't exist anymore), which is not handled correctly.

Therefor we check the vm state when VIR_ERR_OPERATION_INVALID is returned. If the VM is down, then we return success for the destroy operation.


peter-boden (dismissed):
LGTM

PR #486, opened 2026-08-26
thinp: Do not extend drives while monitoring is disabled

dupondje:
The event-driven extend path (on_block_threshold / on_enospc -> _extend_drive_soon -> dispatched _extend_drive) was not gated on the monitor's enabled state, unlike the periodic monitor. Volume monitoring is disabled while the volume chain is being mutated - during live merge pivot, snapshot, and disk replica pivot - or when extension is unsafe (migration destination cannot refresh disks).

When a block threshold or ENOSPC event arrived during such a window, an extend task was still dispatched. During a live merge, the task could block in query_block_stats() until the pivot completed, then look up the old active volume (still in drive.volumeChain) against libvirt's post-pivot chain and raise an unhandled VolumeNotFound. The extend never completed and an ENOSPC-paused VM was left paused.

Gate _extend_drive_soon() on the enabled state so event-driven extends are deferred while monitoring is disabled. The drive is still marked for extension, so once monitoring is re-enabled the periodic monitor extends it and resumes the VM. This is consistent with the intent of every disable() call site, and the added check does not reintroduce the isDomainReadyForCommands() latency that the event path was designed to avoid.

Also make _update_block_info() resilient to VolumeNotFound so a chain change racing with an in-flight extend logs a warning and retries in the next monitoring cycle instead of raising an unhandled exception.

### DIFF
--- a/lib/vdsm/virt/thinp.py
+++ b/lib/vdsm/virt/thinp.py
@@ -248,6 +248,19 @@ class VolumeMonitor(object):
     def _extend_drive_soon(self, drive):
         if self._dispatch is None:
             return
+        if not self._enabled:
+            # Volume monitoring is disabled, typically during live merge while
+            # the volume chain is being modified and cannot be trusted.
+            # Dispatching an extend now may race with the pivot and query a
+            # stale volume chain. The drive is already marked for extension, so
+            # the periodic monitor will extend it once monitoring is enabled
+            # again, resuming the VM if it paused on ENOSPC.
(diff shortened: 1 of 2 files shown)


JasperB-TeamBlue (approved):
LGTM!

peter-boden (approved):
I'm not sure that we need the verbose inline comments in `tinp.py`? 
Not blocking, just a remark. Otherwise LGTM.

PR #488, opened 2026-09-03
tool: disable caw on sanlock

dupondje:
Since sanlock 5.0.0 (commit e68723c12a), it defaults to use caw.

But this has 2 problem.
- First of all it is incompatible with older sanlock versions, and as we want to keep compatibility with EL9/CentOS9 hosts, we need to disable it.
- It also causes sanlock to run as uid=0 (root).

So disable it for now. The new config entry is just ignored on older sanlock versions. So we can just add it without problems.

### DIFF
--- a/lib/vdsm/tool/configurators/sanlock.py
+++ b/lib/vdsm/tool/configurators/sanlock.py
@@ -135,6 +135,14 @@ def _config_for_vdsm():
         # See https://bugzilla.redhat.com/1902468
 
         "max_worker_threads": "50",
+
+        # Since sanlock 5.0.0 (commit e68723c12a), there is a new caw
+        # options. Which is enabled by default.
+        # But this breaks compatibility with older versions of sanlock.
+        # So we disable it explicitly.
+        # Next to that, it also causes sanlock to run under uid 0 (root).
+
+        "use_compare_and_write": "no",
     }
 
 
@@ -154,6 +162,8 @@ def _restart_needed():
     options = _daemon_options()
     if options is not None:
         for key, value in _config_for_vdsm().items():
+            if key == "use_compare_and_write" and key not in options:
+                continue  # unknown to sanlock < 5.0, conf entry is ignored
             # When upgrading sanlock < 3.8.3, "max_worker_threads" is missing.
             # See https://bugzilla.redhat.com/2013383
             if options.get(key) != value:


peter-boden (changes requested):
I think we need something extra. tl;dr we can have an unwanted restart of sanlock on upgrade of vdsm on EL9. I don't think it will have a huge impact, but the fix if really small.

AI explains it better than me 😅 : 
  
   ### The chain

   1. Commit adds `"use_compare_and_write": "no"` to the wanted options:
      [`sanlock.py` `_config_for_vdsm()`](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurators/sanlock.py#L146)

   2. `isconfigured()` → `_restart_needed()`:
      [`sanlock.py` L31-L44](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurators/sanlock.py#L31)

   3. `_restart_needed()` compares every wanted option against the *running daemon's* options
      (parsed from `sanlock client status -D`):
      [`sanlock.py` L156-L170](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurators/sanlock.py#L156)

      ```python
      for key, value in _config_for_vdsm().items():
          if options.get(key) != value:
              return True
      ```

   ### The problem

   - sanlock only prints `use_compare_and_write=` in `client status -D` output since **5.0.0**
     (upstream `src/cmd.c`, `print_state_daemon()`).
   - EL9 ships sanlock 3.8.x → the key is **never** in the output.
   - `options.get("use_compare_and_write")` → `None` ≠ `"no"` → `_restart_needed()` → **True, forever**.
   - Restarting can't fix it: the old binary will never report the key.

   ### Who acts on this

   - `vdsm-tool is-configured` exits non-zero permanently:
     [`configurator.py` L132-L143](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurator.py#L132)
   - `vdsm-tool configure` always considers the sanlock module unconfigured and restarts its services:
     [`configurator.py` `_should_configure()` L293](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurator.py#L293)
     [`configurator.py` service stop/start L262-L276](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurator.py#L262)
   - The restarted service is **sanlock** itself, declared here:
     [`sanlock.py` L23](https://github.com/oVirt/vdsm/blob/master/lib/vdsm/tool/configurators/sanlock.py#L23)
   - Trigger: vdsm RPM `%post` scriptlet runs this on **every package install/upgrade**:
     [`vdsm.spec.in` L727-L729](https://github.com/oVirt/vdsm/blob/master/vdsm.spec.in#L727)

     ```
     if ! vdsm-tool is-configured; then
         vdsm-tool configure --force
     fi
     ```

   ### Impact

   - On EL9 hosts, every vdsm upgrade: sanlock is stopped/started for no reason
     (config file already matches; nothing changes).
  -  vdsm-tool is-configured permanently non-zero on EL9 — misleading diagnostics for anyone/anything relying on it.

   ### Why the existing `max_worker_threads` comment doesn't cover it

   `max_worker_threads` is missing only on sanlock < 3.8.3, where upgrading the package (required anyway)
   makes the key appear → the check converges. `use_compare_and_write` is missing on every EL9 sanlock
   version, permanently — exactly the platform this commit intends to support.

   ### Suggested fix

   Skip the key when the daemon doesn't report it (only for `use_compare_and_write`; keep the
   `max_worker_threads` behavior intact):

   ```python
   for key, value in _config_for_vdsm().items():
       if key == "use_compare_and_write" and key not in options:
           continue  # unknown to sanlock < 5.0, conf entry is ignored
       if options.get(key) != value:
           return True
   ```

dupondje:
@peter-boden : good remark! Fixed

JasperB-TeamBlue (approved):
LGTM!

peter-boden (approved):

PR #489, opened 2026-09-08
virt: set compression alg during parallel migration

dupondje:
When doing parallel migrations, we need to specify a compression algorithm to be used for the parallel migration. There is no default. We need to set it via VIR_MIGRATE_PARAM_COMPRESSION.

Otherwise the following error is thrown:
"No compression algorithm selected for parallel migration"


peter-boden (approved):
Maybe we could make the compression algorithm configurable? But certainly not a must, so LGTM.

JasperB-TeamBlue (approved):
LGTM!

